PR #8, opened 2021-11-11
Add HiGHS open source solver

pz-max:


pz-max:
I haven't really included the "basis_fn" into the code. What is the purpose of it?
Will ask fneum in a moment when testing the installation.

pz-max:
This PR needs a bit more time. The solving works though what does not work is reading out the right results.
To DO's:
- [ ] The installation needs to be fixed to one version.
- [ ] The solution file need to be read out correctly. While the obj. is the same for CBC, highs, glpk, the optimised capacities are different. This suggest that we need to read out the data correctly (maybe also a bug in HiGHS that was fixed in a newer version)

FabianHofmann:
> I haven't really included the "basis_fn" into the code. What is the purpose of it? Will ask fneum in a moment when testing the installation.

The basis_fn is used to write out the basis of the solution (which can be used for a warmstart later), however if I remember correctly the IPM does support this (only used for simplex) so you can just ignore this. 

Thanks for the PR, I can do the rest. Just need some time since I want to prioritze the PIPS first.

pz-max:
@FabianHofmann, I will solve the two points/tasks above.
Yes, I think the PIPS solver might be the most interesting one for big problems where HPC need to be used.

pz-max:
Should work now according to https://github.com/PyPSA/PyPSA/pull/308#issuecomment-968326026

pz-max:
Adjustments according to https://github.com/PyPSA/PyPSA/pull/308. I think the HiGHS solver is close to be ready to merge. Only one thing:

In the above PR, I had to change `solver_options={}` so that the code works also with an undefined options dictionary. Something we might need to deal with here too. Feel free to make this change if required or let me know what you wish to see (I wasn't sure about the prefered architecture)

FabianHofmann:
@pz-max thanks for the code, I made some changes to make things more light-weight. Also moved some of the docstring to the solvers.rst in the documentation. Hope your fine with the changes, see #17 (all your commits will still show up in the contribution)

PR #63, opened 2022-11-13
Composition design review

FabianHofmann:
Most changes are targeting back-wards compatibility.

### DIFF
--- a/linopy/constraints.py
+++ b/linopy/constraints.py
@@ -6,7 +6,7 @@ This module contains implementations for the Constraint{s} class.
 """
 
 import re
-from dataclasses import dataclass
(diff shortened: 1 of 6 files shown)


coroa:
> @coroa I don't know if you even have time to review this. Otherwise I'd just merge this. The only thing that I was wondering about: was there any reason for not defining `LinearExpression` as a `dataclass`?

Agreed, I had quite some time issues the last week. Not much better this week though :).

I didn't `dataclass` some of the things because, i had intended to write actual `__init__` functions as well as `__repr__`, so that there is not much benefit to dataclasses left. But it was all just on a whim.

PR #68, opened 2022-11-30
Define field as default_factory to specify mutable default value

danuriegas:


linopy/__init__.py
```diff
@@ -11,4 +11,7 @@
 from linopy.io import read_netcdf
 from linopy.model import LinearExpression, Model, Variable, available_solvers
 from linopy.remote import RemoteHandler
-from linopy.version import version as __version__
+
+# from linopy.version import version as __version__
+
+__version__ = "0.0.14"
```
FabianHofmann, line 17: Could you revert this? The version specs are automatically  updated.

FabianHofmann (changes requested):
Thank you @danuriegas for your contribution! Looks good. This was also tackled in #63 , but for a temporary fix we can merge this.

linopy/__init__.py
danuriegas: Happy to contribute (for the first time)! I will revert this, no problem. 
Do you know why am I getting an error "ModuleNotFoundError: No module named 'linopy.version'" when importing the module?

FabianHofmann: I would guess this is due to `setuptools-scm` not being installed...

FabianHofmann (approved):

Issue #85, opened 2023-01-11
disturbing output when importing linopy on windows without cbc/glpk as available_solvers


jankaeh:
In solvers.py the lines 30&33 create unwanted(?) output:

if sub.run([which, "glpsol"], stdout=sub.DEVNULL).returncode == 0:
    available_solvers.append("glpk")

if sub.run([which, "cbc"], stdout=sub.DEVNULL).returncode == 0:
    available_solvers.append("cbc")

Everytime I import linopy on windows without cbc/glpk available, I get this twice:
INFORMATION: Es konnten keine Dateien mit dem angegebenen
Muster gefunden werden.

adding "stderr=sub.STDOUT" like so removes this output:
if sub.run([which, "glpsol"], stdout=sub.DEVNULL, stderr=sub.STDOUT).returncode == 0:
    available_solvers.append("glpk")

I am not sure if this is what you want?

FabianHofmann:
I see, but the import still works right? Seems that windows prints out this extra information. I am not familiar with windows machines, would be nice to suppress this information.

jankaeh:
Yes the import still works. I just was confused because I had no idea where that information came from. If the changes do not affect the expected behaviour on linux, it'd be nice to supress the info.

PR #87, opened 2023-01-11
supress disturbing output when cbc/glpk are no available_solvers

jankaeh:
avoid disturbing output (INFORMATION: Es konnten keine Dateien mit dem angegebenen Muster gefunden werden.) see: https://github.com/PyPSA/linopy/issues/85

works on windows. i don't know if this changes the behaviour on linux though


FabianHofmann (approved):
Great, looks like a good solution!

PR #93, opened 2023-01-18
make highs default solver

FabianHofmann:


fneum:
Do you want to include a version constraint for highspy?

FabianHofmann:
Good idea, I didn't see problems with other versions though

FabianHofmann:
@fneum it is actually running through :rocket:

PR #100, opened 2023-03-01
adding lower, greater

Cellophil:
Embarrassingly I ran into this - it works, it just falls back to the xarray mechanics and does something to the indices. Maybe good to prevent with little effort.


FabianHofmann (approved):
Fair enough. Thanks @Cellophil

PR #101, opened 2023-03-01
Constants in linexpr

Cellophil:
Hi, sorry for the delay, I checked it out. Adding constants seems to work, see the tests, but it doesn't yet do constraints. My experience/elegance with xarray is unfortunately lacking, could you check Linexpression class the "const" property?

### DIFF
--- a/linopy/expressions.py
+++ b/linopy/expressions.py
@@ -161,7 +161,8 @@ class LinearExpression:
         if isinstance(data, (int, float)):
             # should allow arraylike
             da = xr.DataArray([data], dims=["_term"])
-            data = Dataset({"coeffs": da, "vars": [CONSTANT]})
+            va = xr.DataArray([CONSTANT], dims=["_term"])
+            data = Dataset({"coeffs": da, "vars": va})
 
         if not isinstance(data, Dataset):
             raise ValueError(
@@ -276,17 +277,16 @@ class LinearExpression:
         if other == 0:
             return self
         else:
-            return NotImplemented
+            return self.__add__(other)
 
     def __sub__(self, other):
         """
         Subtract others from expression.
         """
-        if not isinstance(other, (LinearExpression, variables.Variable)):
-            raise TypeError(
-                "unsupported operand type(s) for +: " f"{type(self)} and {type(other)}"
-            )
-        return merge(self, -other)
+        return self.__add__(-other)
+    
+    def __rsub__(self, other):
+        return self.__radd__(-other)
 
     def __neg__(self):
         """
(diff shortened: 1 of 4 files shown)


FabianHofmann:
Hey @Cellophil, nice work. I won't have time to review it this or next week. On a first glimpse it looks nice. Perhaps in the test you could make sure that array of constant are added as well?

linopy/expressions.py
```diff
@@ -266,17 +284,16 @@ def __radd__(self, other):
         if other == 0:
             return self
         else:
-            return NotImplemented
+            return self.__add__(other)
```
coroa, line 287: Do not do this!

`__add__` is preferred over `__radd__` anyway by the operations engine of python, unless `self` is of a type that inherits from `other`. In the latter case after `NotImplemented` python tries `__add__` anyway. 

Check https://docs.python.org/3.6/library/constants.html#NotImplemented and https://docs.python.org/3.6/library/numbers.html#implementing-the-arithmetic-operations for details.

Cellophil: Ok good to know, sounds logical, but when trying it out the interpreter complains about not knowing what to do with "float + LinearExpr"... Can you get the left addition of constants to work like that?

Cellophil: Ok I see calling __add__ is rather horrible, but radd needs to cover the case where other is a constant... So probably be clean about it and write the radd logic out?

Cellophil:
Ok I have things like 
```
c1 = x >= y
m.add_constraints(c1)
c4 = 0 >= (x - 1) + 2.
m.add_constraints(c4)
```
running. The wildly first broadcasts the dimension of x and y, that is probably not intended...
Constant aggregation is done via numpy, probably would be better to stay within xarray. Both Pandas and Xarray seem to try to handle "pd.Series + LinExpr" via their `__add__` - I'm not sure how to prevent that.

For familiy reasons I will be gone very soon for a month, feel free to adapt and overwrite everything of course!
Best,
Andreas

linopy/expressions.py
coroa: Sorry, i was just going over a lot of different work in parallel, so i did not grasp the deeper meaning of what you were trying to do! Should have asked!

It makes sense now to me.

In your particular case i would go for
```
__radd__ = __add__
```
to clarify that the same code should be used for right addition than for general.

Otherwise i would have tried to abstract. That particular way of course does not work for the rsub, so i guess after all the way you are calling between the methods is actually fine.

And the confusion is mine. Apologies.

coroa:
I'll try to review this properly somewhen on the weekend.

FabianHofmann:
> I'll try to review this properly somewhen on the weekend.

@coroa that would be great

FabianHofmann:
@coroa any chance of finding some time for reviewing? Otherwise, I can give it a try.

Cellophil:
Hi everyone, I'm back and would be interested to continue this. I realized that my current sorting of right and left hand side has a flaw, it would subtract the constant part from the left hand side, ie null it, but it wouldn't remove the columns. Removing the columns instead would probably do the trick.

FabianHofmann:
@Cellophil are you currently working on this? Otherwise, I would look into it later today.

Cellophil:
@FabianHofmann no but happy to pick it up. Maybe you can comment; I definitely need to replace the crude numpy functions when manipulating xarray.

Cellophil:
Hi @FabianHofmann,

I worked a bit on this and dropped the constants from the final lhs LinearExpression. A couple of other tests fail due to not raising errors anymore ("expr + 10"). To progress I would need to understand the structure better... For instance, some tests demand that the rhs is a constant. It could now just be a linearexpression instance, or a DataArray. "expr +10" should probably be broadcasted according to whatever the dimensions of expr are, which would make the number 10 a full DataArray.

What I didn't quite understand is why _term is a dimension without coordinates, whether that is intended. I would have expected a rectangular object where one axis is the variable indexes, and the other is a running index. Are there cases where the variable index is not constant across the "column"? This could very well be me not understanding xarray fully.

In the tests there are examples where two variables x and y of different length are added up, which is just exploded in the dimension, this seems to be intended, probably some kind of broadcasting?

Adding up linear expressions seems to work like in the tests. I would appreciate some feedback on this.

Best,
Andreas

Issue #112, opened 2023-04-19
Performance - omit 1.0000 when writing out LP file


willu47:
linopy LP files are created with lots of extra characters when the coefficients are 1:

```
c2:
-1.000000 x6
+1.000000 x8
=
0.000000
```
These can be omitted:
```
c2:
- x6 + x8 = 0.00000
```

willu47:
I've tested with Gurobi, GLPK and CBC and those seem fine.  I don't have access to other solvers at the moment.

FabianHofmann:
closed by #114

Issue #111, opened 2023-04-19
Precision of floats in lp files


jafluri:
Hello,

I am currently migrating a package from pyomo to linopy (great work btw). When comparing results from the old version to the results with linopy, I noticed that the results always match pretty well but not exactly, with differences up to around ~1% for complex models. 

I came to realize that the main issue was the precision in which the lp files are written. From what I understand, all of the float to string conversion is happening in

https://github.com/PyPSA/linopy/blob/62ac3ea8f4b3cefef9ca7df55132e6161a8fb699/linopy/io.py#L38-L46

Is there anything against upping the precision per default, e.g. from `%f` -> `%.12f` and `%f` -> `%+.12f` respectively?
If not, it would be great if you could add this. Otherwise, I could create a PR where this is a new parameter that can be passed to the model.

FabianHofmann:
Hey @jafluri, thanks for pointing this out. Correct me, if I am wrong, but the default precision seems to be 6. Since we do not want to bloat the LP size, could you try out whether a precision of 8/9 is already satisfying?

jafluri:
Thanks for the quick reply. I just ran the model for 6, 9, 12 and 15 significant digits. The objective for 9, 12 and 15 is identical. Comparing the variable values to the ones from pyomo, you get the smallest differences with 12 and 15, but I guess 9 should be enough as some floating point jitter is to be expected in large models.

jafluri:
In combination with #112, one could do `"%+.12g"`, that strips away the tailing zeros:

`array([1.       , 1.2      , 1.2334567]) -> array(['+1', '+1.2', '+1.2334567'], dtype=object)`

I'd assume that way you'd safe memory in most cases, even if you go up to 12, compared to the current implementation.

FabianHofmann:
Oh, this sounds great! Could you do a PR?

jafluri:
Thanks for the quick resolution, I'm closing this then.

PR #114, opened 2023-04-21
Float representation in lp files

jafluri:
This PR changes the float to string conversion for #111 and #112 . The new format is `.12g` (or `+.12g`), which increases the precision of the float and removes tailing zeros of numbers, e.g.

`array([1.       , 1.2      , 1.2334567]) -> array(['+1', '+1.2', '+1.2334567'], dtype=object)`


FabianHofmann:
Awesome!

PR #121, opened 2023-05-08
Allow addition and subtraction by 0 on lin. exprs. and variables

koen-vg:
This is just a convenience feature, but I think it would be nice.

I was summing up a bunch of different expressions, but some of them only sometimes depending on some configuration settings. These might be `None`. To sum them, I did something like
````python
S = sum(e for e in [a, b, c] if e is not None)
````
With this pull request, I can set each of the summands to `0` instead of `None` as needed, and the above line is just
````python
S = sum([a, b, c])  # or: S = a + b + c
````

### DIFF
--- a/doc/release_notes.rst
+++ b/doc/release_notes.rst
@@ -1,8 +1,10 @@
 Release Notes
 =============
 
-.. Upcoming Release
-.. ----------------
+Upcoming Release
+----------------
+
+* Allow addition of 0 to `Variables` and `LinearExpressions`.
 
 Version 0.1.5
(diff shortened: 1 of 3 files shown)


FabianHofmann:
Great, thanks @koen-vg! Could you merge the latest master to trigger the CI?

FabianHofmann:
Thanks again @koen-vg for the PR.
The feature is already covered by the latest changes, which allows adding numeric values to a linear expression. Please reopen or raise another issue, if you encounter problems

PR #146, opened 2023-07-07
Missing + sign in objective function

aurelije:
I think I have reported this problem long time ago but it is still there. One missing sign, not a big deal but makes confusion


FabianHofmann (commented):
thanks @aurelije

Issue #145, opened 2023-07-06
Implement cumsum of Variables


hblunck:
For many optimization models that consider time, it would be helpful to be able have a shortcum to calcualte the cumulated sum along a dimension (e.g. to calculate the inventory position as the cumulated sum of all inbound orders minus demand).

This functionalty is available in xarray but apparently not in linopy (as of Version 0.2.2).

hblunck:
A walkaround right now is to use the `rolling` method and then call `.sum()` on the result
> inventory_position.rolling(time=10000).sum()
(where 10000 is just a large number, can be longer than the size of the dimension)

FabianHofmann:
Hey @hblunck, this would actually be a nice feature. Since the formulation is quite similar, your approach might even fit for the final function by setting the window size to the size of the dimension. I am not sure about the performance, but anyway this could suit for a start. Would you be interested in drafting a PR?

hblunck:
Hey @FabianHofmann , thanks for the encouraging feedback. I did take a shot at it in #147. I don't think that is complete yet, though.

PR #147, opened 2023-07-12
`.cumsum` for Variables and LinearExpression

hblunck:
Addresses #145 

Implements (to the best of my knowledge) `.cumsum` methods for `Variable` and `LinearExpression` objects. No implementation is provided yet for e.g. `LinearExpressionGroupBy` objects.

### DIFF
--- a/linopy/expressions.py
+++ b/linopy/expressions.py
@@ -23,6 +23,7 @@ from numpy import arange, array, nan
 from scipy.sparse import csc_matrix
 from xarray import DataArray, Dataset
 from xarray.core.dataarray import DataArrayCoordinates
+from xarray.core.types import Dims
 
 from linopy import constraints, expressions, variables
 from linopy.common import (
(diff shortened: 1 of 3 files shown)


linopy/expressions.py
```diff
@@ -586,6 +587,54 @@ def sum(self, dims=None, drop_zeros=False) -> "LinearExpression":
...
+        Returns
+        -------
+        linopy.expression.LinearExpression
+        """
+        # Along every dimensions, we want to perform cumsum along, get the size of the
+        # dimension to pass that to self.rolling.
+        if not dim:
+            # If user did not specify a dimension to sum over, use all relevant
+            # dimensions
+            dim = [d for d in self.data.dims.keys() if d != "_term"]
```
FabianHofmann, line 631:
```suggestion
            dim = self.coord_dims
```

FabianHofmann (commented):
Very nice, thanks @hblunck. Just one small thing

hblunck:
Hey @FabianHofmann , I added some tests, but they fail. On inspection (i.e. printing), the linear expressions are exactly what I would expect and they are also equal. But `assert_linequal` is throwing errors at me that I cannot decipher. I think I would have to hand it over to you at this point.

FabianHofmann (commented):
I don't have a good solution. Let's keep it simplistic then

PR #148, opened 2023-07-13
A PyPSA-based problem is not necessary linear.

LucieRC:
Just changed the log message to be general (before stating "linear model" even if a MILP f. inst.).


FabianHofmann (approved):
Thanks @LucieRC !

PR #149, opened 2023-07-17
Include Transport Tutorial

glatterf42:
The transport tutorial seems to be a standard tutorial for showcasing usage of solvers and related packages. [GAMS](https://www.gams.com/latest/gamslib_ml/libhtml/gamslib_trnsport.html) has it as well as [pyomo](https://nbviewer.org/github/Pyomo/PyomoGallery/blob/master/transport/transport.ipynb) and, maybe in particular, [ixmp](https://github.com/iiasa/ixmp/blob/main/tutorial/transport/py_transport.ipynb) has its own version of it. We are currently rewriting ixmp to be all-python and we are thinking of moving away from GAMS in the near future. Therefore, we are considering how our implementation choices now might enable or harm a future move to solver interfaces such as linopy. 

After reading the documentation and missing the transport tutorial for comparison, I decided to try adding it myself, which is what this PR does. While I followed your [contributing guidelines](https://linopy.readthedocs.io/en/latest/contributing.html#contributing-examples) for the most part, I am not sure where in the docs this tutorial should be linked, so I did not complete step 4 yet. Please also check that:

- [x] this is actually the best way to model the transport problem in linopy (coordinated with @aurelije)
- [x] the text I have written is correct (includes feedback)
- [x] keeping the GAMS/pyomo commands for comparison is not too confusing (restructured based on feedback for readability)

Regarding this last point, I should explain that most comments follow this structure: lines on top are usually from GAMS and marked with single `#`, lines below that (marked with `##`) are from pyomo. If additional lines appear using just one `#` again, they are new comments intended to aid with the tutorial.

I have one note regarding your contributing guidelines: the [pre-commit hook](https://jamesfolberth.org/articles/2017/08/07/git-commit-hook-for-jupyter-notebooks/) you linked did not work out of the box for me (using Ubuntu 22.04) because bash is not the default shell there and [needs to be specified](https://stackoverflow.com/a/2462357).

Other than that, I want to commend your documentation. In general, it was not a big problem to write up this tutorial without prior experience with pyomo or linopy. The most confusing part for me was the lack of sets and parameters as model attributes, it might be worth expanding your docs (migrating from pyomo) in that regard. Or do you plan to include sets and parameters as model attributes in the future?

### DIFF
--- a/doc/transport-tutorial.nblink
+++ b/doc/transport-tutorial.nblink
@@ -1,0 +1,3 @@
+{
+    "path": "../examples/transport-tutorial.ipynb"
+}
--- a/examples/transport-tutorial.ipynb
+++ b/examples/transport-tutorial.ipynb
@@ -10,7 +10,9 @@
     "\n",
     "The goal of the Transport Problem is to select the quantities of a homogeneous good that has several production plants and several punctiform markets as to minimise the transportation costs.\n",
     "\n",
-    "It is the default tutorial for the GAMS language, and GAMS equivalent code is inserted as single-dash comments. The original GAMS code needs slighly different ordering of the commands and it's available at http://www.gams.com/mccarl/trnsport.gms. The Pyomo version of the tutorial can be found at https://nbviewer.org/github/Pyomo/PyomoGallery/blob/master/transport/transport.ipynb. For comparison, the Pyomo code is kept as additional comments for the moment.\n",
+    "It is the default tutorial for the GAMS language, and GAMS equivalent code is inserted as single-dash comments. The original GAMS code needs slighly different ordering of the commands and it's available at http://www.gams.com/mccarl/trnsport.gms. The Pyomo version of the tutorial can be found at https://nbviewer.org/github/Pyomo/PyomoGallery/blob/master/transport/transport.ipynb. \n",
+    "\n",
+    "For comparison, the variable names have been kept the same, though they are not pythonic. The equivalent GAMS code is included at the top of each cell as comments, while the Pyomo-equivalent is stated as text comments below.\n",
     "\n",
     "## Problem Statement\n",
     "\n",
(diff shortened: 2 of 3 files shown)


FabianHofmann:
@glatterf42 thank you very much for the PR! I will have a look at it asap, but I cannot say when that is (likely end of the week)

FabianHofmann:
Thanks @aurelije, for the comment. I think the code you provided is very clean and nice. So I would definitely make sense to include it.

glatterf42:
Thanks for the comment, @aurelije, and sorry for taking so long, everyone. I've now updated the tutorial according to your suggestions, e.g. getting rid of numpy etc. 
There is one question that maybe @FabianHofmann could clarify: where in the docs can I find that `.sum()` is default for the objective? It might be nice to add that to explain why defining `c * x` is enough in this case.

FabianHofmann:
> Thanks for the comment, @aurelije, and sorry for taking so long, everyone. I've now updated the tutorial according to your suggestions, e.g. getting rid of numpy etc. However, I kept the variable names analogous to GAMS/Pyomo to keep comparison with these tutorials easy. There is one question that maybe @FabianHofmann could clarify: where in the docs can I find that `.sum()` is default for the objective? It might be nice to add that to explain why defining `obj = c * x` is enough in this case.

Hey @glatterf42, that is correct. Since `linopy` assumes only one objective function, it automatically takes the full sum. So, either making a small note or explicitly taking the sum should be fine :) 

I really like the example, great work! Would it be possible to pull some of the comments in the code cells into the text? For the comparison with the pyomo code you could insert boxes like 

*Equivalent in Pyomo*
```python
model.i = Set(initialize=['seattle','san-diego'], doc='Canning plans')
model.j = Set(initialize=['new-york','chicago', 'topeka'], doc='Markets')
```

glatterf42:
Thanks for the clarification and suggestion, let me know what you think of the new structure :)

FabianHofmann:
Nice work! Just roughly scanning, I would be happy to merge it. Let me know when you think it is ready.

FabianHofmann:
Thanks @glatterf42 for this nice contribution! I will make sure it appears on the doc asap.

PR #162, opened 2023-09-18
Enclose env usage in with keyword for gurobi

tgi-climact:
**Description of the PR**
When using a commercial WLS (Web License Service) license from Gurobi, we want to avoid any unwanted use of a token. This is especially the case when we are using the WLS license with single-use token (one token allowed at the time). 

License usage is managed through the `gurobipy.Env` object. Therefore, we need to use `gurobipy.Env` wisely and allow the garbage collector to clean the object when appropriate (which is releasing the license). This PR suggest to use a `with` context to easily track/clean the usage of the object. This is espacially a must when using a Jupyter notebook to run the code.

**Gurobi documentation**
https://www.gurobi.com/documentation/current/refman/py_env.html

### DIFF
--- a/linopy/io.py
+++ b/linopy/io.py
@@ -319,7 +319,7 @@ def to_file(m, fn):
     return fn
 
 
-def to_gurobipy(m):
+def to_gurobipy(m, env=None):
     """
     Export the model to gurobipy.
 
@@ -330,6 +330,7 @@ def to_gurobipy(m):
     Parameters
     ----------
     m : linopy.Model
+    env : gurobipy.Env
 
     Returns
     -------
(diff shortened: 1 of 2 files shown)


FabianHofmann (approved):
@tgi-climact thank you for your PR! This sounds and looks very reasonable

aurelije:
One of pluses of Linopy comparing to usage of Pandas with gurobi or even gurobi own pandas integration is the fact that you do not use env until very last point in time. How many times we have clogged gurobi server just by debugging some constraint in code but to get to that point environment gets created and immediately gurobi server gets his resource "used" even if you will never call optimize thus it will never do actual calculation... Smart from them, we are forced to buy a bigger license to have more free slots on server :)

FabianHofmann:
@aurelije good to hear that! 

@tgi-climact I already included some small release note on the changes

PR #165, opened 2023-09-20
limit version of setuptools_scm as it breaks the syntax of the auto-c…

jankaeh:
limit version of setuptools_scm to smaller than 8 [(setuptools_scm/Changelog)](https://github.com/pypa/setuptools_scm/blob/main/CHANGELOG.rst) as it breaks the syntax of the auto-created version.py under windows.


jankaeh:
closes #166

FabianHofmann:
thank you @jankaeh for the PR!

PR #167, opened 2023-09-21
Add note in release note

tgi-climact:
Add release note for https://github.com/PyPSA/linopy/pull/162


FabianHofmann:
thanks @tgi-climact, I actually already added the note for the latest release, see https://github.com/PyPSA/linopy/blob/master/doc/release_notes.rst

PR #172, opened 2023-10-02
add option to pass existing gurobipy.Env to run_gurobipy()

fneum:
Quick successions of gurobi environment instantiation may cause clogging on Gurobi token server since tokens are not released quickly enough.

Gurobi recommends that in this case, one should do multiple solving calls inside the same environment.

https://support.gurobi.com/hc/en-us/articles/4424054948881

```py
import gurobipy as gp

with gp.Env() as env:
    with gp.Model(env=env) as model1:
        # formulate and solve model1
    with gp.Model(env=env) as model2:
        # formulate and solve model2
```

E.g. in PyPSA the primary use case would be rolling horizon optimisation.

```py
with gurobipy.Env() as env:
    n.optimize.optimize_with_rolling_horizon(
        ...,
        env=env,
    )

### DIFF
--- a/linopy/model.py
+++ b/linopy/model.py
@@ -921,6 +921,7 @@ class Model:
         basis_fn=None,
         warmstart_fn=None,
         keep_files=False,
+        env=None,
         sanitize_zeros=True,
         remote=None,
         **solver_options,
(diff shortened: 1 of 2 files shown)


fneum:
This is kind of related to #162, since this was a prerequisite for this PR to work.

FabianHofmann:
A bit unfortunate but the workaround seems reasonable

Issue #159, opened 2023-09-14
logging disabled after call to optimise when Gurobi raises exception


dannyopts:
When running the gurobi solver linopy calls logging.disable(50) before m.optimize() but if an exception is thrown in the solver, logging will remain disabled.

Also a more minor issue is that the line that renables logging calls logging.disable(1), which could also cause people problems if they had explicitly set logging.disable before. 

### Problem this caused me:
I use linopy indirectly through PyPSA. 

Occasionally when running in production all logs would stop being shown. 

I now wrap all calls to optimise in a try except with a logging.disable(logging.NOT_SET) call, but it would be great if I didnt have to do this. It was also a nightmare to debug because it was only happening in production and I had no idea what was causing this, I use linopy indirectly through PyPSA and occasionally logs would just stop appearing, it took quite a long time before I started trawling the linopy source code.

### Potential solutions
1. try except and call logging.disable(1) in the except.
2. add a flag which allows users to opt out of you changing their logging config. 
3. Target only the gurobipy logger - for example https://stackoverflow.com/a/58547894 or https://groups.google.com/g/gurobi/c/sV7xxN_mzCk
```
grblogger = logging.getLogger('gurobipy')
grblogger.propogate = False
```

I think 2 is a better option because if the application calling linopy is multithreaded, then 1 and to a lesser extent 3 will impact those threads. Possibly 2 and 3 could both be added, but that seems unnessecary, I will happily endure duplicate logs and I can simply add 
```
logger = logging.getLogger('gurobipy')
logger.propagate = False
``` in my application code to prevent the duplication.. 

I am happy to implement and raise a PR for any agreed solution.

github.com/PyPSA/linopy/blob/8f8682bb50f94bee742aca655aadc38532787c89/linopy/solvers.py#L542

**Comment by dannyopts** [comment]
Having thought about this some more, I think that there is a forth option which makes the most sense to me. 

4. Remove this code entirely, leave the users of the library to configure their logging. Maybe add a log message or some documentation that instructs users how to correctly configure their logging when using Gurobi.

Out of the box, logging will not be duplicated if you remove the logging.disable(50), unless you have configured a python logger. So is it the responsibility of this lib to handle all the possible crazy ways that people could configure their logging, and also think about all the possible deployment scenarios it could be used in, and make logging work in all of them?

I would say not, but I guess thats up to the core devs to decide.

```
from linopy import Model
import logging

# Uncomment this line on it's own to get duplicate logs. 
[...]

PR #174, opened 2023-10-03
Remove logging.disable from gurobi solve

dannyopts:
The gurobi logger will by default both print to console and propogate to the root logger. If the root logger has a handler set this can result in duplicate logs. 

Disabling the logging causes a variety of issues such as when code is running in parrallel, or if an error is thrown during the solve. 

Instead the user can configure the gurobi logger to not propogate to the root logger. 

This PR removes the disabling of the logger and documents how a user can fix the double logging issue. 

Let me know if you would like any changes.

image

Resolves: https://github.com/PyPSA/linopy/issues/159#issuecomment-1725116554


dannyopts:
Just spotted the typo in double. I will correct and push again

dannyopts:
Fixed the typo

FabianHofmann (approved):
Great! Thanks @dannyopts!

PR #184, opened 2023-10-27
add python 3.12 to CI

### LINKED ISSUE #265: MPS IO changes with highspy >= 1.7
As pointed out in #264, the new highspy package does not switch coefficient signs when exporting a MPS file of a maximizing problem. A warning should be added when people use maximization with MPS IO or the code should be checking the highspy version at https://github.com/PyPSA/linopy/blob/ecbc96266e3195b756fa528ff2faa4afe9243743/linopy/solvers.py#L124

FabianHofmann:
closes #265

### DIFF
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -19,7 +19,7 @@ jobs:
         - 3.9
         - "3.10"
         - "3.11"
-        # - "3.12"
+        - "3.12"
         os:
         - ubuntu-latest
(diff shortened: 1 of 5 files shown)


maurerle (approved):

maurerle:
Just to mention that the latest commit messages states to "disable cplex for python < 3.12" while it actually does the opposite. So you might want to `--amend` that and force push the commit message @FabianHofmann :)

Issue #192, opened 2023-10-31
Gurobi race condition


dannyopts:
I am currently experiencing a bug when using a remote compute gurobi solver. 

The issue is that the env is destroyed before we retrieve the solution, and if we lose a race then we cant retrieve the solution and blow up with an error 

```
AttributeError: Failure writing output to destination (code 23, node http://localhost:61000, command PUT http://localhost:61000/api/v1/jobs/b438a8b5-65db-4c6c-9a00-d5f4e9fefa53/worker?sync=true&cmd=13). Did you mean: 'ObjNVal'?
```

More details on how to reproduce below but the TLDR is:
The problem is that the job is aborted when we exit the stack on line 582. I presume deletion of this env on the compute side is then an async process. If we ask "in time" we can get the solved values back. But if not then we error out. 

I think the fix is simple and we just need to retrieve the solution before we exit the stack. 

A workaround is to pass in an env rather than letting linopy create for you. 

Happy to create a PR to make this fix.

# Steps to reproduce
You need to have a gurobi compute server running with a valid license:
I am doing this via docker compose
``` docker-compose.yml
version: '3.9'
services:
  compute:
    image: gurobi/compute:latest
    restart: always
    command: --hostname=localhost
    volumes:
       - ./gurobi.lic:/opt/gurobi/gurobi.lic
    ports: 
      - "61000:61000"
```

Create a second lic file to point at this compute server
```local_gurobi.lic
COMPUTESERVER=localhost:61000
```

Then build a simple model and solve it pointing to the gurobi server
``` x.py
from linopy import Model

m = Model()
x = m.add_variables(lower=0, name='x')
m.add_constraints(3*x + 1 >= 10)
m.add_objective(2 * x)
# m.add_constraints(3*x + 1 <= 1)
m.solve(solver_name="gurobi")
```

Now run the model:
```
GRB_LICENSE_FILE="./gurobi.lic" python x.py
```

This "should" solve with no issues.

Now add a time.sleep(5) before calling get_solver_solution in run_gurobi (simulating some slowness in the os) - this seems to be realistic when working with larger models.

```linopy/solvers.py:591
status.legacy_status = condition

  import time
  time.sleep(5)
  def get_solver_solution() -> Solution:
      objective = m.ObjVal
```

Now run the original model again

```
GRB_LICENSE_FILE="./local_gurobi.lic" python x.py
```

When I do this I see 

```
Solved in 0 iterations and 0.04 seconds (0.00 work units)
[...]

PR #193, opened 2023-10-31
Retrieve solution before closing the gurobi env

dannyopts:
Currently the Gurobi env is closed before the solution has been retrieved and this can result in an exception being thrown based on a race condition (described in the issue referenced below)

Fixes https://github.com/PyPSA/linopy/issues/192

### DIFF
--- a/linopy/solvers.py
+++ b/linopy/solvers.py
@@ -579,34 +579,36 @@ def run_gurobi(
             m.read(warmstart_fn)
         m.optimize()
 
-    if basis_fn:
-        try:
-            m.write(basis_fn)
-        except gurobipy.GurobiError as err:
-            logger.info("No model basis stored. Raised error: ", err)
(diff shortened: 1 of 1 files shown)


dannyopts:
Does the fact coverage has gone down mean we cant merge this PR?

It seems to me this is not a change that would require a test, but if it would normally be accompanied by a test, can someone give me a steer what that would look like?

FabianHofmann:
Thanks @dannyopts, the diffs are a bit hard to read. But it seems you intended the block which was not in the with block, is that right? I would not worry about the code coverage of the patch.

dannyopts:
@FabianHofmann yes exactly, we dont exit the stack until after we have read the solution so the gurobi env is not destroyed.

FabianHofmann:
great, thanks!

Issue #197, opened 2023-11-07
MOSEK license token retained on import


jotaigna:
Dear developers
I have noticed a small issue when MOSEK is set up with a remote token license server.
If available, linopy will add MOSEK to the available solvers list in line [57](https://github.com/PyPSA/linopy/blob/b437f933863902041788de95340a0610ccc2d183/linopy/solvers.py#L57) of the current version of `linopy/solvers.py`.
Solvers seem to be added if their python module can be successfully be imported. For MOSEK, this import has the additional step of creating an optimiser task and optimising it.
```python
with contextlib.suppress(ImportError):
    import mosek

    with mosek.Task() as m: 
        m.optimize() #<- This checks out a license and retains it for as long as Model is in memory

    available_solvers.append("mosek")
```
As it stands, this has the unfortunate side effect of checking out one license token and retaining it for as long as `Model` is imported. 

This token is not usable to solve models in the same session, because typically one would create new instances of model as

```python
m = Model()
```
but this is a different instance and, at least in my testing, I have found it requires checking out its own license token. It would seem the class itself is checking out and retaining one license token. In fact if the import took out the last (or only) token left, no more optimisations can be started.
Perhaps this additional step might have been conceived to double check the successful import of mosek, but I have tested commenting it out and the functionality of MOSEK seems unaffected (see below). 

```python
with contextlib.suppress(ImportError):
    import mosek

    #with mosek.Task() as m: 
    #    m.optimize() #<- This checks out a license and retains it for as long as Model is in memory

    available_solvers.append("mosek")
```

Happy to put a pull request with the code above, if either my suggestion is sufficient or may start a more suitable solution idk. (maybe a different handling of context managers or a try/catch, etc). 

Edit: Tested with Python 3.10.2, Mosek 10.1.16 and Linopy 0.3.1

Best regards

jotaigna:
A follow up comment. Forcing the mosek `Env` in the context as per below also works, supporting the hypothesis that a license token is retained by the import. Perhaps this is more of an issue with how the creator/desctructor for the `Env()` object interacts with the garbage collector?
```python
with contextlib.suppress(ImportError):
    import mosek

    with mosek.Env() as m:
        t = m.Task()
        t.optimize()
        m.checkinall()

    available_solvers.append("mosek")
```

FabianHofmann:
[...]

PR #199, opened 2023-11-09
MOSEK ensure any license tokens are checked in after initialisation

jotaigna:
Proposed to fix issue #197 
Tested with Python 3.10.11 Mosek 10.1.16 and Linopy 0.31 in a Linux Ubuntu machine.


FabianHofmann:
great thanks @jotaigna

Issue #200, opened 2023-11-10
model.matrices.c is being coherrsed to int


dannyopts:
When accessing model.matrices.c the value is coherrsed to int (see failing test below)

This is because when c is created, it has a default value of 0, which is used to create the initial vector. 
```
create_vector(vars, ds.coeffs, fill_value=0, shape=shape)
```
switching this default to 0.0 fixes the issue. 

I will raise a PR to fix this now

```
def test_matrices_float_c():
    m = Model()

    x = m.add_variables(pd.Series([0, 0]), 1, name="x")
    m.add_objective(x * 1.5)

    c = m.matrices.c
    assert np.all(c == np.array([1.5, 1.5])) 
```
Fails with
>       assert np.all(c == np.array([1.5, 1.5]))
E       assert False
E        +  where False = <function all at 0x1044e5b30>(array([1, 1]) == array([1.5, 1.5])

dannyopts:
This causes a bug when using the direct api that cost coeffs are rounded to ints
>>> m = Model()
>>> x = m.add_variables(name="x")
>>> m.add_objective(0.1 * x)
>>> m.solve("highs", "direct")
>>> m.objective.value
0.0
>>> m.solve("highs", "lp")
>>> m.objective.value
nan

PR #201, opened 2023-11-10
set fill value to 0.0 to prevent cohersion to int for matrices.c

dannyopts:
Fixes https://github.com/PyPSA/linopy/issues/200


FabianHofmann (approved):
awesome, thanks @dannyopts

PR #203, opened 2023-11-13
Use os.makedirs to create parent dirs for solution_fn

dannyopts:
Before this fix the tests fail cbc does not work on OSx. 

On OSx mknod requires privledged access (need to run as sudo)
https://stackoverflow.com/questions/32115715/os-mknod-fails-on-macos, but even when running as sudo, the test still fails with an invalid argument error.  

I assume the mknod is there incase the parent dir of the solution_fn is not present, so I created a test for this case and found that glpk also fails if the solution_fn's parent dir does not exist so I added a fix for this.

### DIFF
--- a/linopy/solvers.py
+++ b/linopy/solvers.py
@@ -171,8 +171,7 @@ def run_cbc(
     if basis_fn:
         command += f"-basisO {basis_fn} "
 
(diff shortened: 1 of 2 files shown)


FabianHofmann (approved):
thanks @dannyopts

PR #212, opened 2023-12-07
Allow removing objective expression from the model

Irieo:
Add functionality to remove a linear expression of the objective from the model.
complements already existing functionality, such as `remove_variables()` and `remove_constraints()`
Useful for tutorials, and more generally, for any type of programming exercises when one wants to keep variables and parameters from `n.optimize.create_model()` but use own objective.

example use:

`n.optimize.create_model()`
`n.model.objective`

> LinearExpression: +70 Generator-p[now, gas]
> Sense: min
> Value: None

`n.model.remove_objective()`

> LinearExpression: +0
> Sense: min
> Value: None

closes https://github.com/PyPSA/linopy/issues/209

### DIFF
--- a/linopy/model.py
+++ b/linopy/model.py
@@ -666,6 +666,16 @@ class Model:
         """
         self.constraints.remove(name)
 
+    def remove_objective(self):
+        """
+        Remove the objective's linear expression from the model.
+
+        Returns
+        -------
+        None.
+        """
+        self.objective = Objective(LinearExpression(None, self), self)
+
     @property
     def continuous(self):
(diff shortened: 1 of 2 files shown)


FabianHofmann:
Awesome @Irieo, to increase the awesomeness, could you add a test for it, probably in test_model.py?

Irieo:
> Awesome @Irieo, to increase the awesomeness, could you add a test for it, probably in test_model.py?

indeed, let's keep the level of awesomeness pumped up. I added a test.

Issue #171, opened 2023-10-01
`linopy` does not pass quadratic objective terms to HiGHS


fneum:
When I use the quadratic_marginal_cost attribute in PyPSA in combination with HiGHS, I get different results than with Gurobi. HiGHS logs also state that it's solving an LP rather than a QP.

This is because HiGHS does not seem to get the quadratic terms in the objective function (although there seems to be code that should):

https://github.com/PyPSA/linopy/blob/74ce2d2eda87b9a4fe042a6dcc208e2af27b5469/linopy/io.py#L416-L422

FabianHofmann:
This is weird, as all the tests seem to be fine. Things like 

```python
m = linopy.Model()

lower = pd.Series(0, range(10))
x = m.add_variables(lower, name="x")
y = m.add_variables(lower, name="y")

m.add_constraints(x + y >= 10)

m.add_objective(- 2 * x + y + x * x)

m.solve(solver_name="highs")
```

yield the same result for highs and gurobi. and also the model definition seems totally fine when running test for pypsa networks (however it does not solve with highs). Did it explicitly say that it is trying to solve a LP in your case? and are the dual non-zero?

fneum:
I think I understand this now.

Öet's take a simple LP file with quadratic terms:

```
min

obj:

-4 x0 -6 x1 + [ + 4 x0 * x0 + 12 x1 * x1 ] / 2

s.t.

c0: +1 x0 +2 x1 >= +4

bounds

+0 <= x0 <= +4
+0 <= x1 <= +4
end
```

If we run

```py
import highspy
h = highspy.Highs()
h.readModel("test-highs.lp")
h.run()
```

we get the correct solution:

```
Running HiGHS 1.5.3 [date: 2023-05-16, git hash: 594fa5a9d-dirty]
Copyright (c) 2023 HiGHS under MIT licence terms
<HighsStatus.kOk: 0>
0, 16.000001, 0, 0.000281, 0.000000, 0, 0.000000, 0.000000
4, -0.071428, 1, 0.000331, 0.000000, 0, 0.000000, 1.000000
Model   status      : Optimal
QP ASM    iterations: 4
Objective value     : -7.1428571429e-02
HiGHS run time      :          0.00
```

However, if we specify "ipm" as solver:

```py
import highspy
h = highspy.Highs()
h.readModel("test-highs.lp")
h.setOptionValue("solver", "ipm")
h.run()
```

we get a different result:

```
[...]

PR #223, opened 2024-02-03
QP support for MOSEK, COPT, MindOpt

fneum:
MOSEK:
- The `solsta.dual_infeas_cer` should be interpreted as "infeasible_or_unbounded"
- An empty constraint list can lead to an `AttributeError` when reading out duals. This is handled like in the Gurobi interface.

COPT
- For reading out the solver status, need to handle the case where `model.type == "QP"`
- An empty constraint list can lead to an `AttributeError` when reading out duals. This is handled like in the Gurobi interface.
- The check in the test `test_quadratic_model` for the `y` variable was too restrictive. 10 is the lowest value `y` has to take in the cost-optimum, but increasing this value (which was what COPT did) yields the same objective value. Degeneracy.

MindOpt:
- According to the developers, MindOpt does not yet support reading in QPs via LP file, only via MPS file.
- An empty constraint list can lead to an `AttributeError` when reading out duals. This is handled like in the Gurobi interface.

HiGHS:
- Added warning regarding #171 

Tests:
- The QP test for unbounded problems was flawed. By reversing the sign of the objective function with quadratic terms, the problem became non-convex. This caused MOSEK to complain (matrix was not positive semidefinite) and was also the cause why "SCIP" needed special handling. I exchanged that with a new test case that is unbounded yet convex.

Environment:
- Was there a reason to not add `xpress` to the solvers in `setup.py`?

### DIFF
--- a/doc/release_notes.rst
+++ b/doc/release_notes.rst
@@ -1,8 +1,10 @@
 Release Notes
 =============
 
-.. Upcoming Release
-.. ----------------
+Upcoming Release
+----------------
+
+- Added support for QP problems with MOSEK and COPT.
 
 Version 0.3.3
 -------------
--- a/linopy/solvers.py
+++ b/linopy/solvers.py
@@ -383,6 +383,12 @@ def run_highs(
     if warmstart_fn:
         logger.warning("Warmstart not available with HiGHS solver. Ignore argument.")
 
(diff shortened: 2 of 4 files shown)


FabianHofmann:
This is awesome, thanks @fneum. The xpress installation does not seem to be available on mac for python 3.11. Perhaps we can find a way to ignore it for mac.

PR #224, opened 2024-02-05
make it possible to suppress solver output when running CBC and GLPK

maurerle:
Currently, the solver output is printed all the time to stdout using `print` when using CBC or GLPK.
This should be fixed, by also using the logging module.

I did set this to INFO, so that users relying on this output can still use it without going to debug logs.

It would be very cool to have release with this change afterwards

### DIFF
--- a/doc/release_notes.rst
+++ b/doc/release_notes.rst
@@ -4,6 +4,11 @@ Release Notes
 .. Upcoming Release
 .. ----------------
 
+Version 0.3.4
+-------------
+
+* Solver output of CBC and GLPK is sent to logging with level INFO instead of stdout
+
 Version 0.3.3
 -------------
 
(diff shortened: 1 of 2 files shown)


maurerle:
It does not look like I can do anything about 3.9-macos and 3.11-macos failing - it says `Fatal Python error: Illegal instruction` coming from `pyscipopt` solving with `scip`

FabianHofmann:
Great! thanks @maurerle, would you mind putting a short note to the release notes? CI problem seemed to be due to a wrong installation (is working now).

FabianHofmann (approved):

maurerle:
Sure, is it fine like this @FabianHofmann ?

FabianHofmann:
awesome, thanks!

maurerle:
@FabianHofmann could you please publish a release with this version soon?

FabianHofmann:
@maurerle already happened :) conda should be available within the next hour

maurerle:
Would you also add the release here:
https://github.com/PyPSA/linopy/releases
This is also missing for 0.3.3

This would make it possible to easily see the changes since the last release :)

PR #226, opened 2024-02-05
Add quadratic objective support and 'direct' mode for MOSEK

ulfworsoe:
This PR adds a `to_mosekpy` (similar to `to_gurobipy`) for passing a model to MOSEK without files. 

This should add support for quadratic objective terms.

I have attempted to fix support for reading solution (warmstart) files and disabled reading basis files since these are not directly supported by MOSEK.

### DIFF
--- a/linopy/io.py
+++ b/linopy/io.py
@@ -307,7 +307,7 @@ def to_file(m, fn, integer_label="general"):
     return fn
 
 
-def to_mosekpy(model, task):
+def to_mosekpy(model, task=None):
     """
     Export model to MOSEK.
 
@@ -324,6 +324,9 @@ def to_mosekpy(model, task):
     """
 
     import mosek
+
+    if task is None: 
+        task = mosek.Task()
 
     task.appendvars(model.nvars)
     task.appendcons(model.ncons)
--- a/linopy/io.py
+++ b/linopy/io.py
@@ -334,7 +334,6 @@ def to_mosekpy(model, task):
 
     ## Variables
 
-
     bkx = [
         (
             (mosek.boundkey.ra if l < u else mosek.boundkey.fx)
--- a/linopy/io.py
+++ b/linopy/io.py
@@ -334,14 +334,15 @@ def to_mosekpy(model, task):
 
     ## Variables
 
+
     bkx = [
         (
             (mosek.boundkey.ra if l < u else mosek.boundkey.fx)
             if u < np.inf
-            else mosek.boundkey.up
+            else mosek.boundkey.lo
         )
         if (l > -np.inf)
-        else (mosek.boundkey.lo if (u < np.inf) else mosek.boundkey.fr)
+        else (mosek.boundkey.up if (u < np.inf) else mosek.boundkey.fr)
         for (l, u) in zip(M.lb, M.ub)
     ]
     blx = [b if b > -np.inf else 0.0 for b in M.lb]
@@ -355,18 +356,18 @@ def to_mosekpy(model, task):
         for i, n in enumerate(names):
             task.putconname(i, n)
         bkc = [
-            mosek.boundkey.lo
+            mosek.boundkey.up
             if s == "<"
-            else (mosek.boundkey.up if s == ">" else mosek.boundkey.fx)
+            else (mosek.boundkey.lo if s == ">" else mosek.boundkey.fx)
             for s in M.sense
         ]
-        blx = M.b
-        bux = M.b
+        blc = M.b
+        buc = M.b
         A = M.A.tocsr()
         task.putarowslice(
             0, model.ncons, A.indptr[:-1], A.indptr[1:], A.indices, A.data
         )
-        task.putconboundslice(0, model.ncons, bkx, blx, bux)
+        task.putconboundslice(0, model.ncons, bkc, blc, buc)
 
     ## Objective
     if model.is_quadratic:
@@ -378,7 +379,6 @@ def to_mosekpy(model, task):
         task.putobjsense(mosek.objsense.maximize)
     else:
         task.putobjsense(mosek.objsense.minimize)
-
     return task
 
 
--- a/linopy/io.py
+++ b/linopy/io.py
@@ -13,7 +13,7 @@ import numpy as np
 import pandas as pd
 import xarray as xr
 from numpy import asarray, concatenate, ones_like, zeros_like
-from scipy.sparse import csc_matrix, triu
+from scipy.sparse import csc_matrix, triu, tril
 from tqdm import tqdm
 
 from linopy import solvers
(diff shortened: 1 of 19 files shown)


linopy/solvers.py
```diff
@@ -884,17 +890,36 @@ def run_mosek(
                 m.linkfiletostream(mosek.streamtype.log, log_fn, 0)
 
             if warmstart_fn:
-                m.readdata(warmstart_fn)
+                # Gurobi solution format is not supported by MOSEK.
+                # What is the warmstart file? A Gurobi .sol file? The gurobi
+                # docs are a bit sparse on the format. Does it include dual
+                # information?
```
FabianHofmann, line 896: This is probably a left-over, can you take that out? 
A warmstart file is a simplex basis which can be used by some solvers as a starting point for the optimization. It is stored as a `.bas` file. However I am not sure whether `xpress` can handle these.  Did you try out the original call?

FabianHofmann (commented):
Great @ulfworsoe, thanks for the PR! could you add the implementation to the test routine? Just add a 

```
if "mosek" in available_solvers:
    params.append(("mosek", "direct"))
```
in the beginning of the test_optimization script https://github.com/PyPSA/linopy/blob/dd54ec19c5851ce12e26c08acf113ae7cab50519/test/test_optimization.py#L29

FabianHofmann:
And a hint: if you install pre-commit locally, the pre-commit bot does not interfere with your PR, as all stylistic changes are already applied before committing.

ulfworsoe:
Ah, that's what was happening. I should probably do that.

linopy/solvers.py
ulfworsoe: The original call used `task.readdata`. I'll check, but I am reasonably sure that it will only read problem files (LP, MPS, etc). MOSEK does have a .bas file but that is an entirely different format.
But I implemented a `.sol` reader here so I'll, have to fix that too.

FabianHofmann:
Hey @ulfworsoe, this looks good. I will merge #223 before, and hope this does not lead to merge conflicts.

ulfworsoe:
@FabianHofmann  I can see that the coverage tests fail. My guess is that the tests don't run for MOSEK at all because there is no license file. We have a similar issue in other places (the MOSEK Julia API, for example), and what we do in these cases is to use http://solve.mosek.com/web/ instead of solving locally. This doesn't require a license for small-ish problems.
I am not sure how to do this, though, as it means calling `task.putoptserverhost()` somewhere.

ulfworsoe:
@FabianHofmann  i think it is ready. The tests actually run and pass now.

linopy/solvers.py
```diff
@@ -845,6 +863,45 @@ def get_solver_solution() -> Solution:
     return Result(status, solution, m)
 
 
+mosek_bas_re = re.compile(r" (XL|XU)\s+([^ \t]+)\s+([^ \t]+)| (LL|UL|BS)\s+([^ \t]+)")
+
+
+def run_mosek_remote(
```
FabianHofmann, line 869: I would honestly avoid this function. This contains one parameter change and does not represent a dedicated solver. I would instead set the parameter as a default in the standard `run_mosek` and let people change it to their needs.

linopy/solvers.py
```diff
@@ -873,24 +932,28 @@ def run_mosek(
...
     problem_fn = model.to_file(problem_fn)
 
-    problem_fn = maybe_convert_path(problem_fn)
+    if io_api != "direct" and io_api is not None:
+        problem_fn = maybe_convert_path(problem_fn)
     log_fn = maybe_convert_path(log_fn)
-    warmstart_fn = maybe_convert_path(warmstart_fn)
-    basis_fn = maybe_convert_path(basis_fn)
+    # warmstart_fn = maybe_convert_path(warmstart_fn)
+    # basis_fn = maybe_convert_path(basis_fn)
```
FabianHofmann, line 946: this should still be fine being uncommented or?

test/test_optimization.py
```diff
@@ -27,6 +27,13 @@
 if "highs" in available_solvers:
     params.append(("highs", "direct"))
 
+if "mosek" in available_solvers:
+    params.append(("mosek", "direct"))
+    params.append(("mosek", "lp"))
+elif "mosek_remote" in available_solvers:
+    params.append(("mosek_remote", "direct"))
+    params.append(("mosek_remote", "lp"))
+
```
FabianHofmann, line 36: see comment above

linopy/io.py
```diff
@@ -307,6 +307,93 @@ def to_file(m, fn, integer_label="general"):
...
+
+    if task is None:
+        task = mosek.Task()
+
+    task.appendvars(model.nvars)
+    task.appendcons(model.ncons)
+
+    M = model.matrices
+    for j, n in enumerate(("x" + M.vlabels.astype(str).astype(object))):
+        task.putvarname(j, n)
```
FabianHofmann, line 336: A pity that mosekpy does not support parallelized assignments. This will take quite long for large problems, but it seems that's the way to go.

linopy/io.py
```diff
@@ -307,6 +307,93 @@ def to_file(m, fn, integer_label="general"):
...
+            (
+                (mosek.boundkey.ra if l < u else mosek.boundkey.fx)
+                if u < np.inf
+                else mosek.boundkey.lo
+            )
+            if (l > -np.inf)
+            else (mosek.boundkey.up if (u < np.inf) else mosek.boundkey.fr)
+        )
+        for (l, u) in zip(M.lb, M.ub)
+    ]
```
FabianHofmann, line 351: Could you short say what is happening here? It's a bit hard to read for me.

FabianHofmann (changes requested):
@ulfworsoe thank you for the huge work! looks almost ready, just some minor points. Sorry that this takes so long. But ones it's there it will be quite stable.

linopy/solvers.py
ulfworsoe: I agree - this is a hack.. The problem is that I don't like setting remote optimization as the default, but I need to use it to make the tests run.

linopy/solvers.py
ulfworsoe: Done

linopy/io.py
ulfworsoe: I have inserted a comment now. MOSEK used a (boundkey,lb,ub) triplet to define a bound to avoid using `inf`. Here I compute the bound keys based on the bound values.

linopy/solvers.py
ulfworsoe: I'll set the default to remote optimization, but I would like that to be a temporary fix only.

ulfworsoe: I have removed the `mosek_remote`. This means that tests won't run in the CI, so I'd like to find a solution around that. What do you do for Gurobi? Does the CI image contain a license?

ulfworsoe:
@FabianHofmann I have removed the `mosek_remote`, but this means that MOSEK will not run in CI. What do you do for Gurobi? Does the CI image contain a license file?

FabianHofmann:
@ulfworsoe for some solvers I am keeping a github secret for the CI. Is something similar possible for mosek?

FabianHofmann:
@ulfworsoe it seems I have mosek license to the github! So if you don't mind I would pull everything into another branch to properly run the CI and then we can merge itl

ulfworsoe:
> @ulfworsoe it seems I have mosek license to the github! So if you don't mind I would pull everything into another branch to properly run the CI and then we can merge itl

Please do. I'm working on a more permanent solution for using licenses in github CI so they won't have to be updated by hand regularly.

PR #238, opened 2024-02-27
Better doc by explicitly using __qualname__ instead of __repr__()

leuchtum:
# Description
By explicitly using the `__qualname__` string of an object (introduced in PEP 3155), the documentation does not include the `<...>` brackets with the memory address. 

# Screenshots
Here`Variable.assign_attrs` as an example, but more methods are affected.

## Screenshot documentation BEFORE changes
Bildschirmfoto 2024-02-27 um 22 48 29

## Screenshot documentation AFTER changes
Bildschirmfoto 2024-02-27 um 22 48 14


FabianHofmann:
great, thanks @leuchtum

PR #251, opened 2024-03-10
Matmul and dot operation

FabianHofmann:
**New Features**

* Variables and LinearExpressions now have a new method `dot`, which allows computing the dot product of two objects. This multiplies objects and sums over common dimensions.

* The matmul operator `@`, which runs the `dot` operation, is now supported for Variables and LinearExpression.

**Bugfixes**

* The multiplication of two linear expression with non-zero constants led to wrong results of the cross terms. Given the multiplication `(v1 + c1)  * (v2 + c2)` with `v` being a variable and `c` a constant, the operation did not calculate the cross terms `v1 * c2 + v2 * c1`. This is fixed now.


hcoste:
Hello FabianHofmann, thanks for the feature. Could your provide an example of dot product between a linopy variable (vector) and a constant matrix ? It seems this is not yet possible but I wanted to have your opinion if possible.

KR

PR #267, opened 2024-04-08
Added the new pdlp solver (highs > 1.7) to the description, added warning

Cellophil:
MIP warning probably relevant to most users; Highs solver is a bit special there, it just ignores integrality without notice.


Cellophil:
I'm not sure what the codecov does

PR #269, opened 2024-04-09
Add a subsection for modification of variable bounds in docs

fwitte:
I was trying to modify variable bounds instead of creating additional constraints for already instantiated variables and could not find a subsection in the respective chapter. It does not work with something like `variable.loc[slice, other_slice].lower = 42` and I did not immediately understand why, so I thought a hint on how to do this might be useful.

If this is bad practice, I would be happy to get a rejection of the PR (and also be curious why it would be :)).

Thank you!

### DIFF
--- a/examples/creating-variables.ipynb
+++ b/examples/creating-variables.ipynb
@@ -348,7 +348,7 @@
    "outputs": [],
    "source": [
     "other = Model(force_dim_names=True)\n",
-    "try: \n",
+    "try:\n",
     "    other.add_variables(lower=lower, coords=coords)\n",
     "except ValueError as e:\n",
     "    print(\"This raised an error:\", e)"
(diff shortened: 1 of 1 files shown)


fwitte:
I blindly overlook the section, I was only reading the variables one... :D

FabianHofmann:
@fwitte thank you for your contribution. Setting the bounds on a selection of variables does not work, as the `.loc` function returns a copy of the original data. Therefore, the bound will only set to the copy and not the variable selection itself. Would you mind adding a note there, that this case is not supported (yet)?

FabianHofmann:
> I blindly overlook the section, I was only reading the variables one... Nevertheless, including how to slice `variable.lower.loc[slice, other]` instead of `variable.loc[slice, other].lower` might be useful

no worries

fwitte:
It does work, if you access the `xarray.DataArray` like this: `variable.lower.loc[slice, other]`

fwitte:
I added the hint and explanation, which of both version work. It is now in the Variables section. If this is of interest, let me know if I should reopen the PR and/or move the information to the Modifying the Model section.

Have a nice day

Issue #271, opened 2024-04-10
Explore PytOptInferface for writing matrix over native C API


pz-max:
## Background
Linopy either constructs a big matrix itself and stores it in a standardized file format such as the “LP” or the “MPS” file format or writes the matrix through direct **Python** solver interfaces. Linopy uses xarray and, therefore, has [good memory and speed performance](https://linopy.readthedocs.io/en/latest/benchmark.html) in creating the matrix/lp files. The syntax is also [easier to read](https://linopy.readthedocs.io/en/latest/syntax.html) compared to Jump.

[PyOptInterface](https://github.com/metab0t/PyOptInterface) was just released now (April 2024) by @metab0t and its [benchmark](https://metab0t.github.io/PyOptInterface/benchmark.html) shows that the speed in constructing the matrix is comparable to Jump.

## Explore
Can we use PyOptInterface within Linopy to speed up the matrix construction?

metab0t:
Thanks for your attention! @pz-max
The design of PyOptInterface is similar to the [direct mode](https://jump.dev/JuMP.jl/stable/manual/models/#Direct-mode) of JuMP.jl. It does not store the model internally and only records the mapping between variable/constraint and column/row index.
Do you plan to use PyOptInterface in place of direct Python solver interfaces in linopy?

pz-max:
@metab0t  We would be keen to explore possibilities. Would you be free for a 30-minute exchange with @FabianHofmann (linopy maintainer) and me? If yes, could you ping me per email: max.parzen(at)openenergytransition.org with few time suggestions that are EU friendly? :)

metab0t:
@pz-max I have emailed you about the discussion.

aurelije:
So this has potential of abstracting things in io module of linopy (where magic of translating linopy model into solver specific one happens)?

metab0t:
After profiling, I think that the bottleneck is in the `MatrixAccessor` class to build matrix instead of gurobipy.
```python
import time
from numpy import arange
from linopy import Model

import pyoptinterface as poi
from pyoptinterface import gurobi

def access_matrix(m):
    M = m.matrices
    vlabels = M.vlabels
    names = "x" + vlabels.astype(str).astype(object)
    vtypes = M.vtypes
    lb = M.lb
    ub = M.ub

    if m.is_quadratic:
        Q = M.Q

    c = M.c

    A = M.A
    clabels = M.clabels
    sense = M.sense
    b = M.b
[...]

PR #272, opened 2024-04-13
Update JuMP syntax in syntax.rst

odow:
I saw @pz-max's comment in https://github.com/PyPSA/linopy/issues/271.

This PR just updates to match the benchmark implementation:
https://github.com/PyPSA/linopy/blob/5d61f958a9b805d4c41ecb606a223d1125fa67c0/benchmark/scripts/benchmark_jump.jl#L9-L18 

It's fair that most people probably write out the scalar `I` and `j` version though, so up to you whether you want to merge this.


FabianHofmann (approved):
Great, thank you

PR #275, opened 2024-04-29
Support for integer/binary vars for MOSEK

ulfworsoe:
Add support for integer/binary vars for MOSEK solver direct mode.


FabianHofmann (approved):
thanks again for your PR @ulfworsoe

PR #279, opened 2024-05-03
replace setup.py by pyproject.yaml

FabianHofmann:


maurerle:
I would add the line `requires-python = ">=3.9"`  which limits the python version as before and automagically adds the matching python attributes to the pypi release.

Starting with python 3.9 one can just add
```
from importlib.metadata import version
__version__ = version("linopy")
```
to this file: https://github.com/PyPSA/linopy/blob/master/linopy/__init__.py

Which takes the entry from the pyproject.toml and removes double book keeping.
Instead of the version.py it seems that https://pypi.org/project/setuptools-scm/ also supports setting the value in the pyproject.toml :)

FabianHofmann:
good point @maurerle, I have added a automatic creation of the version.py in here https://github.com/PyPSA/linopy/pull/279/files#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R71-R72

that should do the think right?

pyproject.toml
```diff
@@ -1,5 +1,72 @@
 [build-system]
-requires = ["setuptools>=45", "wheel", "setuptools_scm>=6.2<8"]
+requires = ["setuptools", "wheel"]
```
maurerle, line 2:
```suggestion
requires = ["setuptools>=60", "setuptools-scm>=8.0"]
```

pyproject.toml
```diff
@@ -1,5 +1,72 @@
...
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "linopy"
+version = "0.3.9"
```
maurerle, line 7:
```suggestion
dynamic = ["version"]
```

pyproject.toml
```diff
@@ -1,5 +1,72 @@
...
+    "cplex",
+    "mosek",
+    "mindoptpy",
+    "coptpy",
+    "xpress; platform_system != 'Darwin' and python_version < '3.11'",
+    "pyscipopt; platform_system != 'Darwin'",
+]
 
 [tool.setuptools_scm]
 write_to = "linopy/version.py"
```
maurerle, line 72: You can remove this line as well as the `version.py` file as setuptools-scm handles setting the correct version internally.

To also provide version information in the conventional `linopy.__version__` you then only need to add my snippet from the comment before :)

FabianHofmann:
@maurerle thanks for your comments! but now I am struggling to understand where the version is defined. does is take it automatically from the git tags?

maurerle:
Yes exactly. If your current commit is a tag, the version is the tag.
Otherwise you get something like `1.1.4.dev0+g9599efa.d20240503` which contains the latest tag on your branch as well as the latest commit id as well as the date it was installed - making it easy to see what the currently used state is.
If you have uncommitted changes, this is not reflected though.

I think that is the usual behavior when using `setuptools_scm` (just like before?)

FabianHofmann:
Wonderful! Yes, that's the same behaviour as before. Thanks @maurerle, for being so supportive!

maurerle (approved):
You can remove the `write_to = "linopy/version.py"` as the generated file is not used anymore and would only confuse (but it does not hurt either).

Other than that, this looks good! :)

pyproject.toml
FabianHofmann: forgot that one, thanks

FabianHofmann:
@maurerle looks like the write_to = version.py is required (at least seeing from my local test)

PR #281, opened 2024-05-03
Improve run_gurobi to wait for available token

tgi-climact:
When using a Gurobi floating token server license or a web single-use license, a program that calls Gurobi Optimizer must obtain a token from a Gurobi token server before it can solve an optimization model. When no token is available, an exception `gurobipy.GurobiError` is raised (source : https://support.gurobi.com/hc/en-us/articles/360029879251-How-do-I-check-the-availability-of-floating-license-tokens).

To avoid Gurobi Optimizer throwing an error, I suggest to optionally wait for an available token. This is the purpose of this PR. The waiting time (in seconds) can be configured in the solver options.

Side note : This enables scenarios in PyPSA-Eur.

### DIFF
--- a/doc/release_notes.rst
+++ b/doc/release_notes.rst
@@ -14,6 +14,8 @@ Upcoming Version
 
 * Up to now the `rhs` argument in the `add_constraints` function was not supporting an expression as an input type. This is now added.
 
+* Improve `run_gurobi` to optionally wait for an available token.
+
 Version 0.3.8
 -------------
 
--- a/linopy/solvers.py
+++ b/linopy/solvers.py
@@ -568,6 +568,25 @@ def run_cplex(
     return Result(status, solution, m)
 
 
+def start_gurobi_env(wait=60):
+    """
+    Try (forever!) to create an environment. Return the environment once started. Waiting time is in seconds.
+    """
+    # see https://support.gurobi.com/hc/en-us/articles/360029879251-How-do-I-check-the-availability-of-floating-license-tokens
+    import time
+
+    while True:
+        try:
+            # Attempt to construct a Gurobi environment
+            env = gurobipy.Env()
+            logging.debug("Gurobi: Token retrieved!")
+            return env
+        except gurobipy.GurobiError:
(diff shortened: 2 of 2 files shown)


FabianHofmann:
thanks @tgi-climact, perhaps 60 secs is a bit long?

tgi-climact:
> thanks @tgi-climact, perhaps 60 secs is a bit long?

Depending on the optimisation in progress, it could be very long or very short. I would like to avoid spam on the Gurobi servers. What would be your suggestion?

FabianHofmann:
> Depending on the optimisation in progress, it could be very long or very short. I would like to avoid spam on the Gurobi servers. What would be your suggestion?

okay I see, then the 60 secs make sense I guess.

dannyopts:
A couple of thoughts:

There are many other things that can go wrong when creating an environment, not just token aquisition, for example:
GurobiError: Server aborted the SSL handshake
For this reason, I would atleast log the error, and not assume that it is the token causing failure. 

Similarly, I would increase the logging level from debug, since if you are stuck in this loop, it would be helpful to see it in logs than just hang - most production systems only log info level and debug logs are only logged at when you are investigating a bug.

Also I would personally prefer it not to infinitely retry. I think it makes more sense to try some number of times and then give up (throw an error) unless very explicitly told otherwise. I'm imagining a fleet of servers all in a busy loop trying to reach a non-existent gurobi token server and having to all be rebooted.

ollie-bell:
I stumbled upon this as I'm currently up against the same problem. In my opinion, this should just be something that users implement themselves, which they can do via the `env` argument to `model.solve`. I don't think linopy should be responsible for managing this.

FabianHofmann:
@tgi-climact given the non-clear picture and the outdated proposal (which is also on me reviewing to late), do you think we should go forward with this pr or close it?

tgi-climact:
@FabianHofmann Let's close it for now. I don't have a recent application of this to build on. However, if it proves useful, we can always reopen an improved PR anyway.

Issue #283, opened 2024-05-08
Improve workflow for releases


leuchtum:
Howdy folks,

just a small suggestion for improvement:

I recently updated our internal dependencies (linopy is one of them) and stumbled over the version number of linopy. On 03.05.2024 linopy 0.3.9 was released on PyPI. However, there is no mention of this here on GitHub, the release is still 0.3.8... I suggest a more consistent approach in the CD. Especially the release notes for 0.3.9 I really had to search for.

Otherwise thanks, linopy rocks!

FabianHofmann:
Thanks for notifying. I always forget about the GH release. I don't know if there is an automatic solution to make a release whenever there is a new tag. But anyway, thanks

PR #292, opened 2024-05-17
Workflow for auto releasing on GH

leuchtum:
Continuing on #283 I suggest to add a workflow for auto releasing on GitHub. This can be done by using the action `gh-release`. One can configure it, so that it auto generates the release notes as well (Same outcome as using the button from the web GUI).

I marked this as draft for two reasons:
- Need to discuss if combine this into the existing workflow `.github/workflows/deploy.yml`
- The job in this PR filters the tag, as linopy uses `v*.*.*` as the tag for releases and you don't want to make a release for every tag. Need to discuss if adding this filter to the `.github/workflows/deploy.yml` as well.

### DIFF
--- a/.github/workflows/gh-release.yml
+++ b/.github/workflows/gh-release.yml
@@ -1,0 +1,17 @@
+name: Release on GitHub
+
+on:
+  push:
+    tags:
+      - v*.*.*
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


leuchtum:
Checks failed due to codecov, same problem as in #290

FabianHofmann (approved):
Great, thank you @leuchtum

PR #298, opened 2024-05-28
ci: fix potential setuptools_scm issues in gh actions

lkstrp:
In reference to https://github.com/PyPSA/PyPSA/pull/907 #279 and #292 I just add a fetch for the tags here so that `setuptools_scm` can detect the version correctly. Right now that doesn't lead to errors, so only to avoid the potential bug search in the future (which I just had).

Fore more details see [here](https://stackoverflow.com/questions/62968271/python-automatic-versioning-not-happening-while-running-in-github-actions).

### DIFF
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -26,6 +26,10 @@ jobs:
         - windows-latest
     steps:
     - uses: actions/checkout@v4
+
+    - name: Fetch all history for all tags and branches (necessary for setuptools_scm)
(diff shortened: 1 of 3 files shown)


FabianHofmann:
flake 8 does not like some syntax here https://github.com/PyPSA/linopy/actions/runs/9273219169/job/25512792210?pr=298 (?)

FabianHofmann:
don't know what is going on with codecov, I am closing and opening the pr in the hope that codecov with work then

PR #305, opened 2024-06-14
Forward Mosek log to stdout

aszekMosek:
The documentation for ``Model.solve()`` says

*log_fn (path_like, optional) – Path of the logging file which is written out during the process. The default None results in the no log file, hence all solver outputs are piped to the python repl.*

This change makes the Mosek behavior consistent with the second part of the description, that is *The default None results in ... all solver outputs are piped to the python repl*. Currently in Mosek if ``log_fn is None`` then nothing gets printed.


linopy/solvers.py
```diff
@@ -947,6 +947,8 @@ def run_mosek(
 
             if log_fn is not None:
                 m.linkfiletostream(mosek.streamtype.log, log_fn, 0)
+            else:
+                m.set_Stream(mosek.streamtype.log, lambda x: sys.stdout.write(x))
```
fneum, line 951:
```suggestion
                m.set_Stream(mosek.streamtype.log, sys.stdout.write)
```
This likely also works

PR #310, opened 2024-06-27
Annotate typing

### LINKED ISSUE #185: Annotate Typing
It seems to be a good way to use [monkeytype](https://github.com/Instagram/MonkeyType) for this:

```bash
pip install monkeytype

monkeytype run -m pytest linopy --doctest-modules test 

monkeytype apply linopy.<module>

# see all modules with monkeytype list-modules

```

one needs to solve circular imports as well. 

Afterwards we can add mypy to pre-commit 

```
-   repo: https://github.com/pre-commit/mirrors-mypy
    rev: v1.6.1  # Use the sha / tag you want to point at
    hooks:
    -   id: mypy
```

### LINKED ISSUE #303: `xarray` DeprecationWarning: Deleting a single level of a MultiIndex is deprecated. Previously
Linopy throws a `xarray` DeprecationWarning quite a few times. If raised as exceptions, 89 tests fail. Mainly in `test/test_linear_expression.py`.

`../site-packages/xarray/core/dataarray.py:3146: DeprecationWarning: Deleting a single level of a MultiIndex is deprecated. Previously, this deleted all levels of a MultiIndex. Please also drop the following variables: {'z'} to avoid an error in the future.`

Needs a deeper dive, but here are some thoughts:
Only happens with multi index actions and when arrays are assigned to a dataset with a new reference. See some examples below. During the merge of two objects, xarray is dropping conflicting coords. But deleting them leads to further problems. I'm also not sure whether this is an xarray issue, because the code can only be adapted in a rather unpleasant way. 

Links:
[Dropping a MultiIndex variable raises an error after explicit indexes refactor · Issue #6505 · pydata/xarray](https://github.com/pydata/xarray/issues/6505)
[Explicit indexes by benbovy · Pull Request #5692 · pydata/xarray](https://github.com/pydata/xarray/pull/5692)
[python - Why cannot I add a DataArray to an existing Dataset with a MultiIndex? - Stack Overflow](https://stackoverflow.com/questions/76471238/why-cannot-i-add-a-dataarray-to-an-existing-dataset-with-a-multiindex)

Occurs in:
https://github.com/PyPSA/linopy/blob/829453055d253dbf80b17247550cb8d99ac15295/linopy/variables.py#L153
https://github.com/PyPSA/linopy/blob/829453055d253dbf80b17247550cb8d99ac15295/linopy/model.py#L457-L476

FabianHofmann:
Add type annotations  & add MyPy to pre-commit hooks. This is quite a large change, but the functional change is kept to a minimum. 

closes #303 
closes #185

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,8 @@ doc/api
 .vscode
 Highs.log
 paper/
+monkeytype.sqlite3
+
 
 benchmark/*.pdf
 benchmark/benchmarks
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -39,3 +39,8 @@ repos:
     hooks:
     -   id: jupyter-notebook-cleanup
         exclude: examples/solve-on-remote.ipynb
+-   repo: https://github.com/pre-commit/mirrors-mypy
+    rev: v1.10.0
+    hooks:
+    -   id: mypy
+        args: [linopy]
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -40,8 +40,8 @@ repos:
     -   id: jupyter-notebook-cleanup
         exclude: examples/solve-on-remote.ipynb
 -   repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v1.10.0
+    rev: v1.10.1
     hooks:
     -   id: mypy
         args: [--explicit-package-bases]
-        additional_dependencies: [numpy, pandas, xarray, scipy]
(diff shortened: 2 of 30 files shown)


FabianHofmann:
@lkstrp could I kindly ask you to take over fixing the pre-commit bot. It looks like the pre-commit CI is sort of stuck with another set of (cached ?) arguments. Locally it works fine. What I want in the end is to run mypy checks on `linopy/` and `test/`

FabianHofmann:
@lkstrp awesome thanks

PR #313, opened 2024-07-08
ci: use checkout with fetch-depth 0

afuetterer:
See: https://github.com/PyPSA/PyPSA/pull/954

### DIFF
--- a/.github/workflows/gh-release.yml
+++ b/.github/workflows/gh-release.yml
@@ -10,9 +10,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
-
-      - name: Fetch all history for all tags and branches (necessary for setuptools_scm)
-        run: git fetch --prune --unshallow
+        with:
+          fetch-depth: 0 # Fetch all history for all tags and branches (necessary for setuptools_scm)
 
       - name: Release
         uses: softprops/action-gh-release@v2
--- a/.github/workflows/gh-release.yml
+++ b/.github/workflows/gh-release.yml
@@ -15,6 +15,5 @@ jobs:
 
       - name: Release
         uses: softprops/action-gh-release@v2
-        if: startsWith(github.ref, 'refs/tags/')
         with:
           generate_release_notes: true
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -26,9 +26,8 @@ jobs:
         - windows-latest
     steps:
     - uses: actions/checkout@v4
-
-    - name: Fetch all history for all tags and branches (necessary for setuptools_scm)
-      run: git fetch --prune --unshallow
+      with:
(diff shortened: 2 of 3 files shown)


.github/workflows/gh-release.yml
```diff
@@ -10,9 +10,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
-
-      - name: Fetch all history for all tags and branches (necessary for setuptools_scm)
-        run: git fetch --prune --unshallow
+        with:
+          fetch-depth: 0 # Fetch all history for all tags and branches (necessary for setuptools_scm)
```
afuetterer, line 14: Actually I think you don't need all the commits here, right?
Maybe it was copy-and-pasted here?

afuetterer:
Two questions.

Regarding gh-release workflow:
- In my opinion this line is redundant, because the workflow only runs, when a tag is pushed:
https://github.com/PyPSA/linopy/blob/20a6dcdcd8dc4221c8f417e219075da0524385f7/.github/workflows/gh-release.yml#L19
- Can we remove it?

Regarding deploy workflow:
- The workflow runs on push, so every commit builds the package, but only on a tag it is deployed.
- Is this intended? Or should deploy run on tag only?
https://github.com/PyPSA/linopy/blob/20a6dcdcd8dc4221c8f417e219075da0524385f7/.github/workflows/deploy.yml#L37
- maybe you could use [build-and-inspect-python-package](https://github.com/hynek/build-and-inspect-python-package) in the ci workflow? It tests if the package is "buildable" and runs a few checks on the files, metadata etc.

FabianHofmann:
thanks for the PR and raising the issue @afuetterer, I think @lkstrp knows :)

afuetterer:
Example of "baipp" workflow: https://github.com/hynek/structlog/blob/main/.github/workflows/ci.yml

Example output: https://github.com/hynek/build-and-inspect-python-package/actions/runs/9817023209

afuetterer:
What does that mean for this PR?

lkstrp:
You can apply everything proposed in [here](https://github.com/PyPSA/linopy/pull/313#issuecomment-2213139573).

.github/workflows/gh-release.yml
afuetterer: How about this one? It should work without pulling all commits, right?

FabianHofmann: @lkstrp let's try it out?

lkstrp: Think so, let's go. 
But I would also merge `gh-release.yml` with `deploy.yml`.

afuetterer:
> @afuetterer ready to merge?

Yes. Will open another PR soon.

.github/workflows/gh-release.yml
afuetterer: Want to do it in this PR or another one?

FabianHofmann:
great thanks @afuetterer

PR #319, opened 2024-07-09
fix module not found error highspy

michalmarszal:
0.3.12 released on pypi.org forces to have highspy installed, without having it added to dependency in pyproject.toml. I do not see point of installing HiGHS when I am using the other solver.
This is cased by small import in linopy/io.py:20 which is used just for typing hint


linopy/io.py
```diff
@@ -729,7 +728,7 @@ def to_gurobipy(m: Model, env: Union[Any, None] = None) -> Any:
     return model
 
 
-def to_highspy(m: Model) -> Highs:
```
lkstrp, line 732: Could you use a the `TYPE_CHECKING` check instead?
As done in here
https://github.com/PyPSA/linopy/blob/f25f208daaff1ef5f5aae916f937eca4df0685fa/linopy/constraints.py#L61-L62

lkstrp (changes requested):

linopy/io.py
michalmarszal: I was thing about it. It should work provided one uses dev set of optional dependencies, which I assume developers are using.

lkstrp: It should work without optional dependencies as well, when wrapped with the type checking flag

michalmarszal: Done

lkstrp (approved):
Thank you @michalmarszal !

PR #322, opened 2024-07-09
Update xarray package version from "xarray>=2023.9.0" to "xarray>=2024.2.0"

kruksik-dev:
Method called is_dict_like first time appears at xarray/namedarray/utils.py unit version 2024.2.0. So each version below 2024.2.0 is depricated.


FabianHofmann (approved):
great, thanks @kruksik-dev

PR #327, opened 2024-07-16
Fix io termination condition

FabianHofmann:


siddharth-krishna (approved):
Works for me, thanks :)

PR #332, opened 2024-07-24
Remove numpy pin

glatterf42:
Gurobipy v11.0.3 was released on July 16 and seems to be compatible with numpy >= 2.0.0. At least locally, running the tests via `pytest .` and an editable install from this branch with numpy 2.0.1 and gurobipy 11.0.3, all tests are passing :)

The only thing I can't quite make sense of is this warning:
```
test/test_io.py::test_model_to_netcdf
  <frozen importlib._bootstrap>:488: RuntimeWarning: numpy.ndarray size changed, may indicate binary incompatibility. Expected 16 from C header, got 96 from PyObject
```


lkstrp:
[Some tests](https://github.com/PyPSA/linopy/actions/runs/10073598425/job/27847997608?pr=332) on Windows also fail. This is not urgent, but feel free to dig in. 
Numpy deprecation with [NPY201](https://docs.astral.sh/ruff/rules/numpy2-deprecation/) is already being applied.

glatterf42:
Not what I expected, and I'm not using Windows. Comparing the `latest py3.12` runs on Windows and Ubuntu, the only difference in the packages I notice is `colorama 0.4.6` being installed on Windows, but I don't think that's the issue. 
Comparing to the successful Windows `latest py3.10` run, the version of numpy differs. `Numpy 2.0.1` is only used on Python 3.11 or greater, it seems. 

My guess is that there's some kind of incompatibility with numpy still, but I'm struggling to troubleshoot this since `2.0.1` works fine on Ubuntu and Macos. The only differences in the workflows I notice are the `Install <OS> dependency` steps and those are very difficult to compare. Apart from that, the python version, package version, etc, are all identical. 
Looking at the traceback, the only package involved seems to be polars and they do actually have an open issue with the title sounding very similar to the error here: https://github.com/pola-rs/polars/issues/16863 
Unfortunately, I'm not sure if this is all that useful. The proposed solution is to not used `concat()` in favor of a little refactoring, but I'm not sure you can apply the same here. 
As far as I can tell from quickly looking at polar's source, I can't find a special treatment of Windows for the functions that seem to cause the error.

Happy to help further, but I'm slowly running out of ideas. Maybe it's still best to hold off on removing the numpy pin for now and hope the responsible party releases a fix soon.
Just a bit inconvenient for my special case: I only want to use Highs with linopy (for now), which works well with numpy 2.x, as does our side of packages, so my solution for now is to use my own branch which already allows numpy 2.x and just be aware that gurobipy can't be used with that (and that I'm anyway not on Windows, it seems).

ollie-bell:
To prevent windows from blocking this change, platform specifier could be used for numpy?

```
numpy>=2; sys_platform != 'win32'
numpy<2; sys_platform == 'win32'
```

pyproject.toml
```diff
@@ -24,7 +24,7 @@ classifiers = [
 
 requires-python = ">=3.9"
 dependencies = [
-    "numpy<2.0",
+    "numpy",
```
FabianHofmann, line 27:
```suggestion
    "numpy"; sys_platform != 'win32'  # or just numpy without version specifier
     "numpy"<2; sys_platform == 'win32'
```

pyproject.toml
```diff
@@ -24,7 +24,8 @@ classifiers = [
 
 requires-python = ">=3.9"
 dependencies = [
-    "numpy<2.0",
+    "numpy"; sys_platform != 'win32' 
+    "numpy"<2; sys_platform == 'win32'
```
FabianHofmann:
```suggestion
    "numpy"; sys_platform != 'win32',
    "numpy"<2; sys_platform == 'win32',
```

pyproject.toml
```diff
@@ -24,7 +24,8 @@ classifiers = [
 
 requires-python = ">=3.9"
 dependencies = [
-    "numpy<2.0",
+    "numpy"; sys_platform != 'win32'
+    "numpy"<2; sys_platform == 'win32'
```
FabianHofmann, line 28:
```suggestion
    "numpy"; sys_platform != 'win32',
    "numpy"<2; sys_platform == 'win32',
```

FabianHofmann:
@ollie-bell I think that is a good idea. applied it. @glatterf42 sorry for the mess, was trying to do it on GH with suggestions first...

ollie-bell:
@FabianHofmann is any help wanted with this? Happy to try and pick it up if so

FabianHofmann:
> @FabianHofmann @glatterf42 is any help wanted with this? Happy to try and pick it up if so

hey @ollie-bell, yes that would be wonderful!

coroa:
Closing, since it was already merged as part of #415

PR #333, opened 2024-07-24
ci: improve test workflow performance

afuetterer:
This PR does two things. Please have a look at it and tell me what you think.

1. it speeds up the package and dependency installation by using [uv](https://github.com/astral-sh/uv/) instead of pip
2. it runs the pytest suite in parallel

PR #334, opened 2024-07-29
ci: disable pytest multiprocessing

lkstrp:
Follow up on your request in #333 @FabianHofmann 

`print` can't handle multiprocessing, so it won't work. And logs can only be shown for all tests or none.
So it's either `pytest` in parallel or the log. I think we can revert multiprocessing, it doesn't take much time anyway (both around 30 sec).


FabianHofmann:
awesome @lkstrp !

PR #337, opened 2024-08-07
linexpression: drop squeeze argument in groupby

FabianHofmann:
remove the deprecated squeeze argument


FabianHofmann:
@lkstrp pypsa test run through with that version

PR #344, opened 2024-08-28
ci: move mypy to gh raction and add more rules

lkstrp:
See https://github.com/PyPSA/PyPSA/pull/1010
- Use github action instead of `pre-commit` for checks
- Add more rules

https://github.com/lkstrp/linopy/blob/325b3e2935cee6f384e5344257cc28d495f875db/pyproject.toml#L99-L110

To enable `disallow_untyped_defs` we need to make some more code changes. It currently throws 217 errors in the package. But I think most of them are `**kwargs` and missing `none` return type.


FabianHofmann:
:rocket:

PR #347, opened 2024-09-06
test: run tests in parallel

afuetterer:
This PR adds the excellent [pytest-xdist](https://github.com/pytest-dev/pytest-xdist) plugin to the dev dependencies.

The tests are run in parallel in CI.

This should speed up the CI run a bit.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -88,7 +88,7 @@ jobs:
       env:
         MOSEKLM_LICENSE_FILE: ${{ secrets.MSK_LICENSE }}
       run: |
-        pytest --cov=./ --cov-report=xml linopy --doctest-modules test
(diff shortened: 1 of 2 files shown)


lkstrp:
But pytest-xdist doesn't fix the missing logs issue raised in #333 and reverted in #334, does it?

afuetterer:
Oh no, I really just remembered that. Let me think about it.

Sorry.

afuetterer:
But looking at that PR, maybe there is also room for trying this action?
https://github.com/astral-sh/setup-uv

Ref: https://docs.astral.sh/uv/guides/integration/github/

lkstrp:
Hey @afuetterer
sorry for the late reply.
We could, but I also don't really see the reason. The speed up is marginal (?) I think and it mainly bloats up all workflow files. So not sure

afuetterer:
Sure, no problem.

Issue #198, opened 2023-11-09
Support iterative solve of the same solver_model (when using highs)


dannyopts:
I'm not sure what the right approach is here, so I'm going to write what I have done and found. 

I want to be able to solve the same model repeatedly with an altered cost function to understand model sensitivity. 

My specific use case is that I have a PyPSA model which takes about 50 seconds to solve using highs. 

I want to solve this model many time with different costs to understand If I update the cost function.

One way is in a loop, just keep resolving the problem, but I want to take advantage of the warm start power of highs.

 I have done this by accessing the solver_model and calling 
 ```
 m.solver_model.changeColsCost(len(coeffsi), np.arange(len(coeffsi), dtype=np.int32), coeffsi)
 ``` 

See below for 15 times taken for this kind of sequential solve, this gives approximately a 5x speed up vs the independent sequential solve. 

```
[58.32, 23.88, 2.74 8.54, 14.34, 17.98, 2.55, 22.21, 20.98, 24.55, 17.32, 13.22, 5.00, 4.12, 3.07]
```

But then I cant apply this solution back to the linopy model without extracting a bunch of code from model.solve related to applying the solution from a solver_model to the model. 

```
def apply_solution(model, h):
    CONDITION_MAP = {}
    condition = h.modelStatusToString(h.getModelStatus()).lower()
    termination_condition = CONDITION_MAP.get(condition, condition)
    status = Status.from_termination_condition(termination_condition)
    status.legacy_status = condition

    def get_solver_solution() -> Solution:
        objective = h.getObjectiveValue()
        solution = h.getSolution()

        sol = pd.Series(solution.col_value, h.getLp().col_names_, dtype=float).pipe(
            set_int_index
        )
        dual = pd.Series(solution.row_dual, h.getLp().row_names_, dtype=float).pipe(
            set_int_index
        )

        return Solution(sol, dual, objective)

    solution = safe_get_solution(status, get_solver_solution)

    result = Result(status, solution, h)
    
    model.objective._value = result.solution.objective
    model.status = result.status.status.value
    model.termination_condition = result.status.termination_condition.value
    model.solver_model = result.solver_model
    
    sol = result.solution.primal.copy()
    sol.loc[-1] = np.nan

    for name, var in model.variables.items():
        idx = np.ravel(var.labels)
        try:
            vals = sol[idx].values.reshape(var.labels.shape)
        except KeyError:
            vals = sol.reindex(idx).values.reshape(var.labels.shape)
[...]

PR #349, opened 2024-09-20
Refactoring of linopy solvers to object oriented architecture

daniel-rdt:
In this PR the `solvers.py` file is refactored to an object-oriented architecture. For this, a Solver superclass is introduced, as well as subclasses for each of the implemented solvers. A given problem can now be solved from a problem file without a linopy model using a specified solver. To solve a problem, for each Solver subclass the method `solve_problem_file()` is implemented. The implementation is based on the following logic suggested by @FabianHofmann in https://github.com/PyPSA/linopy/issues/198#issuecomment-1805530802:

```python
class Solver:
    def __init__(self, **kwargs):
        ....

    def solve_problem(self):
        raise NotImplementedError

class SolverA(Solver):
    ....
    def solve_problem(self, problem_fn):
        # Implementation of the solve method for SolverA
        ...
        return Result(status, solution)

class SolverB(Solver):
    ....
    def solve_problem(self, problem_fn):
        # Implementation of the solve method for SolverB
        ...
        return Result(status, solution)
....
```

To solve a problem directly from a problem file, for example using linopy's Gurobi solver interface:

```python
import linopy
from pathlib import Path

problem_file = Path("path/to/problem.lp")

gurobi = linopy.solvers.Gurobi()
gurobi.solve_problem(problem_fn=problem_file)
```

For direct execution using a linopy model, the method `set_direct_model` used that adds the model to the Solver instance:
```python
....
solver = solver_class()
solver.set_direct_model(model=model)
result = solver.solve_problem()
```

Two unit tests (`test_solver_classes_from_problem_file` and `test_solver_classes_direct`) were also added to test the functionality of the solver classes from a problem file and directly from a model. @finozzifa, any feedback on this is welcome.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,92 @@
+# For most projects, this workflow file will not need changing; you simply need
+# to commit it to your repository.
+#
+# You may wish to alter this file to override the set of languages analyzed,
+# or to provide custom queries or build logic.
+#
+# ******** NOTE ********
+# We have attempted to detect the languages in your repository. Please check
+# the `language` matrix defined below to confirm you have the correct set of
+# supported CodeQL languages.
+#
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ "master" ]
+  pull_request:
+    branches: [ "master" ]
+  schedule:
+    - cron: '19 21 * * 6'
+
+jobs:
+  analyze:
+    name: Analyze (${{ matrix.language }})
+    # Runner size impacts CodeQL analysis time. To learn more, please see:
+    #   - https://gh.io/recommended-hardware-resources-for-running-codeql
+    #   - https://gh.io/supported-runners-and-hardware-resources
+    #   - https://gh.io/using-larger-runners (GitHub.com only)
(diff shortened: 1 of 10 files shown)


FabianHofmann:
Great work @daniel-rdt and really efficient! I try to find time to review this asap. In case I am not able to @lkstrp has to jump in. 

Perhaps you try solving the type and test coverage warnings. I have the feeling that `sense` and `io_api` should not be mandatory arguments for initialization of the solver. The sense is given in the `lp` and `mps` files. So the current implementation would lead to ambiguity. Could you look into this?

daniel-rdt:
I've addressed the type check issues (except the type check for the pandas groupby which was addressed by PR #352). 

I've also adjusted the architecture of the direct solving from a linopy model vs solving from a problem file. The solver instance is not initialized with `io_api` and `sense` anymore. Rather the information is extracted from either the model or from the `lp` or `mps` problem file. I edited the code examples at the top of the PR to explain how it works.

As far as code coverage is concerned, I have not yet been able to get around to tackling that with adjusted unit tests unfortunately. Apart from this, @lkstrp, I think it's ready for your review.

lkstrp (approved):

PR #354, opened 2024-09-30
Improve cleanliness method print_infeasibilities()

kruksik-dev:
Currently, this method takes an argument that is of type None, and by default this parameter is also None. This is confusing when we want to use this method. In this MR, I correct the typing of this method so that it can accept int and None, but by default None is passed.

Additionally, the docstring for this method was difficult to read, so the generated documentation could also mislead the user. In the generated documentation there was no word about what this parameter is and that this method is only available when the gurobi solver is used (only visible in codebase via comment or when exception raised).

image

The revised docstring after generating the documentation should correct this and indicate that this method is only available when using the gurobi solver

### DIFF
--- a/linopy/model.py
+++ b/linopy/model.py
@@ -1162,12 +1162,24 @@ class Model:
                 labels.append(int(line_decoded.split(":")[0][2:]))
         return labels
 
-    def print_infeasibilities(self, display_max_terms: None = None) -> None:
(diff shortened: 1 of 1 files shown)


lkstrp (approved):
Great, thank you @kruksik-dev!

Docs and docstrings improvements are always very welcome!

PR #363, opened 2024-10-13
build: add support for python 3.13

afuetterer:


afuetterer:
The setup is really nice. The build step produces a test job for 3.13.

Now there is a problem with `gurobipy`. Maybe wait until a wheel is available.

lkstrp (approved):
It is great! Can be merged once this is resolved.

What do you think about adding [PEP702](https://peps.python.org/pep-0702/)? Could be a way to get rid of the `deprecation` dependency. I think that's the only practical change for python 3.13, but not sure yet.

afuetterer:
Closing this. Still not working, because of `gurobipy`.

PR #365, opened 2024-10-20
io: write out lp file with sliced variables and constraints

FabianHofmann:
When writing out an LP file, large variables and constraints are now chunked to avoid memory issues. This is especially useful for large models with constraints with many terms. The chunk size can be set with the `slice_size` argument in the `solve` function.

@koen-vg I think we are merging this quite soon, could you have a try with your large pypsa-eur myopic network? I hope this will finally fix the issues.

### DIFF
--- a/doc/release_notes.rst
+++ b/doc/release_notes.rst
@@ -4,6 +4,8 @@ Release Notes
 Upcoming Version
 ----------------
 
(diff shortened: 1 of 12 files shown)


koen-vg:
I would love to try this out, at a conference now so I can't promise to be very quick about it, but I'll report back as soon as I've tested this.

FabianHofmann:
@koen-vg thanks, sure, take your time. @lkstrp merging this as non-intrusive and well tested

PR #366, opened 2024-10-21
io: remove pandas based io for lp writing, use polars instead

FabianHofmann:


lkstrp (approved):
Since we are here, always have types in mind 😅

FabianHofmann:
Word word! Bad old habit... will add missing annotations

Issue #368, opened 2024-10-30
CPLEX: Optimal solution with unscaled infeasibilities is thrown out


thomgeo:
Hi everyone, I have an issue with CPLEX. for some problems, which Highs can solve, I get an unknown status from linopy, with the message that the problem is optimal with unscaled infeasibilities. As I understand, CPLEX introduced the solution of the scaled problem to the unscaled problem, and there were some slight infeasibilities, which usually do not mean a lot (see for examle here: http://iea-etsap.org/forum/showthread.php?tid=162)

Linopy, however, seems to not import this solution, potentially because the status is unknown and not optimal. Would it be possible to change this, so that linopy simply imports this solution, even though there might be slight infeasibilities?

Best,
Georg

thomgeo:
Hi Fabian, 

Thanks for the quick follow up. I updated linopy now, but it still does not import the solution if the status is "optimal with unscaled infeasibilities". Seems like it is still treated as infeasible.

Best,
GEorg

tburandt:
Hi @FabianHofmann , I get the same problem with linopy 4.2.0. Have you found time to look into this again?

Issue #362, opened 2024-10-11
CPLEX Unable to interpret unconstrained variables


darcyzhangapfsds:
Hi all. I'm solving a large optimization problem built with PyPSA. In the past I have used gurobi but am trying to set up cplex so that I can run unlimited number of instances in parallel. In every model I try to solve, CPLEX returns the same error: 

`CplexSolverError: CPLEX Error  1615: Line 2465: Expected number, found 'i'.`

To recreate, I use the linopy optimization example from: https://pypsa.readthedocs.io/en/latest/examples/optimization-with-linopy.html

```
import pypsa
import cplex

n = pypsa.examples.ac_dc_meshed(from_master=True)

n.optimize(solver_name='cplex')

``` 
This code returns the above error. Running this code without specifying a solver defaults to Highs and returns an optimal solution. I ran this in an anaconda environment using python 3.10 and the following package versions:

cplex: 22.1.1.2
linopy 0.3.15
pypsa 0.31.0

From the research I've done, it seems that creating the model using the PyPSA network passes 'inf' on variables with no constraints. CPLEX is interpreting 'inf' as a string rather than a number. 

Solutions tried: 

1. Running the above example using Highs
Result: works properly

2. Building the model and exporting it to csv. Then manually going through all CSV's and replacing 'inf' instances with 1000000. Then running:
```
n = pypsa.Network()
n.import_from_csv_folder("oct 10")
n.optimize(solver_name='cplex')
```
Result: CPLEX returns the same error still

3. Testing CPLEX in the same environment using linopy only in an example problem.
```
m = linopy.Model()
 
x = m.add_variables(lower=0, name="x")
y = m.add_variables(lower=0, name="y");
 
m.add_constraints(3 * x + 7 * y >= 10)
m.add_constraints(5 * x + 2 * y >= 3);
 
m.add_objective(x + 2 * y, sense="min")
 
m.solve(solver_name="cplex")
```
Result: CPLEX runs normally and returns an optimal solution 

4. Converting the PyPSA network to a linopy model and solving:
```
n.import_from_csv_folder("oct 10")
m = n.optimize.create_model()

linopy.solvers.run_cplex(m)
```
Result: same CPLEX error as others. 

Any and all help is appreciated. Thanks in advance.

darcyzhangapfsds:
Found a partial workaround for my use case doing the following:

```n.constraints.remove('Generator-e_sum_max')```

[...]

Issue #154, opened 2023-08-17
CPLEX Error  1615: Expected number, found 'i'.


euronion:
When using `cplex` with PyPSA-EUR and `linopy` I'm getting the following error.

It is linked to an `inf` constraint which I'm lazily imposing on the network, when filtering for the values which are `inf` and not imposing the constraint everything works fine.

`gurobi` does not have that issue and also runs fine with the `inf` constraint.

Probably related: https://github.com/Pyomo/pyomo/issues/364)

**Error message**
```
[snakemake / PyPSA-EUR stuff omitted]

INFO:linopy.model:Solver options:
 - threads: 4
 - lpmethod: 4
 - solutiontype: 2
 - barrier.convergetol: 0.0001
 - feasopt.tolerance: 1e-06
INFO:linopy.io:Writing objective.
Writing constraints.: 100%|████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 32/32 [00:20<00:00,  1.59it/s]
Writing continuous variables.: 100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 14/14 [00:03<00:00,  3.81it/s]
INFO:linopy.io: Writing time: 26.38s
Traceback (most recent call last):
  File "/p/tmp/<user>/pypsa-eur/.snakemake/shadow/tmpywrwfrn1/.snakemake/scripts/tmpp__dz4cb.solve_network.py", line 840, in <module>
    n = solve_network(
  File "/p/tmp/<user>/pypsa-eur/.snakemake/shadow/tmpywrwfrn1/.snakemake/scripts/tmpp__dz4cb.solve_network.py", line 781, in solve_network
    status, condition = n.optimize(**kwargs)
  File "/home/<user>/software/micromamba_1.4.2/envs/pypsa-eur/lib/python3.10/site-packages/pypsa/optimization/optimize.py", line 587, in __call__
    return optimize(self._parent, *args, **kwargs)
  File "/home/<user>/software/micromamba_1.4.2/envs/pypsa-eur/lib/python3.10/site-packages/pypsa/optimization/optimize.py", line 567, in optimize
    status, condition = m.solve(solver_name=solver_name, **solver_options, **kwargs)
  File "/home/<user>/software/micromamba_1.4.2/envs/pypsa-eur/lib/python3.10/site-packages/linopy/model.py", line 1042, in solve
    result = func(
[...]

PR #370, opened 2024-10-31
constraints: sanitize infinity values

FabianHofmann:
closes #369 #368 #154 #362

### DIFF
--- a/linopy/constraints.py
+++ b/linopy/constraints.py
@@ -51,7 +51,14 @@ from linopy.common import (
     to_polars,
 )
 from linopy.config import options
-from linopy.constants import EQUAL, HELPER_DIMS, TERM_DIM, SIGNS_pretty
+from linopy.constants import (
+    EQUAL,
+    GREATER_EQUAL,
+    HELPER_DIMS,
+    LESS_EQUAL,
+    TERM_DIM,
+    SIGNS_pretty,
+)
 from linopy.types import ConstantLike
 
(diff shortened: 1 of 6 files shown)


FabianHofmann:
@lkstrp with this implementation we do two things on top of #369
1. avoid large bounds in rhs by simply skipping unbounded constraints
2. check infeasible constraints of the form `>= inf` and `<= -inf`

linopy/constraints.py
```diff
@@ -872,6 +879,27 @@ def sanitize_missings(self):
...
+        """
+        for name in self:
+            constraint = self[name]
+            invalid_infinity_values = (
+                (constraint.sign == LESS_EQUAL) & (constraint.rhs == -np.inf)
+            ) | ((constraint.sign == GREATER_EQUAL) & (constraint.rhs == np.inf))
+            if invalid_infinity_values.any():
+                raise ValueError(
+                    f"Constraint {name} contains incorrect infinite values."
+                )
```
lkstrp, line 894: Hmm, this check is kind of not related to sanitizing and might make more sense during Constraint initialization? So it can not turned off? Might also be not nice to split that up

lkstrp (approved):
Yeah, I agree, this is much cleaner

linopy/constraints.py
FabianHofmann: you are right, moved it to the constraint assignment

PR #375, opened 2024-11-07
io: allow setting progress=False in solver to disable stdout even for larger models

maurerle:
This was unnecessary when solving a lot of problems programatically.

The `log` param of the solve function does have three states:

* `None`: log tqdm to stdout if model is > 10k constraints
* `True`: always creates tqdm for model
* `False`: never creates tqdm for model

For better usage, the log param is renamed to `progress` on all occurrences

### DIFF
--- a/linopy/io.py
+++ b/linopy/io.py
@@ -87,12 +87,12 @@ def objective_write_quad_terms(
 
 
 def objective_to_file(
-    m: Model, f: TextIOWrapper, log: bool = False, batch_size: int = 10000
+    m: Model, f: TextIOWrapper, progress: bool = False, batch_size: int = 10000
 ) -> None:
     """
     Write out the objective of a model to a lp file.
     """
-    if log:
+    if progress:
         logger.info("Writing objective.")
 
     sense = m.objective.sense
@@ -129,7 +129,7 @@ def objective_to_file(
 def constraints_to_file(
     m: Model,
     f: TextIOWrapper,
-    log: bool = False,
+    progress: bool = False,
     batch_size: int = 50_000,
     slice_size: int = 100_000,
 ) -> None:
@@ -138,7 +138,7 @@ def constraints_to_file(
 
     f.write("\n\ns.t.\n\n")
     names: Iterable = m.constraints
-    if log:
+    if progress:
         names = tqdm(
             list(names),
             desc="Writing constraints.",
@@ -198,7 +198,7 @@ def constraints_to_file(
 def bounds_to_file(
     m: Model,
     f: TextIOWrapper,
-    log: bool = False,
+    progress: bool = False,
     batch_size: int = 10000,
     slice_size: int = 100_000,
 ) -> None:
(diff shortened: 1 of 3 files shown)


linopy/io.py
```diff
@@ -311,9 +311,13 @@ def integers_to_file(
...
 
-def to_lp_file(m: Model, fn: Path, integer_label: str, slice_size: int = 10_000_000):
-    log = m._xCounter > 10_000
-
+def to_lp_file(
+    m: Model,
+    fn: Path,
+    integer_label: str,
+    slice_size: int = 10_000_000,
+    log: bool = None,
```
lkstrp, line 319:
```suggestion
    log: bool = True,
```
Can we make this `True` by default? It will be overwritten by the `to_file` call anyway, so `None` is not needed here?

linopy/io.py
```diff
@@ -559,9 +563,9 @@ def constraints_to_file_polars(m, f, log=False, lazy=False, slice_size=2_000_000
             # formatted.sink_csv(f,  **kwargs)
 
 
-def to_lp_file_polars(m, fn, integer_label="general", slice_size=2_000_000):
-    log = m._xCounter > 10_000
-
+def to_lp_file_polars(
+    m, fn, integer_label="general", slice_size=2_000_000, log: bool = None
```
lkstrp, line 567: Same as above.

linopy/io.py
```diff
@@ -583,6 +587,7 @@ def to_file(
     io_api: str | None = None,
     integer_label: str = "general",
     slice_size: int = 2_000_000,
+    log=None,
```
lkstrp, line 590:
```suggestion
    log: bool | None = None,
```
Fix type checking

linopy/model.py
```diff
@@ -969,6 +969,7 @@ def solve(
         sanitize_infinities: bool = True,
         slice_size: int = 2_000_000,
         remote: None = None,
+        stdout: bool | None = None,
```
lkstrp, line 972:
```suggestion
        log: bool | None = None,
```
I guess log is not really the right name here, but better to keep it consistent.
Please also add it to the docstring and briefly explain its logic.

lkstrp (changes requested):
Looks good, thank you @maurerle 
A few minor things left on types and docstrings.

linopy/model.py
maurerle: it is especially confusion as solve already has a `log_fn` to provide a log filename as well. But yes, consistency is important.

What do you think about renaming it to `progress` - as it basically toggles to show a progress bar?

maurerle:
Looks good from my side now, @lkstrp can you take another look please? :)

lkstrp (approved):
Looks good, thank you!
Two last things:
- Release note is missing
- We can deprecate `log` instead of removing it. So accept both `log` and `progress`, and throw a `DeprecationWarning` when `log` is passed. There is a decorator for this somewhere, I can add it if you don't mind. It may be unnecessary, but I think it's better practice.

maurerle:
If you add the remarks yourself, that would be great!
You should be able to push to my branch

From my point of view, the deprecation is not really necessary as there did not exist a public api to set log before anyway.

lkstrp:
Agreed, thank you @maurerle !

PR #379, opened 2024-11-15
fix: correctly pass `slice_size` argument

FabianHofmann:
... the argument was not correctly passed before


FabianHofmann:
@koen-vg I can solve your model now with this bugfix :)

PR #381, opened 2024-11-21
io: open file using context manager to properly close it again

tburandt:
instead of a blind open(), solvers.py now uses a context manager to close the file handle properly after reading the file contents


FabianHofmann (approved):
thanks @tburandt for the quick fix!

PR #383, opened 2024-11-22
COPT update

PeterKlein11:
Removing this line of code stops COPT calling a floating license from a server in cases where COPT is not used to solve the problem. 

Solver still solves as normal on my side after its removal.

### DIFF
--- a/linopy/solvers.py
+++ b/linopy/solvers.py
@@ -108,8 +108,7 @@ with contextlib.suppress(ModuleNotFoundError):
 with contextlib.suppress(ModuleNotFoundError):
     import coptpy
 
-    with contextlib.suppress(coptpy.CoptError):
-        available_solvers.append("copt")
+    available_solvers.append("copt")
 
 quadratic_solvers = [s for s in QUADRATIC_SOLVERS if s in available_solvers]
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


linopy/solvers.py
```diff
@@ -109,8 +109,6 @@
     import coptpy
 
     with contextlib.suppress(coptpy.CoptError):
```
lkstrp, line 111:
```suggestion
    available_solvers.append("copt")
```

Please remove the check then as well

lkstrp (commented):
Why can you not just uninstall `coptpy` if you are not using it?

But I guess you have a point if `coptpy.Envr()` does not fail when no licence is set up.

PeterKlein11:
The issue is that we do use COPT. We have a floating license, but only a limited number of seats. Each time pypsa is imported it checks out the license even if you not doing solving a problem. Like building a snakemake workflow.

lkstrp:
Understood. You are right, we can just remove the check and the import check will be enough. If you resolve the comment, we can merge this. I don't have access to your fork and can't do it myself.

coroa:
Merging after tests complete.

PeterKlein11:
Thanks. Sorry I fixed locally and I totally forgot about the request.

coroa:
> Thanks. Sorry I fixed locally and I totally forgot about the request.

Sure. I was (again) totally amazed that i receive push rights to your master branch, when you make a PR to here.

Issue #157, opened 2023-09-05
Request: allow retention of original variable names when writing to lp file


JorritHimself:
In pyomo one could specifiy: 
`model.write(lp_file_name, io_options={'symbolic_solver_labels': True})`
Which retains original variable names, rathe rthan having these rewritten to x0, x1, x2 etc.

We use a lot of scripts to parse both problem and solution files for all sorts of analyses/plots, and that'd require having the original variable names used when writing to lp file. 
From the dicord channel I udnerstood the option has not (yet) been included in 
`Model.to_file(fn)`
but it might be possible to include. 
Would help a lot with my projects.
Cheers.

OBeaude:
Having recently started with Linopy, I wanted to check the obtained lp file - also using model.to_file method -  against some reference one obtained with pyomo... and faced the same issue.

Anything new since your post in Sept. 2023 @FabianHofmann ?

Thank you

FabianHofmann:
Hey @OBeaude, unfortunately not. Maintenance swallows a lot of time, and this did not seem pressing to me. What would be your specific use case?

OBeaude:
Hey @FabianHofmann. In turn, we could code a patch and make a pull request. Would it be convenient for you? Any advice?

N.B. Our first use case is to check that we reproduce the writing of the .lp file we got using Pyomo on a given LP

FabianHofmann:
Sure! PR's are always welcome!

FabianHofmann:
resolved in #391

PR #391, opened 2024-12-06
Export variable and constraint names in LP files

olivierjuan:
A small modification to export LP files with variable and constraint names.
Gurobi is able to read the generated LP files.

This modification does not apply to polar export.
I'm ready to improve the code if needed

This is related to issue https://github.com/PyPSA/linopy/issues/157

### DIFF
--- a/doc/release_notes.rst
+++ b/doc/release_notes.rst
@@ -6,6 +6,7 @@ Upcoming Version
 
 * The internal handling of `Solution` objects was improved for more consistency. Solution objects created from solver calls now preserve the exact index names from the input file.
 * Multiplication of a linear expression by a constant value may now introduce new dimensions.
+* Added method `unstack` to `LinearExpression`, `Variable` and `Constraint` to unstack a dimension.
 
 Version 0.4.4
 --------------
--- a/linopy/constraints.py
+++ b/linopy/constraints.py
@@ -675,6 +675,8 @@ class Constraint:
 
     stack = conwrap(Dataset.stack)
 
+    unstack = conwrap(Dataset.unstack)
+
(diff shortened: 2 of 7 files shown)


olivierjuan:
I'll fix the tests first

olivierjuan:
I added a new io_api type called "lp-debug" to write lp file with variable and constraint names. I don't know if you will like that approach. I can replace it by an extra argument to the function to_file instead if you prefer.

FabianHofmann (commented):
pretty cool @olivierjuan! Thank you for your contribution! I would suggest to have an extra argument for this and raise an error if a non-supported io-api is passed. Could you check whether this has any impact on the regular lp file writing?

olivierjuan:
Ok. I'm on it

olivierjuan:
Well I did it in the end.
So the variable and constraint names  can be transfered to the solver either for lp, lp-polars, mps or direct api.

Only CBC and GLPK are incompatible (if trying to solve with_names=True and with one of these solvers, a warning will be printed and the option will be deactivated).
I couldn't test it with mindopt due to issues with my mindopt license. But all other solvers work fine

FabianHofmann:
Wow, well done @olivierjuan ! I promise to have a look soon and then we can merge it 🚀

lkstrp (approved):
Hey @olivierjuan, thanks for the contribution! This was a requested feature
We wanna get this merged now and have only some minor changes, e.g. use `explicit_coordinate_names` as argument name instead. 
Would you mind if I just push those changes?

olivierjuan:
As you want, I fine with you modifying it.

PR #399, opened 2024-12-30
Complete type annotations

FabianHofmann:


.gitignore
```diff
@@ -26,3 +26,4 @@ benchmark/notebooks/.ipynb_checkpoints
 benchmark/scripts/__pycache__
 benchmark/scripts/benchmarks-pypsa-eur/__pycache__
 benchmark/scripts/leftovers/
+.aider*
```
lkstrp, line 29: Just FYI: [Setting up a global .gitignore file | Sebastian De Deyne](https://sebastiandedeyne.com/setting-up-a-global-gitignore-file/)

linopy/common.py
```diff
@@ -474,11 +478,46 @@ def fill_missing_coords(ds, fill_helper_dims: bool = False):
...
+    slice_dims: list | None = None,
+) -> Generator[Variable, None, None]: ...
+
+
+@overload
+def iterate_slices(
+    ds: LinearExpression,
+    slice_size: int | None = 10_000,
+    slice_dims: list | None = None,
+) -> Generator[LinearExpression, None, None]: ...
```
lkstrp, line 505: 🚀 
Were you just playing around, or do you think we could use overloading a lot more everywhere?
I wasn't sure. Probably great for type checking and IDE hints, but also adds a bit of boilerplate. We'll probably want to do stub files at some point.

linopy/constants.py
```diff
@@ -11,7 +11,7 @@
 import numpy as np
 import pandas as pd
 
-logger = logging.getLogger(__name__)
+logger: logging.Logger = logging.getLogger(__name__)
```
lkstrp, line 14: What was the mypy issue here? Type `Any` because logging is not typed? But you did not set that rule

linopy/constraints.py
```diff
@@ -685,7 +688,7 @@ class Constraints:
     """
 
     data: dict[str, Constraint]
-    model: "Model"  # Model is not defined due to circular imports
+    model: Model  # Model is not defined due to circular imports
```
lkstrp, line 691:
```suggestion
    model: Model
```

linopy/types.py
```diff
@@ -22,15 +26,32 @@
...
-ConstraintLike = Union["Constraint", "AnonymousScalarConstraint"]
-MaskLike = Union[numpy.ndarray, DataArray, Series, DataFrame]
-SideLike = Union[ConstantLike, VariableLike, ExpressionLike]
 
+Coords: TypeAlias = (
+    Sequence[Sequence | Index | DataArray]
+    | Mapping
+    | DataArrayCoordinates
+    | DatasetCoordinates
+    | None
```
lkstrp, line 35: Isn't `CoordsLike` still a more consistent name, even if there is no `Coords` class? Otherwise, when reading the types, my first assumption is that this is a class and not a type alias.

Also, should `None` also go in there?

lkstrp (commented):
Tons of types 💃 

Why did you remove `py.typed`? It is currently broken, but the whole goal of this operaiton should be to make linopy a typed package. Otherwise it is only useful for internal development.

linopy/common.py
FabianHofmann: both, played around and think we have to use it a lot more (though I don't like the style too much, it boats quite significantly)

linopy/types.py
FabianHofmann: you are right! I update the name

FabianHofmann: you're right, I'm pulling the None out of it

linopy/constants.py
FabianHofmann: just, very ambitious AI code ;)

FabianHofmann: revert

Issue #396, opened 2024-12-18
Invalid literal for int: error while solving MPS files with longer column names


siddharth-krishna:
I'm facing an error when running linopy on MPS files where column names have more than one letter in them (e.g. `col01` as opposed to `x01`). Here is an example MPS file:
```
NAME        sample_mip
ROWS
 N  obj     
 G  c1      
 L  c2      
 E  c3      
COLUMNS
    col1        obj       5
    col1        c1        2
    col1        c2        4
    col1        c3        1
    MARK0000  'MARKER'                 'INTORG'
    col2        obj       3
    col2        c1        3
    col2        c2        2
    col2        c3        1
    col3        obj       7
    col3        c1        4
    col3        c2        3
    col3        c3        1
    MARK0001  'MARKER'                 'INTEND'
RHS
    RHS_V     c1        12
    RHS_V     c2        15
    RHS_V     c3        6
BOUNDS
 UP BOUND     col1        4
 UI BOUND     col2        3
 UI BOUND     col3        5
ENDATA
```

I run this using the python script
```python
import sys
from pathlib import Path
from linopy import solvers

s = solvers.Highs()
result = s.solve_problem(problem_fn=Path(sys.argv[1]), solution_fn=Path("/tmp/sid-linopy-out.sol"))
print(result)
print(result.solution.objective)
```

Which results in the following error (a similar error appears when using GLPK/Gurobi solvers):
```
Traceback (most recent call last):
  File ".../linopy-test.py", line 6, in <module>                                           
    result = s.solve_problem(problem_fn=Path(sys.argv[1]), solution_fn=Path("/tmp/sid-linopy-out.sol"))                       
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                       
  File ".../lib/python3.12/site-packages/linopy/solvers.py", line 299, in solve_problem
    return self.solve_problem_from_file(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                      
  File ".../lib/python3.12/site-packages/linopy/solvers.py", line 775, in solve_problem_from_file
    return self._solve(                                                                                                       
           ^^^^^^^^^^^^
  File ".../lib/python3.12/site-packages/linopy/solvers.py", line 867, in _solve    
[...]

PR #406, opened 2025-01-29
feat: move set_int_index to solution assignment in model.py

FabianHofmann:
Follow up on #396 

The internal handling of `Solution` objects was improved for more consistency. Solution objects created from solver calls now preserve the exact index names from the input file.

## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,14 @@ Highs.log
 paper/
 monkeytype.sqlite3
 
+# Environments
+.env
+.venv
+env/
+venv/
+ENV/
+env.bak/
+venv.bak/
 
 benchmark/*.pdf
 benchmark/benchmarks
--- a/doc/release_notes.rst
+++ b/doc/release_notes.rst
@@ -1,8 +1,10 @@
 Release Notes
 =============
 
-.. Upcoming Version
-.. ----------------
+Upcoming Version
+----------------
+
+* The internal handling of `Solution` objects was improved for more consistency. Solution objects created from solver calls now preserve the exact index names from the input file, ensuring consistent behavior across all solver functions.
 
 Version 0.4.4
 --------------
--- a/doc/release_notes.rst
+++ b/doc/release_notes.rst
@@ -4,7 +4,7 @@ Release Notes
 Upcoming Version
 ----------------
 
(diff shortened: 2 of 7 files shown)


siddharth-krishna (approved):
Thanks Fabian, this works for me! If I understand the idea correctly, I think it's also a more elegant approach, and ensures we only call `set_int_index` on solutions returned from `linopy.Model`, which is guaranteed to have the form `x1, x2, ...`.

Btw, here is a test file I have been using to test this issue. If it makes sense, I can add it to the tests in this PR, if you can point me to where/how to integrate them. Thanks!
```python
#!/usr/bin/env python3
"""
Created on Tue Jan 28 09:03:35 2025.

@author: sid
"""

import sys
from pathlib import Path
from linopy import solvers

mps_problem = """
NAME        sample_mip
ROWS
 N  obj     
 G  c1      
 L  c2      
 E  c3      
COLUMNS
    col1        obj       5
    col1        c1        2
    col1        c2        4
    col1        c3        1
    MARK0000  'MARKER'                 'INTORG'
    colu2        obj       3
    colu2        c1        3
    colu2        c2        2
    colu2        c3        1
    col3        obj       7
    col3        c1        4
    col3        c2        3
    col3        c3        1
    MARK0001  'MARKER'                 'INTEND'
RHS
    RHS_V     c1        12
    RHS_V     c2        15
    RHS_V     c3        6
BOUNDS
 UP BOUND     col1        4
 UI BOUND     colu2        3
 UI BOUND     col3        5
ENDATA
"""
input_file = "/tmp/sid-linopy.mps"
with open(input_file, "w") as f:
    f.write(mps_problem)

s = solvers.Highs()
result = s.solve_problem(problem_fn=Path(input_file), solution_fn=Path("/tmp/sid-linopy-out.sol"))
print(result)
print(result.solution.objective)
```

FabianHofmann:
great, since this a new class of test (which only tackles the solver implementations) I would suggest we create a new module `test/test_solvers.py` where we test these low-level solver routines. what do you think?

FabianHofmann:
you can push it directly to this PR (permissions should be in place)

siddharth-krishna:
Thanks, Fabian! I added a new test file. I also added some common virtualenv paths to the .gitignore, if you don't mind?

Looks like the test is failing on some solvers: glpk, cbc

Do you think it's worth investigating, or should I reduce the scope of the test to HiGHS?

PR #415, opened 2025-02-08
fix: unpin numpy<2 for python>3.10

ollie-bell:
Remove numpy<2 pin for python>3.10

Taking over from https://github.com/PyPSA/linopy/pull/332

### DIFF
--- a/linopy/common.py
+++ b/linopy/common.py
@@ -8,7 +8,6 @@ This module contains commonly used functions.
 from __future__ import annotations
 
 import operator
-import os
 from collections.abc import Generator, Hashable, Iterable, Sequence
 from functools import reduce, wraps
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


ollie-bell:
@FabianHofmann simple solution is to just keep numpy<2 for python 3.9 and 3.10 as these encounter a segfault on ubuntu when running pytest otherwise... which may be quite time consuming to try and debug. I'm not sure it's worth it for these soon to be unsupported python versions anyway?

Also not sure why the RTD build failed, maybe it just needs re-triggering.

ollie-bell:
Think this is good to go!

FabianHofmann (approved):

FabianHofmann:
Great Job @ollie-bell

PR #425, opened 2025-03-12
fix: change empty to property

coroa:
In view of #424 (ie. empty being mostly True automatically), i would argue the current use of `empty` is quite limited and i'd prefer to take this chance to make it a `property` as it is in pandas, since it is today too easy to misuse as `expr.empty` which is always truthy.

~Contrarily the upgrade path from `expr.empty()` to `expr.empty` will always raise a clear Exception.~

Update: Added the wrapper that @FabianHofmann sketched below to smooth the transition.

## Checklist

### DIFF
--- a/doc/release_notes.rst
+++ b/doc/release_notes.rst
@@ -3,6 +3,13 @@ Release Notes
 
 Upcoming Version
 ----------------
+**Deprecations**
+
+* Renamed `expression.empty()` to `expression.empty` to align with the use of empty in
+  pandas. A custom wrapper ensures that `expression.empty()` continues to work, but emits
+  a DeprecationWarning.
+
+**Features**
 
 * Added support for arithmetic operations with custom classes.
 * Avoid allocating a floating license for COPT during the initial solver check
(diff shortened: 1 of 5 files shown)


FabianHofmann:
I like that and wanted to do this for a long time, was scared of the breaking change though. could we instead define a callable `Empty` class returned by the property, which if called raises a deprecation warning? Something like 


```
import warnings
import functools

class Empty:
    def __init__(self, value):
        self._value = bool(value)
    
    def __bool__(self):
        return self._value
    
    def __call__(self):
        warnings.warn(
            "Calling `.empty()` is deprecated, use `.empty` property instead",
            DeprecationWarning,
            stacklevel=2
        )
        return self._value
    
```

and then in the property

```
    @property
    def empty(self):
        is_empty = ...
        return Empty(is_empty)
 ```

coroa:
I guess that would work, quite a bit of boilerplate for that simple change, but everyone loves downward compatibility!

lkstrp (commented):
We might also want to add `__repr__` to the descriptor. Otherwise the memory reference is shown, which is confusing.

Even cleaner, which would not mock a simple boolean, the process would probably be to have a module wide option setting, raise a deprecation warning for `.empty()` without breaking behaviour, and allow opting in to use `.empty`.

coroa:
Added the `__repr__`, but went with the wrapper. I am not a big fan of opt-in. I directly want to use the good shit.

PR #441, opened 2025-04-15
fix: don't disable presolve and fix dual value retrieval for SCIP

siddharth-krishna:
Closes # (if applicable).

Feedback from SCIP developers in https://github.com/open-energy-transition/solver-benchmark/issues/148 indicates that linopy might be making an error when retrieving dual solution values from SCIP. Also, they recommend against disabling presolve, which might have been done in order to retrieve dual solution values. This PR makes both changes.

I think we might not need a separate function for LPs and MILPs, since the dual retrieval code is in an if-then block?

We should also test that removing the code that filters dual variable names to only those starting with `c` or `cf` is no longer necessary.

## Checklist

### DIFF
--- a/linopy/solvers.py
+++ b/linopy/solvers.py
@@ -472,11 +472,6 @@ class CBC(Solver):
             sol = df[variables_b][2]
             dual = df[~variables_b][3]
 
-            # import ipdb
-            # ipdb.set_trace()
-            # TODO the problem is that this sol file has both rows and columns in it, and uses original names
-            # linopy requires highs for MPS files anyway, so we can use highs for MPS to tell which ones are vars
-
             return Solution(sol, dual, objective)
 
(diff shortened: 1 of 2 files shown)


linopy/solvers.py
```diff
@@ -471,6 +471,12 @@ def get_solver_solution() -> Solution:
 
             sol = df[variables_b][2]
             dual = df[~variables_b][3]
+
+            # import ipdb
+            # ipdb.set_trace()
+            # TODO the problem is that this sol file has both rows and columns in it, and uses original names
+            # linopy requires highs for MPS files anyway, so we can use highs for MPS to tell which ones are vars
+
```
FabianHofmann, line 479: remove this comment?

siddharth-krishna: Yes sorry, got added by mistake, will push a fix shortly!

FabianHofmann (approved):

PR #447, opened 2025-04-25
feat: termination condition map for SCIP solver

siddharth-krishna:
Closes # (if applicable).

Currently, if the user sets the `limits/gap` SCIP solver option to stop solving MIPs when a certain duality gap is reached, then linopy reports the termination condition as unknown. This PR tries to map the SCIP statuses to linopy termination conditions, using
https://github.com/scipopt/scip/blob/b2bac412222296ff2b7f2347bb77d5fc4e05a2a1/src/scip/type_stat.h#L40

(CC @pfetsch, I couldn't find the exact mapping from the enum integers to the strings returned by `pyscipopt` so I guessed that it's the same as the names of the enums)

## Checklist

### DIFF
--- a/doc/release_notes.rst
+++ b/doc/release_notes.rst
@@ -1,9 +1,12 @@
 Release Notes
 =============
 
-.. Upcoming Version
-.. ----------------
+Upcoming Version
+----------------
 
+**Bug Fixes**
+
+* Improve the mapping of termination conditions for the SCIP solver
 
 Version 0.5.3
 --------------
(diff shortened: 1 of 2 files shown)


FabianHofmann (approved):

pfetsch:
The mapping looks good, except possibly for `gap limit reached`. This should probably be handled differently. Currently, you cannot determine whether this happened or some other limit was reached. I think that for the other solvers, a gap limit reached is treated as optimal.

siddharth-krishna:
Thank you @pfetsch ! That makes sense, I've opened a new PR to fix that: https://github.com/PyPSA/linopy/pull/448

PR #475, opened 2025-07-18
MOSEK: Remove explicit use of mosek.Env

### LINKED ISSUE #474: Using MOSEK from Linopy
When using MOSEK from linopy a new `mosek.Env` is created for each new `mosek.Task`.

It would be better to create `mosek.Task` with no environment, which translates to using the hidden global environment. Since some things, like licenses, are checked out, cached and checked in for each environment, it may create unnecessary  overhead and potentially check or more licenses than necessary.

ulfworsoe:
Closes #474 

- Remove explicit use of `mosek.Env`, use global environment instead. This avoids checking out new licenses for each new task created, and avoids creating and destroying environment object for each task.

## Checklist

### DIFF
--- a/doc/release_notes.rst
+++ b/doc/release_notes.rst
@@ -8,6 +8,7 @@ Release Notes
 
 * Improved variable/expression arithmetic methods so that they correctly handle types
 * Gurobi: Pass dictionary as env argument `env={...}` through to gurobi env creation
+* Mosek: Remove explicit use of Env, use global env instead
 
 **Breaking Changes**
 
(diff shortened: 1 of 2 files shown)


lkstrp (commented):
Thanks @ulfworsoe !

Could you update the docstring and add a warning that when `env` is passed, a `DeprecationWarning` is thrown to indicate that global env should be used instead? Perhaps also link to their documentation. Right now `env` is still in the signature but not doing something without any warning

Also pre-commit and types need to be fixed

maurerle:
@ulfworsoe you should also run the pre-commits:
pip install pre-commit
and then `pre-commit run --all-files`

ulfworsoe:
Fixed. I also ran the pre-commits (thank you for instructions).

PR #485, opened 2025-09-03
fix: lp-polars api

CharlieFModo:
Closes # (if applicable).
#484 

## Changes proposed in this Pull Request
Makes `lp-polars` API behave the same as `lp`. All screenshots below use the same problem.

linopy 0.5.6 - io_api="lp-polars"
image

this branch - io_api="lp-polars"
image

linopy 0.5.6 - io_api="lp"
image 

## Checklist

- [X] Code changes are sufficiently documented; i.e. new functions contain docstrings and further explanations may be given in `doc`.
- [X] Unit tests for new features were added (if applicable).
- [X] A note for the release notes `doc/release_notes.rst` of the upcoming release is included.
- [X] I consent to the release of this PR's code under the MIT license.

PR #495, opened 2025-09-07
Add sos constraints

### LINKED ISSUE #437: Special Ordered Sets (SOS)
# Defining Special Ordered Sets SOS1 & SOS2

I would really like to see the option to define special ordered sets through linopy (of course only for compatible solvers) 
Below is a quick overwview on what SOS are and why they would be a nice addition.
If I get some approval I might try to implement it myself. Has someone tried it before?

## Special Ordered Sets (SOS)
Special Ordered Sets (SOS) are specialized mathematical structures used in mixed-integer programming to model certain types of constraints more efficiently. There are two main types:

### Special Ordered Set Type 1 (SOS1)
A set of variables where at most one variable can be non-zero. This is useful for mutually exclusive choices.

### Special Ordered Set Type 2 (SOS2)
A set of ordered variables where at most two consecutive variables can be non-zero, and those two must be adjacent in the ordering. This is frequently used to model piecewise linear functions.

## Key Benefits

1. **Computational Efficiency**: Branch-and-bound algorithms can exploit the structure of SOS constraints to make more intelligent branching decisions, often leading to faster solution times compared to using standard binary variables.

2. **Modeling Elegance**: SOS constraints provide a more natural way to express certain types of problems, particularly piecewise linear approximations of nonlinear functions.

3. **Solution Quality**: For some problem types, formulations using SOS can lead to tighter LP relaxations, resulting in better bounds during the solution process.

4. **Reduced Model Size**: In some cases, using SOS constraints can lead to models with fewer variables or constraints than equivalent formulations using binary variables.

5. **Specialized Solver Support**: Most commercial solvers (like CPLEX, Gurobi) have specialized algorithms to handle SOS constraints efficiently.

SOS constraints are particularly valuable in applications like:
- Supply chain optimization with fixed costs
- Facility location problems
- Energy production scheduling
- Financial portfolio optimization
- Approximating nonlinear functions in various optimization contexts

coroa:
Closes #437  (if applicable).

### Add Special Ordered Sets (SOS) Constraints Support

I added SOS1/2 constraints as an extension of variables. ie. in the example

```python
import linopy

m = linopy.Model()
# Create variables for facility locations
locations = pd.Index([0, 1, 2], name="locations")
build = m.add_variables(coords=[locations], name="build", binary=True)

# Add SOS1: at most one facility can be built
m.add_sos_constraints(build, sos_type=1, sos_dim="locations")
```

The `add_sos_constraints` only adds `sos_type` and `sos_dim` as attributes to the `build` dataset.

These attributes have then to be interpreted by the lp file writer or the direct apis. In this draft
version this is implemented for the gurobipy direct api as well as for the lp-polars writer.

The coordinates of the chosen dimension are also always used as weights for the SOS. It felt natural,
but i am open to other suggestions. Weights are only used by SOS2 constraints to order them, and then
only two adjacent variables can be non-zero at the same time. In principle always taking the coordinate
order directly (ie use something like `np.arange(var.sizes[sos_dim])` as weights) would be the another
option.

The documentation that i added was claude written and might still contain mistakes. The code is tested,
but the claude tests are to extensive to be helpful. I need to clean them up first.

### Core SOS Constraints API
- **`Model.add_sos_constraints()`**: New method to add SOS1 and SOS2 constraints to variables
- **Multi-dimensional support**: SOS constraints work with multi-dimensional variables, creating constraints for each combination of non-SOS dimensions
- **Attribute-based tracking**: SOS constraints are stored as variable attributes (`sos_type`, `sos_dim`)
- **Variables.sos property**: Easy access to all SOS-constrained variables in a model

### SOS Constraint Types
- **SOS1 (Type 1)**: At most one variable in the ordered set can be non-zero
- **SOS2 (Type 2)**: At most two adjacent variables (based on ordering weights) can be non-zero

### Solver Support
- **File-based solvers**: LP/MPS file format output with SOS constraint sections
- **Gurobi direct API**: Native SOS constraint support via `gurobipy`

#### Core Implementation
- `linopy/model.py`: Added `add_sos_constraints()` method with validation
- `linopy/variables.py`: Enhanced Variable class with SOS attribute support and `Variables.sos` property
- `linopy/io.py`: SOS constraint export for file-based solvers and Gurobi direct API

#### Documentation & Examples
- `doc/sos-constraints.rst`: Comprehensive documentation with theory and examples
- `doc/index.rst`: Updated to include SOS constraints documentation

### Testing
to be done

## Checklist


coroa:
Note that with the implementation here, it is currently unclear how to remove an SOS constraint again or where to access dual values. Is that a thing for SOS constraints?

FabianHofmann (commented):
Impressive and wonderful! the IO currently only support the polars lp file writing, not the "standard" pandas based, right? for non-breaking compat we would then need to focus on the IO

* export to highspy & gurobipy  (and/or raise an error if we don't support for now)
* export with standard lp file (generally also happy to skip, in case we merge in #366 after merging #485 and making a final benchmark)

coroa:
1. **weights**: Thanks for this quote of the gurobi docs, which confirms that the weights do not have an inherent meaning any more. For avoiding surprises and for ease of use, i would suggest to use the dimension order directly, since we already tied it now to a dimension (with an order), ie. i'll remove the weights code again.

2. **raises/checks**: i'll add the remaining io_api and solver implementations, but not for the standard lp writer due to #496 . Can anyone try to find out which solvers do and don't support it?

3. **representation like m.constraints**. hmm. yes, that is what i meant what is tricky. in this implementation sos constraints are a variable quality and you get all variables with an sos constraint with: `m.variables.sos`, which currently looks like:

```
linopy.model.Variables
----------------------
 * build (locations)
```

i guess we could improve this representation to:
```
linopy.model.Variables
----------------------------
 * build (locations) - sos1 on locations
```

fneum:
> representation like m.constraints. hmm. yes, that is what i meant what is tricky. in this implementation sos constraints are a variable quality and you get all variables with an sos constraint with: m.variables.sos, which currently looks like:

I guess this is fine as long as it is possible to distinguish between SOS1 and SOS2 sets.

fneum:
Awesome! Looks great!

PR #497, opened 2025-09-08
perf: use xpress c interface, xpress solution order of operations improvement, remove writing of redundant solution files

CharlieFModo:
Closes # (if applicable).

- performance improvements to `Xpress._solve_problem_from_file`
- use the solver's C interface directly to extract constraint and variable names rather than list comprehension
  - casting to string the xpress objects returned by `problem.getConstraints()` etc. is expensive and so is accessing the names via `.name`.
  - this gives small performance improvement on larger models
- improve order of operations when extracting dual
  - for MILPs, let code hit the exception on `getDuals` before doing the expensive constraint name extraction
- use `writebinsol` for a c. 10x improvement from `writesol`
  - @FabianHofmann it would be nice to not write the solution at all for xpress if `keep_files=False`

## Checklist

### DIFF
--- a/linopy/model.py
+++ b/linopy/model.py
@@ -59,7 +59,12 @@ from linopy.io import (
 from linopy.matrices import MatrixAccessor
 from linopy.objective import Objective
 from linopy.remote import OetcHandler, RemoteHandler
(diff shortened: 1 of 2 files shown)


FabianHofmann (approved):
that was fast ;)

PR #516, opened 2025-10-29
Add support for solver cuPDLPx

mal84emma:

Add support for GPU-accelerated solver [cuPDLPx](https://github.com/MIT-Lu-Lab/cuPDLPx)

Note: due to the lower numerical precision of this solver, some of the equality assertions in the tests have been replaced with `np.isclose`
Also note: this solver does not support QP or MIP, and currently the file IO is not available through the Python API (will add when later versions support it)

## Checklist

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -36,7 +36,7 @@ jobs:
     permissions:
       id-token: write
     steps:
-      - uses: actions/download-artifact@v6
+      - uses: actions/download-artifact@v7
         with:
           name: Packages
           path: dist
--- a/.github/workflows/test-models.yml
+++ b/.github/workflows/test-models.yml
@@ -59,7 +59,7 @@ jobs:
           echo -ne "url: ${CDSAPI_URL}\nkey: ${CDSAPI_TOKEN}\n" > ~/.cdsapirc
           echo "week=$(date +'%Y%U')" >> $GITHUB_ENV # data and cutouts
 
-      - uses: actions/cache@v4
+      - uses: actions/cache@v5
         if: env.pinned == 'false'
         with:
           path: |
@@ -74,7 +74,7 @@ jobs:
 
       - name: Cache Conda env
         if: env.pinned == 'false'
-        uses: actions/cache@v4
+        uses: actions/cache@v5
         with:
           path: ${{ env.CONDA }}/envs
           key: conda-pypsa-eur-${{ env.week }}-${{ hashFiles('envs/linux-64.lock.yaml') }}
@@ -101,7 +101,7 @@ jobs:
 
       - name: Upload artifacts
         if: env.pinned == 'false'
-        uses: actions/upload-artifact@v5
+        uses: actions/upload-artifact@v6
         with:
           name: results-pypsa-eur-${{ matrix.version }}
           path: |
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -74,7 +74,7 @@ jobs:
         choco install glpk
 
     - name: Download package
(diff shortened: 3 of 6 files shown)


fneum (commented):
Looks pretty good!

The higher tolerances in this PR largely apply to all solvers. It could be an option to introduce a constant that sets the tolerance based on whether it's a GPU or CPU-based solver.

mal84emma:
Great point :) I've just added global variables to the tests that set the acceptable relative solution tolerances for CPU and GPU based solvers.

mal84emma:
@FabianHofmann Fix looks great, very happy with it, thanks for catching!

mal84emma:
@FabianHofmann absolutely, delighted to!

1. Reading the CI logs, the underlying issue is `Failed to find nvcc.` from `cmake`.
I'm not super familiar with CI workflows, but my best guess is that the machines that the workflows are being tested on don't have GPUs, and so trying to install cuPDLPx will always fail on them.
This is probably the major caveat of the GPU accelerated solvers. Trying to get a working CUDA/NVCC install is pretty painful. I've found I need to manually install CUDA and build the cuPDLPx binary locally before doing the pip install.
I'm not sure if it's possible to remove it from the CI workflow and have an experimental/hard dev warning?

2. This looks much neater, very happy to give it a look in. But to be upfront I have my PhD viva today and I'm starting a job soon, so it might be a while before I get the time to make the edits.

ollie-bell:
FWIW [cuOpt has already implemented the core concepts outlined in the cupdlpx paper](https://github.com/NVIDIA/cuopt/pull/388), and obviously is more mature and better maintained as a library. Similarly, HiGHS has also committed to a releasing built in GPU-enabled PDLP solver in the coming months (and [the implementation is well under way](https://github.com/ERGO-Code/HiGHS/tree/hi-pdlp)) which will also implement core concepts from the cupdlpx paper.

For HiGHS it goes without saying that from a CI perspective, nothing would need to change. I haven't tested it, but I think it is likely the case that cuOpt could also install and run in CI without needing a GPU (i.e. it just falls back to CPU behaviour, albeit it's linux-only).

mal84emma:
> 1. the installation of the cupdlpx solver seems broken and it is hard to understand why (see CI), do you have a clue?
> 2. we recently introduced the solver registry (see [Feature/solver features registry #528](https://github.com/PyPSA/linopy/pull/528) ) which I would also like to leverage in the test scripts to avoid things like `if solver not in feasible_mip_solvers:`  would you mind looking into this?

@FabianHofmann I've added GPU acceleration and other relevant features to the solver registry and the tests seem to be working nicely (lovely pattern, the patch was very smooth). CI testing is still an open issue, which unfortunately I don't know enough about to suggest fixes.

In other news FICO xpress just [announced GPU acceleration](https://www.fico.com/en/newsroom/fico-achieves-massive-speedups-optimization-using-nvidia-gpus), so I updated its registry entry as well.

pyproject.toml
```diff
@@ -81,6 +82,7 @@ solvers = [
     "coptpy!=7.2.1",
     "xpress; platform_system != 'Darwin' and python_version < '3.11'",
     "pyscipopt; platform_system != 'Darwin'",
+    "cupdlpx>=0.1.2",
```
ollie-bell: Btw cupdlpx isn’t regularly uploaded to pypi and what is there is sdist only (hence the build errors you’re seeing when the CI job tries to pip install it).

There has been a few months of development and multiple GitHub releases beyond the version that is available on pypi, which looks like it was manually uploaded as a one off and isn’t part of a release process.

PR #522, opened 2025-11-24
ci: fix model runs

lkstrp:
See https://github.com/PyPSA/PyPSA/pull/1443


FabianHofmann:
@lkstrp is this pr stale/unneeded?

lkstrp:
> @lkstrp is this pr stale/unneeded?

It's not, I just lost it. Thanks @brynpickering for fixing this!

brynpickering:
@lkstrp I think my latest updates should be doing the trick. Since we likely want this kind of check in multiple places (pypsa/linopy/atlite), we should consider creating a repository of reproducible actions. I did this in some past projects and it helped keep maintenance to a minimum since the actions are centralised. Only makes sense if we have a common set of them across projects.

.github/workflows/test-models.yml
```diff
@@ -43,11 +42,23 @@ jobs:
...
+          echo "WEEK=$(date +'%Y%U')" >> $GITHUB_ENV # data and cutouts
+
       # Only run check if package is not pinned
-      - name: Check if inhouse package is pinned
+      - name: Check if linopy package is pinned
         run: |
-          grep_line=$(grep -- '- pypsa' envs/environment.yaml)
-          if [[ $grep_line == *"<"* || $grep_line == *"=="* ]]; then
+          grep_line=$(grep -- '${{ github.event.repository.name }} = ' pixi.toml)
+          if [[ $grep_line == *"<* || $grep_line == *"==* ]]; then
```
lkstrp, line 61: Isn't this breaking the pin detection? Why remove the `"`

lkstrp:
> Since we likely want this kind of check in multiple places (pypsa/linopy/atlite), we should consider creating a repository of reproducible actions. I did this in some past projects and it helped keep maintenance to a minimum since the actions are centralised. Only makes sense if we have a common set of them across projects.

I do agree. I did think about it when copying all the release workflows from repo to repo a while ago, which are most likely now also not in sync anymore. Can you point me to where you set this up already? I never did

.github/workflows/test-models.yml
brynpickering: Because it's checking the pixi toml, and deps are defined differently there:

```
linopy = ">=0.6.1"
```

brynpickering:
> I do agree. I did think about it when copying all the release workflows from repo to repo a while ago, which are most likely now also not in sync anymore. Can you point me to where you set this up already? I never did

Annoyingly, I went to go to check on it and they've made the repo private 😩. I'll see if I have a clone somewhere since it was public and openly licensed at one point. EDIT: @lkstrp found it and uploaded a clone [here](https://github.com/brynpickering/actions-city-modelling-lab)

.github/workflows/test-models.yml
lkstrp: No, the bash script is broken and fails silently. `>` needs quoting and the quote is not part of the string:

``` bash
for v in ">=0.6.1" "<0.6.1" "==0.6.1"; do
    grep_line="linopy = \"$v\""

    # current
    if [[ $grep_line == *"<* || $grep_line == *"==* ]]; then
      cur="PINNED"; else cur="unpinned"
    fi

    # fixed
    if [[ $grep_line == *"<"* || $grep_line == *"=="* ]]; then
      fix="PINNED"; else fix="unpinned"
    fi

    printf '%-22s  current=%-8s  fixed=%s\n' "$grep_line" "$cur" "$fix"
  done
```

linopy = ">=0.6.1"      current=unpinned  fixed=unpinned
linopy = "<0.6.1"       current=unpinned  fixed=PINNED
linopy = "==0.6.1"      current=unpinned  fixed=PINNED

brynpickering:
@FabianHofmann I think this is now good to go, following your review. It runs as expected but fails on Zenodo retrieval. I assume this is beyond the scope of this PR to resolve?

.github/workflows/test-models.yml
FabianHofmann: @brynpickering, this is resolved?

FabianHofmann: anyway I merge the pr now as it is for sure improving

PR #524, opened 2025-11-26
Add pre-solve scaling support

FabianHofmann:
## Summary

- Add optional pre-solve scaling of constraints and continuous variables to improve numerical robustness
- Row scaling (default) rescales constraint rows by their largest (or RMS) coefficient
- Column scaling (optional) rescales continuous variables so column norms are close to 1
- Scaling is undone automatically on primal/dual values and the objective before storing results
- Integer/binary variables remain unscaled by default to preserve integrality

## API

Enable via `Model.solve(scale=...)`:
- `scale=True`: Enable row scaling with max-norm (default behavior)
- `scale="row-max"` or `scale="row-l2"`: Select norm for row scaling
- `scale=ScaleOptions(...)`: Full control over row/column scaling, integer handling, etc.

## Test plan

- [x] Unit tests for row scaling and dual unscaling
- [x] Unit tests for column scaling with continuous-only variables
- [x] Integration tests comparing scaled vs unscaled solutions
- [x] Benchmark script for performance testing


FabianHofmann:
supersed by #817

PR #528, opened 2025-12-03
Feature/solver features registry

FBumann:
## Add a registry for solver features in preparation for both quadratic constraints and SOS constraints

**Also for future proofing and documentation**

This is a proof of concept and should be discussed regarding the design and how far we should go regarding solver features (only features needed in linopy or more)

Look into the `dev-scripts/` for details and usage for now

**Important catch**:
We should separate Solver supported features and if we implemented them. DIRECT_API might be better located in the IO of each solver, as its more of an implementation thing.

## Checklist

### DIFF
--- a/linopy/solver_capabilities.py
+++ b/linopy/solver_capabilities.py
@@ -23,6 +23,7 @@ class SolverFeature(Enum):
     QUADRATIC_OBJECTIVE = auto()
 
     # I/O capabilities
+    DIRECT_API = auto()  # Solve directly from Model without writing files
     LP_FILE_NAMES = auto()  # Support for named variables/constraints in LP files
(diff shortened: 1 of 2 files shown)


linopy/model.py
```diff
@@ -1196,7 +1195,10 @@ def solve(
         if problem_fn is None:
             problem_fn = self.get_problem_file(io_api=io_api)
         if solution_fn is None:
-            if solver_name in NO_SOLUTION_FILE_SOLVERS and not keep_files:
+            if (
+                solver_supports(solver_name, SolverFeature.SOLUTION_FILE_NOT_NEEDED)
+                and not keep_files
+            ):
```
FabianHofmann: I guess here we could initialize the solver instance and support the features functions on it, instead of relying on helper functions with `solver_name` passed. but happy to keep it like this for now.

PR #534, opened 2025-12-16
Feature/dummy solve

### LINKED ISSUE #422: Add test mode that mocks solving
## Problem statement
- testing Pypsa workflows such as pypsa-eur with snakemake is difficult. Pytests doesn't integrate well with it for example. This is because snakemake jobs are subprocesses and tests conditions can't easily be passed to the subprocess
- in a pure python package, the pypsa network solve step would be mocked (e.g. return q_nom_opt = q_nom)

## Request and possible implementation
- it would be great to have an "is_test" or "mock" flag. In this case the unoptimised network would be returned with mocked opt values (linopy would not run at all)
- the argument is_test could be passed to linopy as a kwarg in network.optimize if network.optimize already takes kwargs.

Thanks!

RobbieKiwi:
Closes #422 .

Add an option to use "mock solve", to get a fake result with mock data from the linopy model for testing purposes

## Checklist

### DIFF
--- a/linopy/solvers.py
+++ b/linopy/solvers.py
@@ -550,7 +550,7 @@ class CBC(Solver[None]):
         variables = {v.name for v in h.getVariables()}
 
         def get_solver_solution() -> Solution:
-            m = re.match(r"Optimal.* - objective value (\d+\.?\d*)$", first_line)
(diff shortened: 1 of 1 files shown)


FabianHofmann (changes requested):
great, can we add a info logging that mock is running? that would be needed to remind people they should deactivate in the end

PR #539, opened 2025-12-23
allow constant in objective cost function

### LINKED ISSUE #236: Support constant values in objective
1. if the problem need to convert multi-objectives to singe-objective, it's necessary to make normalization involving some constant variables.
It's not conveniet to construct objective containing constant, such as x+1
Here is a trick to add constant
`constant_1=m.add_variables(lower=1, upper=1,name="constant_1")`
2. class Variable only have sum function, other operations are also needed, such as std

RobbieKiwi:
Closes #236 

Allow the objective cost function to contain a constant. While the constant is not needed for optimization, it can make life easier in some cases (see linked issue for discussion).

It was previously forbidden to have a constant in the objective cost function, I assume this is because most solvers expect the ocf to only contain variables.

Now the objective can contain a constant if it is explicitly allowed with 
``` python
model.add_objective(..., allow_constant=True)
```
Otherwise adding an objective with a constant will result in an error.

A bit of refactoring was required. To avoid having to repeat the same logic on multiple code paths and touch too much of the core solve logic, I made a decorator which strips the constant term before calling the `Model.solve` method and then adds it back again after.
``` python
@strip_and_replace_constant_objective
def solve(
    self,
    ...
```
## Checklist

### DIFF
--- a/linopy/model.py
+++ b/linopy/model.py
@@ -102,22 +102,21 @@ def strip_and_replace_constant_objective(func: Callable[P, R]) -> Callable[P, R]
         assert isinstance(self, Model), (
             f"First argument must be a Model instance, got {type(self)}"
         )
-        model = self
-        if not self.objective.has_constant:
+        model: Model = self
+        if not model.objective.has_constant:
             # Continue as normal if there is no constant term
             return func(*args, **kwargs)
 
         # The objective contains a constant term
         # Modify the model objective to drop the constant term
-        model = self
-        constant = float(self.objective.expression.const.values)
-        model.objective.expression = self.objective.expression.drop_constant()
+        constant = float(model.objective.expression.const.values)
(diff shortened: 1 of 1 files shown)


FabianHofmann (commented):
@RobbieKiwi thank you for the pr! I need a bit more time to think about the approach you've taken. I would rather like to have the constant integrated in the full pipeline, such that it appears at the solver side as well (and skip those which are not supporting it). because now you would always see a discrepancy between the linopy objective and the solver objective which on the one hand is confusing, on the other make debugging other things harder.

RobbieKiwi:
@FabianHofmann what if we added a dummy variable and constrained it to be a certain value and included it in the objective cost function seen by the solver?
Then I guess it would be consistent from end to end apart from the fact we would add/remove the dummy variable before and after solving

PR #544, opened 2026-01-03
Fix -0.0 handling in lp writer

### LINKED ISSUE #541: bug: lp-file error: Unrecognized variable bound

- [ ] I have confirmed this bug exists on the lastest [release](https://github.com/pypsa/linopy/releases) of Linopy.

### Issue Description

When solving a linopy model through the lp-api get an error from gurobipy:
```
Set parameter Username
Academic license - for non-commercial use only - expires 2026-12-18
Error reading LP format file /private/var/folders/7j/18_93__x4wl2px44pq3f570m0000gn/T/linopy-problem-zxm32_f3.lp at line 21326335
Unrecognized variable bound
Neighboring tokens: " +-0.0 <= x76271 <= +0.0 +-0.0 <= x76272 <= "
```

### Reproducible Example

```python
Working on it
```

### Expected Behavior

Solve as expected (-0.0 == +0.0)

### Installed Versions

<details>
Replace this line.
</details>

FBumann:
Closes #541

Bug: LP file writer produced invalid syntax +-0.0 when bounds/coefficients were -0.0, causing Gurobi to reject the file.

Root cause: -0.0 >= 0 is True, so a + prefix was added, but str(-0.0) is "-0.0", producing +-0.0.

Fix: Normalize -0.0 to +0.0 using when(abs(x) == 0).then(0.0).otherwise(x) before formatting.

Performance: ~4% overhead on 10M rows

Changes:
- linopy/io.py: Added signed_number_expr() helper, applied to 4 locations
- test/test_io.py: Added 7 test cases

## Benchmark script

```python
#!/usr/bin/env python3
"""
Benchmark script for LP file writing performance.

Can be used ti check the impact of the negative zero fix on LP file generation speed.
"""

import tempfile
import time
from pathlib import Path

import numpy as np

from linopy import Model


def create_model(n: int) -> Model:
    """Create a test model with n^2 variables."""
    m = Model()
    N = np.arange(n)
    x = m.add_variables(coords=[N, N], name="x")
    y = m.add_variables(coords=[N, N], name="y")
    m.add_constraints(x - y >= N, name="c1")
    m.add_constraints(x + y >= 0, name="c2")
    m.add_objective((2 * x).sum() + y.sum())
    return m


def benchmark_lp_writing(sizes: list[int], iterations: int = 10) -> None:
    """Benchmark LP file writing for different model sizes."""
    print(f"Benchmarking LP file writing ({iterations} iterations each)")
    print("-" * 60)

    with tempfile.TemporaryDirectory() as tmpdir:
        for n in sizes:
            m = create_model(n)
            nvars = m.nvars
            ncons = m.ncons

            # Warmup
            fn = Path(tmpdir) / f"warmup_{n}.lp"
            m.to_file(fn, progress=False)

            # Benchmark
            times = []
            for i in range(iterations):
                fn = Path(tmpdir) / f"bench_{n}_{i}.lp"
                start = time.perf_counter()
                m.to_file(fn, progress=False)
                elapsed = time.perf_counter() - start
                times.append(elapsed)

            avg = np.mean(times)
            std = np.std(times)
            print(
                f"N={n:4d} ({nvars:>8,} vars, {ncons:>8,} cons): "
                f"{avg*1000:7.1f}ms ± {std*1000:5.1f}ms"
            )


if __name__ == "__main__":
    # Test with various model sizes
    sizes = [50, 100, 200, 500, 1000]
    benchmark_lp_writing(sizes, iterations=10)

```

## Checklist

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -36,7 +36,7 @@ jobs:
     permissions:
       id-token: write
     steps:
-      - uses: actions/download-artifact@v6
+      - uses: actions/download-artifact@v7
         with:
           name: Packages
           path: dist
--- a/.github/workflows/test-models.yml
+++ b/.github/workflows/test-models.yml
@@ -59,7 +59,7 @@ jobs:
           echo -ne "url: ${CDSAPI_URL}\nkey: ${CDSAPI_TOKEN}\n" > ~/.cdsapirc
           echo "week=$(date +'%Y%U')" >> $GITHUB_ENV # data and cutouts
 
-      - uses: actions/cache@v4
+      - uses: actions/cache@v5
         if: env.pinned == 'false'
         with:
           path: |
@@ -74,7 +74,7 @@ jobs:
 
       - name: Cache Conda env
         if: env.pinned == 'false'
-        uses: actions/cache@v4
+        uses: actions/cache@v5
         with:
           path: ${{ env.CONDA }}/envs
           key: conda-pypsa-eur-${{ env.week }}-${{ hashFiles('envs/linux-64.lock.yaml') }}
@@ -101,7 +101,7 @@ jobs:
 
       - name: Upload artifacts
         if: env.pinned == 'false'
-        uses: actions/upload-artifact@v5
+        uses: actions/upload-artifact@v6
         with:
           name: results-pypsa-eur-${{ matrix.version }}
           path: |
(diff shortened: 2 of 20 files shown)


coroa (changes requested):
I like the idea of a signed_number function, but i would let it directly act on expressions and return a simple expression.

This expression could then be a simple case statement.

PR #551, opened 2026-01-20
Redesign as_dataarray: strict public API + internal ensure_dataarray helper

FBumann:
## Summary

Redesigns `as_dataarray()` by splitting it into two functions with clear responsibilities:

- **`as_dataarray(arr, coords?, dims?)`** — Public API. When `coords` is provided: converts to DataArray, validates shared dim coords match exactly, rejects extra dims, broadcasts missing dims via `expand_dims`. When `coords` is `None`: pure type conversion only.
- **`ensure_dataarray(arr, coords?, dims?)`** — Internal helper. Pure type conversion (scalar, numpy, pandas, polars, list → DataArray). No validation, no `expand_dims`. Coords are used only as construction hints. Callers handle alignment themselves (xarray broadcasting, `reindex_like`, etc.).

Both share a common `_type_dispatch` function for the actual type conversion logic.

## Motivation

Previously `as_dataarray()` mixed two conflicting roles:
1. **Strict validation** for `add_variables` where user-provided bounds must match coords exactly
2. **Lenient conversion** in expression arithmetic where xarray alignment handles coord reconciliation

This led to bugs where DataArray inputs to `add_variables` had their coords silently ignored, and attempts to fix it broke expression arithmetic.

## How variable coordinates are determined

| Scenario | Behavior |
|----------|----------|
| `coords` provided, bounds are scalars/numpy/pandas | `coords` defines the variable's coordinate space; inputs are converted using `coords` |
| `coords` provided, bounds are DataArrays | `coords` defines the variable's coordinate space; bounds are **validated** against `coords` (shared dims must match, extra dims raise `ValueError`, missing dims are broadcast) |
| `coords` is `None`, bounds are DataArrays | Coordinates are inferred from the DataArray bounds directly |
| `coords` is `None`, bounds are scalars | Scalar (dimensionless) variable is created |

## Changes

| File | Change |
|------|--------|
| `linopy/common.py` | Extract `_type_dispatch` (shared conversion), `_expand_missing_dims`, `_validate_dataarray_coords` (strict, no `allow_extra_dims` flag). Add `ensure_dataarray`. Refactor `as_dataarray` to delegate to `ensure_dataarray` when no coords, or validate+expand when coords provided. |
| `linopy/expressions.py` | Switch all call sites to `ensure_dataarray` (arithmetic, constants, constraint comparisons, `dot`) |
| `linopy/variables.py` | Switch `Variable.to_linexpr` to `ensure_dataarray` |
| `linopy/model.py` | `add_variables` lower/upper use `as_dataarray` (strict). Masks, sign, rhs use `ensure_dataarray` (lenient). |
| `test/test_common.py` | Tests for `_validate_dataarray_coords`, `ensure_dataarray` (no expand, allows extra dims, no coord validation), and `add_variables` integration. |
| `doc/release_notes.rst` | Release note entry. |

### Mask handling

The `mask` parameter uses `ensure_dataarray` (no validation). The existing `broadcast_mask` function handles misaligned mask coords by filling with `False`. A TODO is left for a future deprecation path toward stricter validation (see #591).

## Examples

```python
# Mismatched coords → ValueError
lower = xr.DataArray([0, 0, 0], dims=["time"], coords={"time": [0, 1, 2]})
m.add_variables(lower=lower, coords=[pd.RangeIndex(5, name="time")], name="x")
# ValueError: Coordinates for dimension 'time' do not match

# Extra dims → ValueError
lower = xr.DataArray([[1, 2], [3, 4]], dims=["x", "y"])
m.add_variables(lower=lower, coords={"x": [0, 1]}, name="x")
# ValueError: DataArray has extra dimensions not in coords: {'y'}

# Subset dims → broadcast
lower = xr.DataArray([0, 0], dims=["x"], coords={"x": [0, 1]})
m.add_variables(lower=lower, coords={"x": [0, 1], "y": [0, 1, 2]}, name="x")
# works, lower broadcast to shape (2, 3)

# No coords → pass through unchanged
m.add_variables(lower=lower, name="x")

# Internal arithmetic uses ensure_dataarray (lenient)
expr = m.variables["x"] + np.array([1, 2])  # xarray handles alignment
```

## Checklist

- [x] Code changes are sufficiently documented
- [x] Unit tests pass
- [x] A note for the release notes
- [x] I consent to the release of this PR's code under the MIT license

### DIFF
--- a/linopy/common.py
+++ b/linopy/common.py
@@ -140,8 +140,28 @@ def _coords_to_mapping(coords: CoordsLike, dims: DimsLike | None = None) -> Mapp
     seq = list(coords)
     if dims is not None:
         dim_names: list[str] = [dims] if isinstance(dims, str) else list(dims)  # type: ignore[arg-type]
+        if len(dim_names) != len(seq):
+            raise ValueError(
+                f"Length of dims ({len(dim_names)}) does not match "
+                f"length of coords sequence ({len(seq)})."
+            )
         return dict(zip(dim_names, seq))
-    return {c.name: c for c in seq if hasattr(c, "name") and c.name}
+    result: dict = {}
+    for c in seq:
+        if not hasattr(c, "name") or not c.name:
+            warn(
+                f"Coordinate {c!r} has no .name attribute and will be ignored. "
+                "Pass dims explicitly or use named indexes (e.g. pd.RangeIndex).",
+                UserWarning,
+                stacklevel=3,
+            )
+            continue
+        if c.name in result:
+            raise ValueError(
+                f"Duplicate coordinate name '{c.name}' in coords sequence."
+            )
+        result[c.name] = c
+    return result
 
 
 def _validate_dataarray_coords(
--- a/test/test_common.py
+++ b/test/test_common.py
@@ -457,6 +457,30 @@ def test_validate_dataarray_coords_sequence_mismatch() -> None:
         _validate_dataarray_coords(da, coords=[idx], dims=["x"])
 
 
+def test_coords_to_mapping_dims_length_mismatch() -> None:
+    from linopy.common import _coords_to_mapping
+
+    with pytest.raises(ValueError, match="Length of dims"):
+        _coords_to_mapping([pd.RangeIndex(3), pd.RangeIndex(5)], dims=["x"])
+
+
+def test_coords_to_mapping_unnamed_index_warns() -> None:
+    from linopy.common import _coords_to_mapping
+
+    with pytest.warns(UserWarning, match="no .name attribute"):
+        result = _coords_to_mapping([np.array([1, 2, 3])])
+    assert result == {}
+
+
+def test_coords_to_mapping_duplicate_name() -> None:
+    from linopy.common import _coords_to_mapping
+
+    idx1 = pd.RangeIndex(3, name="x")
+    idx2 = pd.RangeIndex(5, name="x")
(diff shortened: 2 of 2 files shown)


FBumann:
Maybe remove explicit broadcast from masking added in a prior patch

FBumann (commented):
The validation in `_validate_dataarray_coords` only runs inside `add_variables`, but `as_dataarray` itself still ignores the `coords` parameter for pandas and DataArray inputs:

```python
# pandas: warns but keeps original index
s = pd.Series([10,20,30], index=pd.Index([5,6,7], name="x"))
as_dataarray(s, coords={"x": [0,1,2]})  # returns [5,6,7] coords, just a warning

# DataArray: passes through unchanged
da = xr.DataArray([10,20,30], dims=["x"], coords={"x": [5,6,7]})
as_dataarray(da, coords={"x": [0,1,2]})  # returns [5,6,7] coords, silent

# scalar: crashes
as_dataarray(5.0, coords={"x": [0,1,2]})  # CoordinateValidationError
```

So `as_dataarray` still can't be relied on to produce correctly-coordinated arrays. The validation only catches mismatches for DataArray bounds in `add_variables` — not for parameters used in expressions or constraints, which is where most data flows through.

Would it make more sense to have `as_dataarray` itself do the alignment (reindex/broadcast to `coords` when provided), rather than just validating downstream in `add_variables`? That way any caller passing `coords` gets correct output regardless of input type.

PR #555, opened 2026-01-24
Add auto_mask parameter to Model class

FBumann:
## Disclaimer

Im not sure how much value this adds, as i didnt use masking that much before. Im happy to discuss the ups and downsides of such a feature. I would have raised an issue, but i thought adding a benchmark and proof of concept would be more valuable.

## Summary

Adds `auto_mask` parameter to `Model` that automatically masks variables and constraints where input data contains NaN values.

```python
# Before: manual masking required
m = Model()
mask = gen_capacity.notnull()
x = m.add_variables(lower=0, upper=gen_capacity, mask=mask, name="x")

# After: automatic
m = Model(auto_mask=True)
x = m.add_variables(lower=0, upper=gen_capacity, name="x")  # NaN auto-masked
```

## Benefits

- **Convenience**: No manual mask creation/tracking
- **Safety**: Can't forget to mask NaN values
- **Performance**: Uses optimized numpy internals (`np.where` instead of xarray `where` = 38x faster)

## Auto-mask conditions

**Variables** masked out where:
- `lower` is NaN, OR
- `upper` is NaN

**Constraints** masked out where:
- All variable references are invalid (null expression), OR
- `rhs` is NaN

### Variables
| Potential | Active | No Mask | Manual | Auto |
|-----------|--------|---------|--------|------|
| 10K | 7K | 14ms | 2ms | 2ms |
| 100K | 70K | 2ms | 2ms | 2ms |
| 500K | 350K | 3ms | 3ms | 3ms |
| 2M | 1.4M | 6ms | 6ms | 5ms |

### Constraints
| Count | Terms | No Mask | Manual | Auto |
|-------|-------|---------|--------|------|
| 5K | 500 | 17ms | 16ms | 14ms |
| 10K | 1000 | 44ms | 45ms | 30ms |
| 20K | 1000 | 109ms | 61ms | 46ms |
| 50K | 500 | 171ms | 101ms | 61ms |

Auto-mask is **equal or faster** than manual masking due to optimized numpy operations.

## Checklist

- [x] Code documented
- [x] Tests added (62 pass)
- [x] Release notes added
- [x] MIT license consent

### DIFF
--- a/doc/release_notes.rst
+++ b/doc/release_notes.rst
@@ -7,10 +7,7 @@ Upcoming Version
 * Fix docs (pick highs solver)
 * Add the `sphinx-copybutton` to the documentation
 * Add ``auto_mask`` parameter to ``Model`` class that automatically masks variables and constraints where bounds, coefficients, or RHS values contain NaN. This eliminates the need to manually create mask arrays when working with sparse or incomplete data.
-
-Upcoming Version
-----------------
-
+* Speed up LP file writing by 2-2.7x on large models through Polars streaming engine, join-based constraint assembly, and reduced per-constraint overhead
 * Fix multiplication of constant-only ``LinearExpression`` with other expressions
 
 Version 0.6.1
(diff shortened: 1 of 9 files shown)


FabianHofmann (commented):
I very much like this, it is non-intrusive and sensible to try out in the future

PR #557, opened 2026-01-24
Feature/example of drop=True

### LINKED ISSUE #464: Add example of using `where`  with `drop = True`
> Apologies to jump into an issue but I have some thoughts that are related to sparsity so thought I'd add to this discussion rather than create a new one for now.
> 
> 
> 
> I've just started using linopy (and not from a linear programming background either but have used xarray extensively in other contexts) and came across behaviour which I found unintuitive - which is the default behaviour of `LinearExpression.where`. I found I needed to pass `drop=True` to get rid of "null" terms in the equations (i.e. significantly reduce the size of the `_term` dimension)... but my intuition as a new user would have expected this to be the default.
> 
> 
> 
> So the question / discussion point is what are the reasons for `drop=True` not being default in `LinearExpression.where` (and presumably `Variable.where` as well but I've not thought about that so much yet). Are there situations where a user would want to keep entirely null terms in their equations? Or is it simply to align with the xarray defaults?
> 
> 
> 
> EDIT
> 
> ----
> 
> On further thought I see why `drop=False` remains the default, because now in application code I need to deal with the case of `drop=True` resulting in misaligned / missing coordinates relative to other data. However, I think the documentation would benefit from a small example of using `drop=True`. Happy to open a PR if that is of interest?
> 
> 
> 
> Perhaps also some indication in the `__repr__` of `LinearExpression` of the sparsity. I think most of my confusion came from the fact that the `__repr__` hides all the `nan` coeffs, but they are still present in the `_terms` dimension of the underlaying data. 

 _Originally posted by @ollie-bell in [#248](https://github.com/PyPSA/linopy/issues/248#issuecomment-2348983408)_

RobbieKiwi:
Closes #464

Add an example to the docs of `LinearExpression.where(..., drop=True)`
Add new property to base expression `BaseExpression.variable_names`

## Not addressed / further ideas
In the issue mentioned, I think the reporter wanted to see the unused terms in `LinearExpression.__repr__`. After having a bit of a look at it, I think this is impossible because the term dim has no coordinate names. There seems to no way to know what variable originally belonged in a position of the term dim if the original variable has been completely masked (replaced with -1).

Maybe when drop is True we could also run `.simplify` to get rid of any empty positions in the term dim?
Currently `drop=True` will drop unused index positions along all dimensions apart from the term dim, it might make more sense to drop on all dims?

## Checklist


FabianHofmann (changes requested):
thanks for the pr @RobbieKiwi, one comment

FBumann:
Hi @RobbieKiwi 
Why close this?

RobbieKiwi:
I get an email from GitHub about this every day for 4 months and I don't know another way to stop it

FBumann:
Understood. But such an example would still be usefull right? WOuld you mind reopening an issue and adding it back to the milestone?

Issue #570, opened 2026-02-06
Inconsistent behavior when multiplying with subset coordinates


FabianHofmann:
### Version Checks

- [x] I have confirmed this bug exists on the current `master` branch of Linopy.

### Issue Description

Multiplying variables/expressions with constants that have a subset of coordinates produces inconsistent behavior:

1. `Variable * subset_const`: Creates NaN coefficients with valid variable labels
2. `LinearExpression * subset_const`: Fails with AssertionError in `_multiply_by_constant`
3. `Variable + subset_const`: Works correctly (fills missing with 0)

The root cause is that addition uses `merge()` with controlled `join='outer'` and `fill_value=0`, while multiplication uses uncontrolled xarray operations.

### Reproducible Example

```python
import xarray as xr
import pandas as pd
import linopy

m = linopy.Model()
x = m.add_variables(coords=[pd.Index(range(5), name='i')])
subset = xr.DataArray([10.0, 30.0], dims=['i'], coords={'i': [1, 3]})

# Works but produces NaN coefficients with valid var labels
result1 = x * subset
print(result1.coeffs.squeeze().values)  # [nan, 10., nan, 30., nan]
print(result1.data.vars.squeeze().values)  # [0, 1, 2, 3, 4] - all valid!

# Fails with AssertionError
expr = 1 * x
result2 = expr * subset  # AssertionError in _multiply_by_constant

# Addition works correctly
result3 = x + subset
print(result3.const.values)  # [0., 10., 0., 30., 0.]
```

### Expected Behavior

Multiplication should behave consistently:
- Either fill missing coordinates with 0 (like addition)
- Or raise an error for misaligned coordinates
- Both `x * c` and `(1*x) * c` should behave the same

### Installed Versions

<details>

linopy @ master (commit 59f92ae)

</details>

FBumann:
Re-tested on linopy 0.7.0: the inconsistency reported here — `Variable * subset` producing NaN coefficients, `LinearExpression * subset` raising `AssertionError`, the two paths disagreeing — is **fixed**. Both now give `[0, 10, 0, 30, 0]` consistently, no crash:

```python
import pandas as pd, xarray as xr, linopy
m = linopy.Model()
x = m.add_variables(coords=[pd.RangeIndex(5, name="i")])
subset = xr.DataArray([10.0, 30.0], dims=["i"], coords={"i": [1, 3]})
(x * subset).coeffs.squeeze().values        # [ 0. 10.  0. 30.  0.]
[...]

PR #571, opened 2026-02-06
Make LinearExpression multiplication with constant containing subset of dimensions consistent with Variable behaviour

### LINKED ISSUE #569: Inconsistent multiplication behavior: Variable vs LinearExpression with subset coordinates
### Issue Description

Multiplying with a constant that has a subset of coordinates on a shared dimension behaves differently for `Variable` vs `LinearExpression`:

- `x * subset_da` works (produces NaN coefficients)
- `(1 * x) * subset_da` fails with `AssertionError`

Mathematically these should be equivalent.

### Reproducible Example

```python
import pandas as pd
import xarray as xr
import linopy

m = linopy.Model()
x = m.add_variables(coords=[pd.Index(range(5), name="i")])

subset_da = xr.DataArray([10.0, 30.0], dims=["i"], coords={"i": [1, 3]})

# This works
result1 = x * subset_da
print(result1.coeffs.values)  # [nan, 10., nan, 30., nan]

# This fails with AssertionError
expr = 1 * x
result2 = expr * subset_da  # AssertionError in _multiply_by_constant
```

### Expected Behavior

Both operations should behave consistently. Either:
1. Both succeed with same result, or
2. Both raise an error indicating misaligned coordinates

### Installed Versions

<details>

master branch (commit 59f92ae)

</details>

RobbieKiwi:
Closes #569 #570 

Now LinearExpression behaves the same way as Variable when multiplying with a constant containing a susbet of dimensions. Still it is not consistent with addition.

| Action  | Coords | Missing coord behaviour  |
| ------------- | ------------- |------------- |
| Addition | Same coords  | add 0 |
| Variable Multiplication | Same coords   | nan  |
| Expression Multiplication | Same coords   | nan  |

For addition, when a coord is mising we just add 0 (I.e do nothing at that coord)
Perhaps the equivalent for multiplication would be to multiply by 1 (also do nothing) rather than fill with nan?

## Checklist

### DIFF
--- a/doc/release_notes.rst
+++ b/doc/release_notes.rst
@@ -4,6 +4,7 @@ Release Notes
 Upcoming Version
 ----------------
 
+* Fix the handling of multiplication between ``LinearExpression`` and constants with a subset of dimensions. Align with ``Variable`` behaviour
 * Fix docs (pick highs solver)
 * Add the `sphinx-copybutton` to the documentation
 * Add ``auto_mask`` parameter to ``Model`` class that automatically masks variables and constraints where bounds, coefficients, or RHS values contain NaN. This eliminates the need to manually create mask arrays when working with sparse or incomplete data.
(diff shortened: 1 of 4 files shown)


RobbieKiwi:
@FabianHofmann I like your PR #572 better so I will close this one

PR #575, opened 2026-02-09
fix: revert np.where to xarray.where when adding vars/ constraints

lkstrp:
#555 introduced masking and used `np.where` instead of `xarray.where` since is is "38x faster than xarray where".

But `xarray` is slower for a reason, since it aligns operands by dimension name, not just position. `np.where` requires shapes to match positionally, which is not guaranteed in `add_variables` or `add_constraints`. This PR just reverts the changes, we should add tests and release a patch tomorrow.

See https://github.com/PyPSA/linopy/pull/555/changes#diff-1a9bf239769f2af33c90e00a21f7f08e2f40d626ce06ee470bb2daa65ac06a04R788-R789


FBumann:
@lkstrp Sorry if this causes troubles.

FabianHofmann:
> @lkstrp Sorry if this causes troubles.

no this is rather on me, I though we had the internal constraint already that masks have to have the same shape as constraints/variables

FabianHofmann:
sorry, for the gibberish in the last comment, I was in a meeting. I would say let's revert this, in particular given the failing pypsa functions. then updating linopy to the latest and keeping pypsa v1.0.x does not break. But I would like to require `mask`  to be properly aligned in the first place. Linopy already has so many reindexing logics that you sometimes don't know what's going on...

PR #580, opened 2026-02-10
Reinsert broadcasted mask

FabianHofmann:
Sorry for the back and forth. 

This reverts the FutureWarning from #579 and re-enables `np.where` (from #555) for applying masks, which indeed is significantly faster than `xr.where`.

The root issue was that DataArray masks weren't broadcast to match data.labels, unlike masks passed as pd.Series, pd.DataFrame, or np.ndarray (which go through `as_dataarray` broadcasting). This is now handled by a new `broadcast_mask` helper that calls `broadcast_like` to align/transpose the mask, then detects and warns about missing coordinates, falls back to NaN fill with False.

Dimension validation is now again a hard assert (mask.dims ⊆ data.dims) instead of a FutureWarning.

## Checklist

### DIFF
--- a/linopy/common.py
+++ b/linopy/common.py
@@ -287,6 +287,16 @@ def as_dataarray(
 
 
 def broadcast_mask(mask: DataArray, labels: DataArray) -> DataArray:
+    """
+    Broadcast a boolean mask to match the shape of labels.
+
+    Ensures that mask dimensions are a subset of labels dimensions, broadcasts
+    the mask accordingly, and fills any NaN values (from missing coordinates)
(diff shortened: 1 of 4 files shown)


FBumann:
@FabianHofmann See #581 for small modifications. Even without: LGTM

lkstrp:
>  Thanks for your trust :). Maybe we can setup a call at some point and get to know each other, and then discuss this (briefly)?

@FBumann  We are happy to have a call to meet and align! Just send me an email again with some dates this or next week or whenever it fits you. @FabianHofmann will join as well

PR #583, opened 2026-02-11
feat: add unified SolverMetrics

### LINKED ISSUE #428: Achieved gap of the objective
## Feature ##
It would be good to have the achieved gap of the objective in the results. Is there a way to get it (independently of the used solver)?
## Background ##
I do many calculations and want to check afterwards in postprocessing routines how well the solver worked.

FBumann:
## Summary

Add a unified `SolverMetrics` dataclass that provides solver-independent access to performance metrics after solving. Accessible via `Model.solver_metrics`.

**Fields:** `solver_name`, `solve_time`, `objective_value`, `dual_bound`, `mip_gap`, `peak_memory` — all default to `None`; solvers populate what they can.

**Design:**
- Frozen dataclass (immutable after creation)
- Base `Solver._extract_metrics()` populates `solver_name` and `objective_value`
- Each solver subclass overrides to add solver-specific fields via `dataclasses.replace()`
- All attribute access wrapped in `_safe_get()` with debug logging so extraction never breaks the solve
- The pattern is easily extensible for new solvers — just override `_extract_metrics()` and use `_safe_get()`. PRs adding metrics for additional solvers are welcome!

**Solver coverage:**

| Solver | solve_time | dual_bound | mip_gap | peak_memory | tested |
|--------|:---:|:---:|:---:|:---:|:---:|
| Gurobi | ✓ | ✓ | ✓ | ✓ | ✓ |
| HiGHS | ✓ | ✓ | ✓ | | ✓ |
| SCIP | ✓ | ✓ | ✓ | | ✓ |
| CPLEX | ✓ | ✓ | ✓ | | ✓ |
| Xpress | ✓ | ✓ | ✓ | ✓ | ✓ |
| Mosek | ✓ | ✓ | ✓ | | ✓ |
| CBC | ✓* | | ✓* | | |
| COPT | base only | | | | |
| MindOpt | base only | | | | |
| cuPDLPx | base only | | | | |
| GLPK | base only | | | | |

*CBC parses `solve_time` and `mip_gap` from log output via regex and populates them in `SolverMetrics`. These fields depend on the CBC log format, which varies across versions.

`peak_memory` (MB) is only available for solvers that expose it via their Python API. Verified that HiGHS, SCIP, CPLEX, Mosek, and CBC do **not** expose peak memory. Gurobi provides it via `MaxMemUsed` (GB, converted to MB) and Xpress via `peakmemory` (bytes, converted to MB).

Solvers without a tested `_extract_metrics` override still get `solver_name` and `objective_value` from the base class. We intentionally did not add solver-specific overrides for untested solvers — incorrect attribute names would silently return `None`, giving a false sense of coverage.

**Bugs fixed along the way:**
- **HiGHS**: Used non-existent info key `mip_objective_bound` — fixed to `mip_dual_bound`. Also fixed status comparison (`== 0` → `== highspy.HighsStatus.kOk`).
- **Xpress**: `miprelgap` attribute doesn't exist — compute gap manually from `mipbestobjval` and `bestbound`.
- **CPLEX**: `m.solution.progress.get_time()` doesn't exist — use `time.perf_counter()` around the solve call.

## Test plan
- [x] `SolverMetrics` dataclass unit tests (defaults, partial, repr, frozen)
- [x] `Result` backward compatibility tests (with/without metrics)
- [x] `Model` integration tests (metrics before solve, after mock solve, after reset, after re-solve with different solver)
- [x] Parametrized LP tests over direct and file-IO solvers
- [x] Parametrized MIP tests asserting `mip_gap` and `dual_bound` are populated
- [x] Parametrized `peak_memory` test for Gurobi and Xpress
- [x] All solver-specific overrides tested with real solves (Gurobi, HiGHS, SCIP, CPLEX, Xpress, Mosek)

Closes #428

### DIFF
--- a/doc/release_notes.rst
+++ b/doc/release_notes.rst
@@ -4,6 +4,8 @@ Release Notes
 Upcoming Version
 ----------------
 
+* Add unified ``SolverMetrics`` dataclass accessible via ``Model.solver_metrics`` after solving. Provides ``solver_name``, ``solve_time``, ``objective_value``, ``best_bound``, and ``mip_gap`` in a solver-independent way. Gurobi, HiGHS, SCIP, and CBC populate solver-specific fields; other solvers provide the baseline.
+
 Version 0.6.3
 --------------
 
--- a/doc/release_notes.rst
+++ b/doc/release_notes.rst
@@ -4,7 +4,7 @@ Release Notes
 Upcoming Version
 ----------------
 
-* Add unified ``SolverMetrics`` dataclass accessible via ``Model.solver_metrics`` after solving. Provides ``solver_name``, ``solve_time``, ``objective_value``, ``best_bound``, and ``mip_gap`` in a solver-independent way. Gurobi, HiGHS, SCIP, and CBC populate solver-specific fields; other solvers provide the baseline.
+* Add unified ``SolverMetrics`` dataclass accessible via ``Model.solver_metrics`` after solving. Provides ``solver_name``, ``solve_time``, ``objective_value``, ``best_bound``, and ``mip_gap`` in a solver-independent way. All solvers populate solver-specific fields where available.
 
 Version 0.6.3
 --------------
--- a/doc/release_notes.rst
+++ b/doc/release_notes.rst
@@ -12,7 +12,7 @@ Upcoming Version
 * Add the `sphinx-copybutton` to the documentation
 * Add SOS1 and SOS2 reformulations for solvers not supporting them.
 * Enable quadratic problems with SCIP on windows.
-* Add unified ``SolverMetrics`` dataclass accessible via ``Model.solver_metrics`` after solving. Provides ``solver_name``, ``solve_time``, ``objective_value``, ``best_bound``, and ``mip_gap`` in a solver-independent way. All solvers populate solver-specific fields where available.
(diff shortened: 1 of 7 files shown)


FBumann:
Covereage fails, as not all solvers are in CI. But this is a change which might affect CI in general.

FBumann:
#428

lkstrp (commented):
I would really like to see something like this. Not sure though if this would be enough to reliable benchmark the model in automated runs, but with a fixed node/ environment I don't see why this wouldn't already be helpful. Any thoughts on getting memory usage as well? I think gurobi gives you peak memory, not sure for the other solvers

FBumann:
> I would really like to see something like this. Not sure though if this would be enough to reliable benchmark the model in automated runs, but with a fixed node/ environment I don't see why this wouldn't already be helpful. Any thoughts on getting memory usage as well? I think gurobi gives you peak memory, not sure for the other solvers

I tried to design this in a way to be extensible, add new attributes and populate them for solvers that provide them.
I dont know how/which solvers monitor this, but feel free to modify my PR as you wish and try if the pattern feels as extensible as i want it to be.

FBumann:
> I would really like to see something like this. Not sure though if this would be enough to reliable benchmark the model in automated runs, but with a fixed node/ environment I don't see why this wouldn't already be helpful. Any thoughts on getting memory usage as well? I think gurobi gives you peak memory, not sure for the other solvers

@lkstrp My concern is not benchmarking here. Im mostly solving MILP programs, and having the MIP GAP is very relevant. I need to save this alongside the solution. I think there are others that would benefit from such. Memory usage on the other hand is purely benchmarking imo.

FabianHofmann:
closed by #682

Issue #586, opened 2026-02-13
Constraints do not match RHS based on coordinates


thomgeo:
### Version Checks (indicate both or one)

- [ ] I have confirmed this bug exists on the lastest [release](https://github.com/pypsa/linopy/releases) of Linopy.

- [ ] I have confirmed this bug exists on the current [`master`](https://github.com/pypsa/linopy/tree/master) branch of Linopy.

### Issue Description

When constants are supplied to a linopy model, the model appears to introduce these only based on order, not match them based on the relevant coordinates. 

### Reproducible Example

```python
import linopy

m = linopy.Model()

idx = ["A1", "A5", "A11", "A100"]

x = m.add_variables(coords=[idx], name="x")

idx2 = ["A100", "A1", "A5", "A11"]

coef = pd.Series({i: int(i[1:]) for i in idx2})
m.add_constraints(x >= coef)
```

### Expected Behavior

linopy should build the constraint so that the variables are matched with the constant of the same index. 

```
[A1]: +1 x[A1]     ≥ 100.0

[A5]: +1 x[A5]     ≥ 1.0

[A11]: +1 x[A11]   ≥ 11.0

[A100]: +1 x[A100] ≥ 5.0
```

Instead, it returns a constraint of the form:

```
[A1]: +1 x[A1]     ≥ 1.0

[A5]: +1 x[A5]     ≥ 5.0

[A11]: +1 x[A11]   ≥ 11.0

[A100]: +1 x[A100] ≥ 100.0
```

### Installed Versions

<details>
0.5.8</details>

PR #572, opened 2026-02-09
refac: introduce consistent convention for linopy operations with subsets and supersets

FabianHofmann:
Related to #550 #571 

## Changes proposed in this Pull Request

Establish a consistent coordinate alignment when linopy objects interact with DataArrays that have subset or superset coordinates

# TODO: 

- [x] extend test suit to linopy + linopy objects
- [x] manifest convention in the docs
- [x] clarify relation to `linopy.align`
- [x] really ensure convention holds true for all combinations of `addition`/`multiplication`/`comparison` + `subset`/superset` + order + object types  
- [x] come up with a convention for equally shaped objects #550 

## Checklist

- [x] Code changes are sufficiently documented; i.e. new functions contain docstrings and further explanations may be given in `doc`.
- [x] Unit tests for new features were added (if applicable).
- [x] A note for the release notes `doc/release_notes.rst` of the upcoming release is included.
- [x] I consent to the release of this PR's code under the MIT license.

PR #550, opened 2026-01-20
Fix coordinate misalignment in expression merge

FBumann:
## Summary

Fixes silent data corruption when adding expressions whose coordinates have the same values but in different order (e.g. `['costs', 'Penalty']` vs `['Penalty', 'costs']`). Previously, `merge()` used positional concatenation via `join='override'`, ignoring labels entirely.

## What changed

When `merge()` detects that two datasets share the same coordinate values in a different order, it now reindexes to the first dataset's order before concatenation. This only triggers when the coordinate value **sets are identical** — different subsets or different sizes are unaffected.

## What is NOT affected

- Expressions with **identical coordinates** (same values, same order) — fast override path, unchanged
- Expressions with **different dimension sizes** — outer join path, unchanged  
- Positional patterns like `v.loc[:9] + v.loc[10:]` — different value sets, override preserved

<details>
<summary>Bug reproduction script</summary>

```python
import pandas as pd
import linopy

m = linopy.Model()
var = m.add_variables(coords=[pd.Index(['costs', 'Penalty'], name='effect')], name='var')
factors = pd.Series([2.0, 1.0], index=pd.Index(['Penalty', 'costs'], name='effect'))

result = var - 10 * factors

costs = float(result.const.sel(effect='costs'))
penalty = float(result.const.sel(effect='Penalty'))
print(f"costs: {costs} (expected -10), penalty: {penalty} (expected -20)")
assert costs == -10.0 and penalty == -20.0, "BUG: values misaligned!"
```

</details>

<details>
<summary>Benchmark script (<code>dev-scripts/benchmark_merge_alignment.py</code>)</summary>

```python
"""Benchmark merge performance: compare override (same coords) vs outer join paths.

Run from repo root:
    python dev-scripts/benchmark_merge_alignment.py
"""

import statistics
import timeit

import pandas as pd

import linopy

ROUNDS = 10
CALLS_PER_ROUND = 10

m = linopy.Model()

time = pd.RangeIndex(8760, name="time")  # hourly year
space = pd.Index([f"node_{i}" for i in range(100)], name="space")
tech = pd.Index([f"tech_{i}" for i in range(20)], name="tech")

gen = m.add_variables(coords=[time, space, tech], name="gen")
cap = m.add_variables(coords=[space, tech], name="cap")

# Warm up
_ = 1 * gen


def bench(label: str, func, rounds: int = ROUNDS, number: int = CALLS_PER_ROUND):
    times = [
[...]

PR #591, opened 2026-02-20
refac: Harmonize linopy operations and introduce a new predictable and strict convention

FBumann:
### Why: silent bugs in the current (legacy) arithmetic

linopy's current coordinate alignment has several classes of silent correctness bugs. None of them raise errors — they produce results that *look* reasonable but are *quietly wrong*.

**1. Positional alignment ignores labels** (#586, #550, #257)

When two operands have the same shape on a shared dimension, linopy matches them by *position*, ignoring coordinate labels entirely:

```python
x = m.add_variables(coords=[["costs", "penalty"]], name="x")
factors = xr.DataArray([2.0, 1.0], dims=["effect"],
                       coords={"effect": ["penalty", "costs"]})
x * factors
# Result: x["costs"] * 2.0, x["penalty"] * 1.0  — SWAPPED!
# Expected: x["costs"] * 1.0, x["penalty"] * 2.0
```

This affects all arithmetic operators and constraint creation. The optimization solves without error, but with the wrong constraints.

**2. Subset constants break associativity** (#572)

When a constant has different-sized coordinates, a left-join drops coordinates that might be needed by a later operation:

```python
# a has 3 time steps, b has 5, factor has 5
a + factor + b    # factor at time=3,4 → 40, 50  ✓
a + b + factor    # factor at time=3,4 → LOST     ✗
```

`(a + factor) + b ≠ a + (b + factor)` — the result depends on operand order.

**3. User NaN silently swallowed** (#620)

NaN in user-supplied data (often indicating missing or erroneous data) is silently filled with inconsistent neutral elements:

```python
data = xr.DataArray([1.0, np.nan, 3.0], ...)
x + data    # NaN → 0 (silent)
x * data    # NaN → 0 (kills the variable!)
x / data    # NaN → 1 (leaves variable unchanged!)
```

The fill values differ across operations with no consistent principle.

**4. Absent slots indistinguishable from zero** (#620)

`Variable.to_linexpr()` does not mark absent variable slots (from `.shift()`, `.where()`) as NaN, so multiplication and `fillna()` cannot distinguish "absent" from "zero":

```python
xs = x.shift(time=1)   # time=0 is absent
xs * 3                  # absent slot looks like 0 — not NaN
xs.fillna(42)           # no-op — nothing to fill
```

**5. Inconsistent Variable vs Expression paths** (#569, #571)

`x * subset_constant` and `(1 * x) * subset_constant` previously gave different results (one crashed, one didn't).

---

### What: the v1 convention

This PR introduces `linopy.options["arithmetic_convention"]` with two modes and a non-breaking transition:

- **`"legacy"` (default)** — reproduces current master behavior exactly. Emits `LinopyDeprecationWarning` on every legacy codepath.
- **`"v1"`** — strict coordinate matching, explicit NaN handling. Silent bugs become loud errors.

```python
linopy.options["arithmetic_convention"] = "v1"
```

#### Strict coordinate matching

Arithmetic operators (`+`, `-`, `*`, `/`) require matching coordinates on shared dimensions. Mismatched coordinates raise `ValueError` with suggestions:

```python
x[i=0,1,2] + y[i=1,2,3]  # ValueError: use .add(y, join="inner")
```

Named methods with explicit join — `.add()`, `.sub()`, `.mul()`, `.div()`, `.le()`, `.ge()`, `.eq()` accept `join=` parameter:

```python
x.add(y, join="inner")      # intersection
x.add(y, join="outer")      # union with fill
x.add(y, join="left")       # keep x's coordinates
x.add(y, join="override")   # positional alignment (opt-in only)
```

Named methods with fill_value — `.add()`, `.sub()`, `.mul()`, `.div()` accept `fill_value=` for explicit NaN filling before the operation (see [PR #620](https://github.com/PyPSA/linopy/pull/620)):

```python
expr.add(5, fill_value=0)      # fill NaN const with 0 before adding
expr.mul(factor, fill_value=1) # fill NaN const with 1 before multiplying
```

Free broadcasting — Constants can introduce new dimensions without restriction.

#### Algebraic laws

All standard algebraic laws hold under v1 **and** under legacy for same-coordinate operands:
- Commutativity: `a + b == b + a`, `a * c == c * a`
- Associativity: `(a + b) + c == a + (b + c)`
- Distributivity: `c * (a + b) == c*a + c*b`, `(a + b) / c == a/c + b/c`
- Identity, negation, zero

Legacy breaks associativity only when operands have mismatched coordinate ranges (the subset-constant case above). The v1 convention prevents this by requiring explicit coordinate handling.

---

### v1 NaN convention

**NaN means "absent term" — never a numeric value.**

NaN enters only from `mask=` at construction or structural operations (`.shift()`, `.where()`, `.reindex()`, `.reindex_like()`, `.unstack()`). Operations like `.roll()`, `.sel()`, `.isel()` do not produce NaN.

**Arithmetic with absent slots:**

| Operation | Absent slot | Why |
|---|---|---|
| `shifted + 5` | `+5` | Fills const with 0 (additive identity), required for associativity |
| `shifted - 5` | `-5` | Same — subtraction is addition of negation |
| `shifted * 3` | absent | NaN propagates — no correct implicit fill (0 kills, 1 preserves) |
| `shifted / 2` | absent | Same as multiplication |

When merging expressions (e.g., `x + y.shift(time=1)`), NaN marks individual terms, not entire coordinates. A coordinate is only fully absent when **all** terms are absent — `isnull()` checks this.

**User-supplied NaN raises `ValueError`** — users must handle NaN explicitly before arithmetic:

```python
x + data.fillna(0)           # NaN = "no offset"
x * factor.fillna(1)         # NaN = "no scaling"
expr.mul(3, fill_value=0)    # fill_value= shorthand on named methods
```

**`fillna()`** — `Variable.fillna(numeric)` returns `LinearExpression`, `Variable.fillna(Variable)` returns `Variable`, `Expression.fillna(value)` fills const at absent slots.

---

### Source changes

| File | Change |
|------|--------|
| `config.py` | `LinopyDeprecationWarning`, `arithmetic_convention` setting |
| `expressions.py` | All arithmetic paths branch on convention; `merge()` pre-validates user-dim coords under v1; `merge()` preserves NaN const when all inputs NaN; `_add_constant` fills const with additive identity; `to_constraint` has separate legacy/v1 paths; NaN validation at API boundaries; `fill_value=` on `.add()/.sub()/.mul()/.div()` |
| `common.py` | `align()` reads convention (legacy→inner, v1→exact) |
| `variables.py` | Scalar fast path in `__mul__`, explicit `TypeError` in `__div__`, `.reindex()` methods, `Variable.fillna(numeric)` returns `LinearExpression`, `to_linexpr()` sets `const=NaN` at absent slots in v1 |
| `piecewise.py` | `.fillna(0)` on breakpoint data for v1 compatibility |
| `monkey_patch_xarray.py` | DataArray/Dataset arithmetic with linopy types |
| `model.py` | Convention-aware model methods |

### Documentation

| Notebook | Content |
|----------|---------|
| `arithmetic-convention.ipynb` | Coordinate alignment rules, join parameter, migration guide |
| `missing-data.ipynb` | NaN convention principles, fillna patterns, masking with .sel() and mask=, legacy comparison |
| `_nan-edge-cases.ipynb` | Dev notebook: shift, roll, where, reindex, isnull, absent slot arithmetic, fillna/fill_value API, FILL_VALUE internals |

### Test structure

- **Marker-based separation**: `@pytest.mark.v1_only` and `@pytest.mark.legacy_only` for convention-specific tests
- **Shared fixtures** in `conftest.py` with auto-convention switching
- **`test_algebraic_properties.py`** (92 tests): formal specification of algebraic laws — commutativity, associativity, distributivity, identity, negation, zero, division/subtraction distributivity, multi-step constant folding, mixed-type commutativity, expression-expression laws. All pass under both conventions except 8 NaN-propagation tests (v1 only).
- **`test_legacy_violations.py`** (23 tests): catalog of concrete legacy bugs with paired legacy/v1 tests — positional alignment, subset associativity, user NaN handling, variable/expression inconsistency, absent-slot propagation. Each test traces to a specific issue number.
- Additional convention-specific tests in `test_linear_expression.py`, `test_constraints.py`, `test_convention.py`

### Rollout plan

1. **This PR**: Default `"legacy"` — nothing breaks
2. **Downstream**: Users opt in with `linopy.options["arithmetic_convention"] = "v1"`
3. **linopy v1**: Flip default to `"v1"` and drop legacy mode

### Open questions

- [ ] **`from_tuples` / `linexpr()`** — Currently follows the global convention. In practice always called with same-coord variables, so convention doesn't matter. Low-priority.
- [ ] **Pipe operator** — Only linopy objects, or also constants? (follow-up PR)

### Sub-PRs

- [x] #620 — Fix NaN propagation consistency and add fill_value API

## Test plan

- [x] All tests pass under both conventions
- [x] Legacy tests validate backward compatibility
- [x] v1 tests validate strict coordinate matching and NaN raises
- [x] Algebraic properties verified (92 tests): all laws hold under both conventions
- [x] Legacy violations documented (23 tests): concrete bugs with issue references
- [x] Piecewise and SOS tests run under both conventions
- [x] Documentation notebooks execute cleanly


coroa:
> The convention should be `"exact"` for all of `+`, `-`, `*`, `/`, with an additional check that neither side may introduce dimensions the other doesn't have — also for all operations.
> 

Let's clearly differentiate between dimensions and labels.

# labels

I agree with "exact" for labels by default, but we need an easy way to have inner or outer joining characteristics. I found the [pyoframe conventions](https://bravos-power.github.io/pyoframe/latest/learn/concepts/addition/#keep_extras-drop_extras
)
strange at the beginning, but they grew on me:

x + y.keep_extras()  to say that an outer join is in order and mismatches should fill with 0.

x + y.drop_extras() to say that you want an outer join.
x.drop_extras() + y does the same, though.

I have in a different project used | 0 to indicate keep_extras  ie (x + y | 0).


# dimensions

i am actually fond of the ability to auto broadcast over different dimensions. and would want to keep that (actually my main problem with pyoframe).

your first example actually implicitly assumes broadcasting.

FBumann:
# Quick scan of options to improve ergonomics using CLAUDE:

# Improving join ergonomics in linopy without mutable flags

The default convention should be `"exact"` for all arithmetic operations, with a pre-check that neither side introduces dimensions the other doesn’t have. For the cases where a non-exact join is intentional, here are three approaches that avoid mutable state on objects:

## 1. Named methods with `join=` (already in the PR)

Clean and explicit — intent is visible at the call site:

```python
x.add(y, join="outer", fill_value=0)
x.add(y, join="inner")
```

Best for one-off cases where a single operation needs a different join.

## 2. Context manager

Scoped override, no object mutation:

```python
with linopy.join("outer", fill_value=0):
    expr = x + y + z
```

Similar to `xr.set_options()` but limited to a block. Everything inside uses the specified join, everything outside falls back to `"exact"`. Clean for building a whole expression from misaligned pieces without wrapping every operation in a named method call.

## 3. `|` operator as fill shorthand

Borrowing from coroa’s `| 0` idea — tag a fill value onto an operand without mutating it:

```python
x + (y | 0)   # outer join with 0 fill for y's missing coords
```

The `|` returns a lightweight immutable wrapper carrying the fill intent, consumed and discarded immediately by `__add__`. Syntactically lightweight but potentially too implicit.

## Summary

|Approach                  |Ergonomics                |Explicitness|Complexity|
|--------------------------|--------------------------|------------|----------|
|Named method `.add(join=)`|Good for single ops       |High        |Low       |
|Context manager           |Good for expression blocks|Medium      |Medium    |
|`|` operator              |Most concise              |Low         |Low       |

The named method is the safest and already exists. The context manager is the most ergonomic for complex expressions. All three avoid the state management problems of mutable flags like `.keep_extras()` / `.drop_extras()`.

In all cases, `linopy.align(x, y, join=..., fill_value=...)` remains available for explicit pre-alignment before using the default `+` operator.

brynpickering:
@FBumann

> if you find pain points with having to align/fill data much more, we can maybeimprove the convention or add some shortcuts.

I don't think the complexity is strictly bad, it's just something that I can see users battling with.

One case I was analysing was where you minimise the size of Variable objs by removing some dim items but have input parameters with data shared across all dim items:

```py
m = Model()
tech_a, tech_b, tech_cost = ["wind", "solar"], ["gas"], ["wind", "solar", "gas"]
cap_a = m.add_variables(lower=0, coords=[tech_a], name="cap_a")
cap_b = m.add_variables(lower=0, coords=[tech_b], name="cap_b")
cap_c = m.add_variables(lower=0, coords=[tech_cost], name="cap_c")
cost_a = xr.DataArray([7, 9, float("nan")], coords=[("dim_0", tech_cost)])
cost_b = xr.DataArray([float("nan"), float("nan"), 11], coords=[("dim_0", tech_cost)])
cost_c = xr.DataArray([13, 17, 19], coords=[("dim_0", tech_cost)])

# could use dropna; could use outer or inner join when multiplying costs with caps
combined = (
    cap_a.mul(cost_a.fillna(0), join="left")
    .add(cap_b.mul(cost_b.fillna(0), join="left"), join="outer")
    .add(cap_c.mul(cost_c.fillna(0), join="left"), join="outer")
)
combined
```

Creates:

```
LinearExpression [dim_0: 3]:
----------------------------
[gas]: +11 cap_b[gas] + 19 cap_c[gas]
[solar]: +9 cap_a[solar] + 17 cap_c[solar]
[wind]: +7 cap_a[wind] + 13 cap_c[wind]
```

It works, it's just quite verbose since the default is `join="exact"` so you _must_ define all those joins.

I then extended it slightly by adding an extra parameter with reduced dimension size, and it becomes harder to get it to produce an expected result:

```py
m = Model()
tech_a, tech_b, tech_cost = ["wind", "solar"], ["gas"], ["wind", "solar", "gas"]
cap_a = m.add_variables(lower=0, coords=[tech_a], name="cap_a")
cap_b = m.add_variables(lower=0, coords=[tech_b], name="cap_b")
cap_c = m.add_variables(lower=0, coords=[tech_cost], name="cap_c")
cost_a = xr.DataArray([7, 9, float("nan")], coords=[("dim_0", tech_cost)])
cost_b = xr.DataArray([float("nan"), float("nan"), 11], coords=[("dim_0", tech_cost)])
cost_c = xr.DataArray([13, 17, 19], coords=[("dim_0", tech_cost)])
rate = xr.DataArray([1.04], coords=[("dim_0", ["gas"])])

# could use dropna; must use outer join when multiplying costs with caps
combined = (
    cap_a.mul(cost_a.fillna(0), join="outer")
    .add(cap_b.mul(cost_b.fillna(0), join="outer"), join="inner")
    .add(
        cap_c.mul(cost_c.fillna(0), join="outer")
        .mul(rate, join="outer", fill_value=1),
        join="outer"
    )
)
combined
```

This doesn't produce the expected result, but instead:

```
LinearExpression [dim_0: 3]:
----------------------------
[gas]: +11 cap_b[gas] + 19.76 cap_c[gas]
[solar]: +9 cap_a[solar] + 0 cap_c[solar]
[wind]: +7 cap_a[wind] + 0 cap_c[wind]
```

To get it to work, you have to first prepare the parameter:

```py
m = Model()
tech_a, tech_b, tech_cost = ["wind", "solar"], ["gas"], ["wind", "solar", "gas"]
cap_a = m.add_variables(lower=0, coords=[tech_a], name="cap_a")
cap_b = m.add_variables(lower=0, coords=[tech_b], name="cap_b")
cap_c = m.add_variables(lower=0, coords=[tech_cost], name="cap_c")
cost_a = xr.DataArray([7, 9, float("nan")], coords=[("dim_0", tech_cost)])
cost_b = xr.DataArray([float("nan"), float("nan"), 11], coords=[("dim_0", tech_cost)])
cost_c = xr.DataArray([13, 17, 19], coords=[("dim_0", tech_cost)])
rate = xr.DataArray([1.04], coords=[("dim_0", ["gas"])])
cost_c_rate = cost_c * rate.reindex_like(cost_c).fillna(1)

# could use dropna; must use outer join when multiplying costs with caps
combined = (
    cap_a.mul(cost_a.fillna(0), join="outer")
    .add(cap_b.mul(cost_b.fillna(0), join="outer"), join="inner")
    .add(cap_c.mul(cost_c_rate, join="outer"), join="outer")
)
combined
```

> What exactly would you like to achieve there?

Not combining a variable with another variable, but with a parameter, it's just a shortcut for `var.to_linexpr.mul(..., fill_value=...)` as `var.mul(..., fill_value=...)` since you already provide an equivalent shortcut for `join`.

> Explicitly pass fill_value=0 (or 1, etc.) to opt into filling

That seems appropriate. We don't want to be automatically filling with any default value where this could have an impact on the optimisation problem.

MaykThewessen:
Downstream PyPSA-Eur-style consumer perspective. We run an 8760h × NL+neighbors LP with several custom constraint modules that all sit in the v1-arithmetic blast radius described in the PR body. Sharing concrete patterns in case they're useful as regression cases.

**Patterns we'd test against v1:**

1. **NaN-mostly per-snapshot equality pins.** We pin BESS terminal SoC via `n.storage_units_t.state_of_charge_set` as a `DataFrame(NaN, columns=batteries)` with exactly one non-NaN entry per chunk. PyPSA forwards this to linopy as a constraint RHS. Under the legacy "NaN silently swallowed → 0" rule (case 3 in the PR), the wrong semantics would force SoC=0 every hour. Today this works because PyPSA pre-filters NaN before constraint creation; under v1 we'd want NaN-as-absent to be a first-class signal so the pre-filter could be dropped. A regression test: build `m.add_constraints(x == rhs)` with `rhs` a DataArray that is NaN-everywhere-except-one-snapshot, confirm only one constraint row is materialized.

2. **Per-snapshot time-varying `marginal_cost_t` with sparse non-NaN coverage.** We override foreign offshore wind `mc_t` to equal landing-bus LMP only on snapshots where the gen would otherwise dispatch (NaN elsewhere → static base cost). Same NaN-handling concern as (1), but on the objective side rather than the constraint side. Currently works; v1 needs to keep working.

3. **Label-indexed pandas Series multiplied by per-country/per-carrier DataFrames.** Our scarcity-rent and neighbor-markup-propagation modules construct factors like `pd.Series({'NL': 0.15, 'BE': 0.12, ...})` and multiply against per-snapshot generation expressions. This is exactly case 1 (positional alignment ignores labels) territory. Mixing pandas (label-aligned) and xarray (position-aligned) operands is where we feel least safe. Strict label alignment in v1 would be a clear win — we'd happily accept a `ValueError` on mismatch over silent swap.

4. **`phase_shift_extendable` PST custom constraint** (PyPSA fork PR #1661, currently in our pin). Writes Transformer phase-shift variables and flow constraints with mixed label/positional coords. Would benefit from a clear v0→v1 migration story — we'd rather port once, with strict errors guiding us, than chase silent dispatch shifts later.

5. **Sweep-orchestrator multi-process reproducibility.** Adjacent to #710's `_broadcast_points` set-iteration fix: we run many variants in parallel processes. Any hash-randomized dim ordering produces apparent run-to-run drift that masks real model deltas. Strict v1 ordering is a correctness win even beyond the silent-bugs angle.

**Migration concern:** the breakage surface is large for downstream PyPSA-Eur derivatives. A deprecation cycle that emits a `LinopyAlignmentWarning` on any operation whose v0 vs v1 result *differs* — rather than blanket on all alignment — would let us find and port only the affected call sites. Otherwise the warning noise floor will be too high to act on.

**Offer:** happy to run our 8760h LP against a v1-enabled branch and report any silently shifted constraints (compare LP file hash + objective). Let us know which SHA you want exercised.

FBumann:
@MaykThewessen Thanks for the perspective. We are working on this currently, but it will take a few weeks probably until we have it ready for you to run.

Would you be open for discussion about the nan stuff?

MaykThewessen:
@FBumann yes, happy to. NaN semantics are the part that touches our code most
directly — constraint-RHS side (SoC pin via `state_of_charge_set`) and objective
side (`marginal_cost_t` and `p_max_pu` with sparse non-NaN coverage). Async on a
dedicated issue is probably most efficient; sync call also fine if it helps
unblock design.

Concrete starting point: we ran a controlled test of #627's draft against our
SoC-pin pattern this morning.

- Under `arithmetic_convention="legacy"` (default), behaviour is identical to
  current master — SoC pin holds (`SoC[-1]=100.0` in a 24h LP).
- Under `arithmetic_convention="v1"`, the same pattern raises
  `ValueError: Constraint RHS contains NaN values` at `expressions.py:1325`,
  before PyPSA's existing NaN-RHS auto-mask runs. Loud, not silent — which is
  great. But it means every PyPSA-Eur downstream that relies on the auto-mask
  (storage with `state_of_charge_set`, sparse `marginal_cost_t`, gapped
  `p_max_pu`) will fail at constraint-build time the moment they flip the
  option.

The design question is where the pre-mask should live. Two options:

1. **Keep linopy strict, push mask to consumers.** Every downstream
   pre-filters NaN before `add_constraints`. One rule, but high migration cost
   and duplicated masking logic across PyPSA, PyPSA-Eur, atlite-derived
   builders, custom forks.
2. **`add_constraints` accepts NaN-as-absent natively, per-call.** E.g.
   `add_constraints(lhs == rhs, on_nan="raise" | "skip_row" | "propagate")`
   with `raise` as the v1 default. One implementation, fused with the existing
   xarray alignment, no global mode, explicit at each call site.

I lean toward (2) because the NaN in question is created by linopy's own
`xr.align(join="outer")`, not by user intent — the consumer can't distinguish
"user meant skip" from "alignment fill", only linopy sees the align op. So
handling it in the alignment layer feels structurally correct. But I see the
argument for keeping linopy policy-free.

Happy to open a discussion issue with the repro + sketch of both options.
File on linopy, or prefer to scope on your side first?

FBumann:
> @FBumann yes, happy to. NaN semantics are the part that touches our code most directly — constraint-RHS side (SoC pin via `state_of_charge_set`) and objective side (`marginal_cost_t` and `p_max_pu` with sparse non-NaN coverage). Async on a dedicated issue is probably most efficient; sync call also fine if it helps unblock design.
> 
> Concrete starting point: we ran a controlled test of #627's draft against our SoC-pin pattern this morning.
> 
> * Under `arithmetic_convention="legacy"` (default), behaviour is identical to
>   current master — SoC pin holds (`SoC[-1]=100.0` in a 24h LP).
> * Under `arithmetic_convention="v1"`, the same pattern raises
>   `ValueError: Constraint RHS contains NaN values` at `expressions.py:1325`,
>   before PyPSA's existing NaN-RHS auto-mask runs. Loud, not silent — which is
>   great. But it means every PyPSA-Eur downstream that relies on the auto-mask
>   (storage with `state_of_charge_set`, sparse `marginal_cost_t`, gapped
>   `p_max_pu`) will fail at constraint-build time the moment they flip the
>   option.
> 
> The design question is where the pre-mask should live. Two options:
> 
> 1. **Keep linopy strict, push mask to consumers.** Every downstream
>    pre-filters NaN before `add_constraints`. One rule, but high migration cost
>    and duplicated masking logic across PyPSA, PyPSA-Eur, atlite-derived
>    builders, custom forks.
> 2. **`add_constraints` accepts NaN-as-absent natively, per-call.** E.g.
>    `add_constraints(lhs == rhs, on_nan="raise" | "skip_row" | "propagate")`
>    with `raise` as the v1 default. One implementation, fused with the existing
>    xarray alignment, no global mode, explicit at each call site.
> 
> I lean toward (2) because the NaN in question is created by linopy's own `xr.align(join="outer")`, not by user intent — the consumer can't distinguish "user meant skip" from "alignment fill", only linopy sees the align op. So handling it in the alignment layer feels structurally correct. But I see the argument for keeping linopy policy-free.
> 
> Happy to open a discussion issue with the repro + sketch of both options. File on linopy, or prefer to scope on your side first?

Feel free to open one. I like your take on nan. Im leaning on defaulting to rais/strict, but allowing it via opt in seems reasonable. Im not sure if it should be in add_constraints though, because a user using pypsa cant reach that, and it would not work for expressions, as they are created bedore add_constraints() does its job.

FBumann:
Superseeded by #717

PR #596, opened 2026-02-26
feat(xpress): add direct io

CharlieFModo:
Closes # (if applicable).

- Adds `xpress` direct solver IO
- To ensure `xpress` direct IO is faster than `lp-polars` this PR removes `pandas` from the hot-path of matrix operations required to get problem coefficients to write to `xpress`
  - This seems to have large performance gains throughout the package beyond what I have tested for this PR
  - I would like some guidance on whether these changes are ok in the broader context of the package. I've only tested them on the problems that I am interested in solving and found this gives a c. 10x improvement from the lp-polars IO

## Checklist


coroa:
We probably will need to restructure the package a bit as more and more alternative io paths are crowding into `io.py`. Maybe some of it should go into solvers or we start an io directory with specific sub modules.

FabianHofmann:
@CharlieFModo thank you for your PR, this is great stuff and triggered a good initiative on our side #630. Given that #630 leads to even further performance improvements for the matrix creation, I would like to only include the `to_xpress` function from this pr. So, could you either create a new PR with only the export function and the relevant tests or let me know if I should do it? thanks!

CharlieFModo:
@FabianHofmann - no worries, I'm glad it put you towards a better solution. I'll close this and open a smaller PR with just the direct io when I get some spare time :)

PR #622, opened 2026-03-16
Add Solver.resolve() and Model.apply_result() for iterative re-solve workflows

### LINKED ISSUE #198: Support iterative solve of the same solver_model (when using highs)

FabianHofmann:
closes #198

This PR adds support for iterative re-solve workflows where users solve a model once, modify the native solver object directly (e.g. change objective coefficients or bounds), and re-solve without rebuilding the linopy model.

**New APIs:**

- `Solver.resolve(solver_model, sense)` — re-solves an existing native solver model and returns a `Result`. Implemented for HiGHS, Gurobi, and Mosek; raises `NotImplementedError` for other solvers.
- `Model.apply_result(result, solver_name)` — resets the current solution and maps a solver `Result` back onto the model's variables (primal) and constraints (dual).

**Refactoring:**

- The solution-mapping logic previously inlined in `Model.solve()` has been extracted into `apply_result()`, which `solve()` now calls internally. No behavioral change for existing users.

**Documentation:**

- New notebook `examples/iterative-resolving.ipynb` demonstrating the full workflow: initial solve, native model modification, re-solve loop, parameter sweep, and infeasible result handling.
- Notebook linked into the Sphinx docs under the User Guide section.

**Tests:**

- `test/test_resolve.py` covering `apply_result` (optimal, infeasible, `None` solution), `Highs.resolve()` end-to-end, and `NotImplementedError` for unsupported solvers.

## Checklist

### DIFF
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -116,6 +116,7 @@ This package is published under MIT license.
    piecewise-linear-constraints
    piecewise-linear-constraints-tutorial
    manipulating-models
+   iterative-resolving
    testing-framework
    transport-tutorial
    infeasible-model
--- a/doc/iterative-resolving.nblink
+++ b/doc/iterative-resolving.nblink
@@ -1,0 +1,3 @@
+{
+    "path": "../examples/iterative-resolving.ipynb"
+}
--- a/doc/release_notes.rst
+++ b/doc/release_notes.rst
@@ -4,6 +4,8 @@ Release Notes
 Upcoming Version
 ----------------
 
+* Add ``Solver.resolve()`` method for re-solving a modified native solver model without rebuilding the linopy model. Implemented for HiGHS, Gurobi, and Mosek.
+* Add ``Model.apply_result()`` to map a solver ``Result`` back onto model variables and constraints, enabling iterative solve workflows.
 * Harmonize coordinate alignment for operations with subset/superset objects:
   - Multiplication and division fill missing coords with 0 (variable doesn't participate)
   - Addition and subtraction of constants fill missing coords with 0 (identity element) and pin result to LHS coords
--- a/examples/iterative-resolving.ipynb
+++ b/examples/iterative-resolving.ipynb
@@ -1,0 +1,302 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "d21785bf",
+   "metadata": {},
+   "source": [
+    "# Iterative Re-solving\n",
+    "\n",
+    "In many optimization workflows you need to solve a model, tweak parameters on the native solver object, and re-solve — without rebuilding the entire linopy model. This is common in:\n",
+    "\n",
+    "- Sensitivity analysis (varying objective coefficients or bounds)\n",
+    "- Decomposition algorithms (Benders, column generation)\n",
+    "- Rolling-horizon / receding-horizon schemes\n",
+    "\n",
+    "Linopy supports this via two methods:\n",
+    "\n",
+    "- **`Solver.resolve()`** — re-solves an existing native solver model and returns a `Result`\n",
+    "- **`Model.apply_result()`** — maps a `Result` back onto the linopy model's variables and constraints\n",
+    "\n",
+    "Currently `resolve()` is implemented for the **HiGHS**, **Gurobi**, and **Mosek** solvers."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "f7f604f5",
+   "metadata": {},
+   "source": [
+    "## Setup\n",
+    "\n",
+    "We start with a simple two-variable LP."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "122ee88e",
+   "metadata": {
+    "execution": {
+     "iopub.execute_input": "2026-03-16T08:48:03.207441Z",
+     "iopub.status.busy": "2026-03-16T08:48:03.207185Z",
+     "iopub.status.idle": "2026-03-16T08:48:03.649145Z",
+     "shell.execute_reply": "2026-03-16T08:48:03.648818Z"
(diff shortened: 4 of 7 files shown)


FBumann (commented):
Overall: Nice Change. THe extraction of the apply_result method itself is a nice addition regarding code quality.

coroa (commented):
Read the PR now fully, with the notebook it makes sense. (Sorry, i find reading notebooks normally quite annoying (diff is soo noisy)).

I like the `apply_result` addition, definite keeper.

But i would aim for a stronger refactor, and re-think the Solver objects, they should own the solver_model as well as the model parameter and io_api and also incorporate capabilities and updates.

The resolve function looks like a band-aid to fix a leaking abstraction.

MaykThewessen:
@coroa Thinking more about this — the model-diffing direction you mentioned seems like the bigger opportunity here.

Right now our blueprint reuse system (~100 lines in our solver module) manually tracks which HiGHS rows are time-varying, computes new bounds, and pushes them via the C API. It works, but it's tightly coupled to HiGHS internals and our specific LP structure.

If linopy could diff the model state automatically and push only incremental updates to the native solver, that entire layer could be replaced by standard linopy API calls. The workflow would become:

1. `model.solve()` (chunk 1 — full build + solve)
2. Modify linopy model (update constraint RHS, objective coefficients for chunk 2)
3. `model.resolve()` or equivalent — linopy diffs internally, pushes changes to the cached native solver, re-solves

No user code touching `changeRowsBounds()` or `changeColsCost()`. No manual solution mapping. The solver-specific update dispatch lives in linopy, not in every downstream project.

Happy to share our specific modification patterns (which ~15% of rows change and why) as a concrete test case if useful for the diffing design.

FabianHofmann:
closing in favor of #682 and #699

PR #630, opened 2026-03-23
perf: matrix accessor rewrite

coroa:

Ok, i went ahead and tried to optimize the direct solver path which looked very promising from #596 by @CharlieFModo . Most solvers accept scipy sparse matrices in CSR format, those CSR matrices are easy to stack and correspond quite closely to how constraints are stored within xarray currently.

### New constraint class hierarchy

A new abstract base class `ConstraintBase` provides all
shared read-only properties and methods. A new frozen `CSRConstraint` class
stores constraint data directly as a scipy CSR sparse matrix plus flat numpy
arrays for RHS and duals — avoiding the xarray Dataset overhead
constraints.

```
ConstraintBase (ABC)
├── Constraint   — xarray Dataset-backed, mutable
└── CSRConstraint          — CSR-backed, immutable ("frozen")
```

One can losslessly convert between Constraint and CSRConstraint, by calling mutable() or freeze().

`add_constraints(..., freeze=True)` converts the `Constraint` to a `CSRConstraint` before assigning it to the model. I measured memory savings of up to 90% for constraints with many terms.

EDIT: 

The feature is made opt-in and the globally freezing constraint can be configured with 
```python 
Model(freeze_constraints = True)
```
additional gains from not setting the name (even though I made it more performant)
```python
Model(set_names_in_solver_io=False)
```

`.to_matrix()` on both of these now always returns a `csr matrix` of `shape(n_active_constraints, n_active_variables)` and `con_labels` a numpy array of size `n_active_constraints` with the label of the constraint in it.

So that the number of columns of these non-missing matrices can match with n_active_variables there is a new `VariableLabelIndex` on `m.vars` which holds two cached properties:
- `vlabels`: active variable labels in encounter order, shape (n_active_vars,)
- `label_to_pos`: derived from vlabels; size _xCounter, maps label -> position (-1 if masked)

Chunking is not immediately useful anymore, because each constraint will be turned quite quickly into a numpy array before being reduced to the csr matrix. We probably could implement an optimized to_matrix path which uses the chunking.

## Open ToDos

- [x] Benchmark against current main branch
- [x] Document API changes: need for .mutable(), .freeze(), DataArray API only on Constraint
- [x] LP writer works by reconstructing mutable constraints out of the frozen ones, should be replaced by a new implementation.
- [x] Tests of direct apis other than highs
- [ ] ~Think about Chunking~
- [x] mypy is quite unhappy with my changes :(

## Quick benchmarking with pypsa.examples.carbon_management

Slightly cheated, because i commented out the names setting in `to_highspy`, which is unnecessary for solving and retrieving the solution, because we can just work with the indices. This takes ~2s. But i removed it from both branches (because it affected them equally).

### On main branch

```python
In [1]: import pypsa, linopy as lp
   ...: n = pypsa.examples.carbon_management()
In [2]: %time m = n.optimize.create_model()
[...]
CPU times: user 4.49 s, sys: 483 ms, total: 4.97 s
Wall time: 4.97 s
In [5]: %time lp.io.to_highspy(m)
Running HiGHS 1.12.0 (git hash: 755a8e0): Copyright (c) 2025 HiGHS under MIT licence terms
CPU times: user 10.9 s, sys: 1.78 s, total: 12.7 s
Wall time: 12.7 s
Out[5]: <highspy.highs.Highs at 0x70228531d010>
```

### On `perf/matrix-acccesor-rewrite` (this PR)
```python
In [1]: import pypsa, linopy as lp
   ...: n = pypsa.examples.carbon_management()
In [2]: %time m = n.optimize.create_model()
CPU times: user 5.33 s, sys: 519 ms, total: 5.84 s
Wall time: 5.86 s
In [5]: %time lp.io.to_highspy(m)
Running HiGHS 1.12.0 (git hash: 755a8e0): Copyright (c) 2025 HiGHS under MIT licence terms
CPU times: user 446 ms, sys: 266 ms, total: 712 ms
Wall time: 713 ms
Out[5]: <highspy.highs.Highs at 0x767c48a17ad0>
```

**EDIT (by @FabianHofmann):** Changed back naming to keep `Constraint` class untouched and add `CSRConstriaint`

## Checklist


FBumann (commented):
I commented on some minor things. Im not using the direct interface and wont be able to add that much value probably.
I verified the speed claims on the direct io paths.
I also verified that the **existing LP writer is a lot slower**.

linopy/model.py
```diff
@@ -214,7 +217,10 @@ def __init__(
             gettempdir() if solver_dir is None else solver_dir
         )
 
-        self.matrices: MatrixAccessor = MatrixAccessor(self)
+    @property
```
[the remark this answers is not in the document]
coroa: yes, that was one of the key goals.

FabianHofmann:
this is great, one big question that came up when looking at the code is: why renaming `Constraint` to `MutableConstraint` and making `Constraint` the frozen one, and not instead keep `Constraint` and add a `FrozenConstraint`? From a user perspective the current renaming is breaking and the exposed object has a non-intuitive name.

linopy/common.py
```diff
@@ -939,6 +939,49 @@ def find_single(value: int) -> tuple[str, dict] | tuple[None, None]:
...
+        for _, var in self._variables.items():
+            labels = var.labels.values.ravel()
+            mask = labels != -1
+            label_lists.append(labels[mask])
+        return (
+            np.concatenate(label_lists) if label_lists else np.array([], dtype=np.intp)
+        )
+
+    @cached_property
+    def label_to_pos(self) -> np.ndarray:
```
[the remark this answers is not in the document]
coroa: Added one

linopy/common.py
```diff
@@ -939,6 +939,49 @@ def find_single(value: int) -> tuple[str, dict] | tuple[None, None]:
...
+    - ``label_to_pos``: derived from vlabels; size _xCounter, maps label -> position (-1 if masked)
+
+    Invalidated by clearing the instance ``__dict__`` when variables are added or removed.
+    """
+
+    def __init__(self, variables: Any) -> None:
+        self._variables = variables
+
+    @cached_property
+    def vlabels(self) -> np.ndarray:
```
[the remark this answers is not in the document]
coroa: Added one

FBumann:
@FabianHofmann I didnt realize that freezing was opt in. I just ran integration tests again with freezing on Model, and my testing fails due to linopy.testing.assert_con_equal raising.

## Bug report

`assert_linequal` compares `coeffs` and `vars` positionally along `_term`. With `freeze_constraints=True`, the CSR round-trip sorts terms by variable label, so stored constraints have different term order than freshly built expressions. This causes false failures on mathematically equivalent constraints.

```python
from linopy import Model
import pandas as pd

m = Model(freeze_constraints=True)
coords = pd.RangeIndex(3, name="time")
x = m.add_variables(coords=[coords], name="x")
y = m.add_variables(coords=[coords], name="y")

m.add_constraints(y - x == 0, name="con")

desired = y == x
assert_linequal(m.constraints["con"].lhs, desired.lhs)  # fails
```

Sorting both sides by `vars` along `_term` before comparing would fix it probably. I think this is a quick fix in testing.py, but maybe this is unintended behaviour after all?

FabianHofmann:
> @FabianHofmann I didnt realize that freezing was opt in. I just ran integration tests again with freezing on Model, and my testing fails due to linopy.testing.assert_con_equal raising.
> 
> ## Bug report
> `assert_linequal` compares `coeffs` and `vars` positionally along `_term`. With `freeze_constraints=True`, the CSR round-trip sorts terms by variable label, so stored constraints have different term order than freshly built expressions. This causes false failures on mathematically equivalent constraints.
> 
> ```python
> from linopy import Model
> from linopy.testing import assert_linequal
> import pandas as pd
> 
> m = Model(freeze_constraints=True)
> coords = pd.RangeIndex(3, name="time")
> x = m.add_variables(coords=[coords], name="x")
> y = m.add_variables(coords=[coords], name="y")
> 
> m.add_constraints(y - x == 0, name="con")
> 
> desired = y == x
> assert_linequal(m.constraints["con"].lhs, desired.lhs)  # fails
> ```
> 
> Sorting both sides by `vars` along `_term` before comparing would fix it probably. I think this is a quick fix in testing.py, but maybe this is unintended behaviour after all?

@FBumann (I am back from a small break I had to take to restore my computer) good catch. that is a good question. I would argue we want to test the semantic equality of the linear expressions so sorting before would be fair. could you spot where exactly the resorting takes place?

FBumann:
> > @FabianHofmann I didnt realize that freezing was opt in. I just ran integration tests again with freezing on Model, and my testing fails due to linopy.testing.assert_con_equal raising.
> > ## Bug report
> > `assert_linequal` compares `coeffs` and `vars` positionally along `_term`. With `freeze_constraints=True`, the CSR round-trip sorts terms by variable label, so stored constraints have different term order than freshly built expressions. This causes false failures on mathematically equivalent constraints.
> > ```python
> > from linopy import Model
> > from linopy.testing import assert_linequal
> > import pandas as pd
> > 
> > m = Model(freeze_constraints=True)
> > coords = pd.RangeIndex(3, name="time")
> > x = m.add_variables(coords=[coords], name="x")
> > y = m.add_variables(coords=[coords], name="y")
> > 
> > m.add_constraints(y - x == 0, name="con")
> > 
> > desired = y == x
> > assert_linequal(m.constraints["con"].lhs, desired.lhs)  # fails
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Sorting both sides by `vars` along `_term` before comparing would fix it probably. I think this is a quick fix in testing.py, but maybe this is unintended behaviour after all?
> 
> @FBumann (I am back from a small break I had to take to restore my computer) good catch. that is a good question. I would argue we want to test the semantic equality of the linear expressions so sorting before would be fair. could you spot where exactly the resorting takes place?

constraints.py:478 — csr.sum_duplicates() sorts CSR column indices by variable label within each row. When CSRConstraint._to_dataset() (line 730-731) reconstructs the constraint, terms come out in variable-label order instead of insertion order.

`csr.sum_duplicates()` does the sorting using `scipy`. We cant change that afaik, and term order doesnt matter for math, so we need to adjust the testing instead.

linopy/constraints.py
```diff
@@ -319,87 +393,737 @@ def __repr__(self) -> str:
...
+    def _refreeze_after(self, mutate: Callable[[Constraint], None]) -> None:
+        mc = self.mutable()
+        mutate(mc)
+        refrozen = CSRConstraint.from_mutable(mc, self._cindex)
+        self._csr = refrozen._csr
+        self._con_labels = refrozen._con_labels
+        self._rhs = refrozen._rhs
+        self._sign = refrozen._sign
+        self._coords = refrozen._coords
+        self._dual = None
```
coroa: I really don't like this hack

linopy/constraints.py
```diff
@@ -319,87 +393,737 @@ def __repr__(self) -> str:
...
+            raise AttributeError(
+                "Underlying is optimized but does not have dual values stored."
+            )
+        return self._active_to_dataarray(self._dual, fill=np.nan)
+
+    @dual.setter
+    def dual(self, value: DataArray) -> None:
+        """Set dual values from a DataArray aligned with the full coord shape."""
+        vals = np.asarray(value).ravel()
+        self._dual = vals[self.active_positions]
```
coroa, line 673: Why do we want to store _dual in full coord shape? Condensed active constraints should be fine!

linopy/constraints.py
```diff
@@ -319,87 +393,737 @@ def __repr__(self) -> str:
...
+        slice_dims: list | None = None,
+    ) -> Generator[CSRConstraint, None, None]:
+        """Yield row-batched sub-Constraints without Dataset reconstruction."""
+        nnz = self._csr.nnz
+        if slice_size is None or nnz <= slice_size:
+            yield self
+            return
+
+        n = self._csr.shape[0]
+        cumulative = np.cumsum(np.diff(self._csr.indptr))
```
coroa: np.cumsum \circ np.diff is a no-op? isn't it?

linopy/constraints.py
```diff
@@ -319,87 +393,737 @@ def __repr__(self) -> str:
...
+        batch_start = 0
+        for batch_end_nnz in range(slice_size, nnz + slice_size, slice_size):
+            batch_end = int(np.searchsorted(cumulative, batch_end_nnz, side="right"))
+            batch_end = max(batch_end, batch_start + 1)
```
coroa: That's strangely convoluted as an algorithm.

```python
def equal_nnz_slices(indptr, slice_size):
    n_rows = len(indptr) - 1
    start = 0
    while start < n_rows:
        offset = np.searchsorted(indptr[start + 1:], indptr[start] + slice_size, side='left')
        end = min(start + 1 + offset, n_rows)
        yield slice(start, end)
        start = end
```

and then

```python
for rows in equal_nnz_slices(self._csr.indptr, slice_size):
    ...
    yield CSRConstraint(
         csr=self._csr[rows],
         ...
    )
```

linopy/constraints.py
```diff
@@ -97,199 +104,266 @@ def _conwrap(con: Constraint, *args: Any, **kwargs: Any) -> Constraint:
...
+
+    coeffs_final = con.coeffs.values.ravel().reshape(vars_2d.shape)[row_mask]
+    cols = label_to_pos[vars_final[valid_final]]
+    data = coeffs_final[valid_final]
+
+    counts = valid_final.sum(axis=1)
+    indptr = np.empty(len(con_labels) + 1, dtype=np.int32)
+    indptr[0] = 0
+    np.cumsum(counts, out=indptr[1:])
+    return con_labels, row_mask, cols, data, indptr
```
coroa: This should become a method on `Constraint`

linopy/constraints.py
```diff
@@ -97,199 +104,266 @@ def _conwrap(con: Constraint, *args: Any, **kwargs: Any) -> Constraint:
...
+        Get the number of active constraints (non-masked, with at least one valid variable).
         """
-        return self._data
+        labels = self.labels.values
+        vars_arr = self.vars.values
+        if labels.ndim == 0:
+            return int(labels != FILL_VALUE["labels"] and (vars_arr != -1).any())
+        return int(
+            ((labels != FILL_VALUE["labels"]) & (vars_arr != -1).any(axis=-1)).sum()
+        )
```
coroa: Let's make this abstract here and put in on constraint?

linopy/constraints.py
```diff
@@ -97,199 +104,266 @@ def _conwrap(con: Constraint, *args: Any, **kwargs: Any) -> Constraint:
...
+            return result.astype(bool)
+        return None
 
     @property
-    def is_assigned(self) -> bool:
-        return self._assigned
+    def lhs(self) -> expressions.LinearExpression:
+        """Get the left-hand-side linear expression of the constraint."""
+        data = self.data[["coeffs", "vars"]].rename({self.term_dim: TERM_DIM})
+        return expressions.LinearExpression(data, self.model)
```
coroa: I'd move this onto Constraint

linopy/constraints.py
```diff
@@ -319,87 +393,737 @@ def __repr__(self) -> str:
...
+
+        df = to_dataframe(ds, mask_func=mask_func)
+
+        # Group repeated variables in the same constraint
+        agg_custom = {k: "first" for k in list(df.columns)}
+        agg_standards = dict(coeffs="sum", rhs="first", sign="first")
+        agg = {**agg_custom, **agg_standards}
+        df = df.groupby(["labels", "vars"], as_index=False).aggregate(agg)
+        check_has_nulls(df, name=f"{self.type} {self.name}")
+        return df
```
coroa, line 424: Do we still need `flat` anywhere?

linopy/constraints.py
```diff
@@ -319,87 +393,737 @@ def __repr__(self) -> str:
...
+            Shape (n_active_cons, n_active_vars).
+        con_labels : np.ndarray
+            Active constraint labels in row order.
+        """
+        con_labels, _, cols, data, indptr = _matrix_export_data(self, label_index)
+        csr = scipy.sparse.csr_array(
+            (data, cols, indptr), shape=(len(con_labels), label_index.n_active_vars)
+        )
+        csr.sum_duplicates()
+        return csr, con_labels
```
coroa: I'd make this abstract here and move the implementation onto Constraint

linopy/constraints.py
coroa: Would at least deprecate

linopy/constraints.py
```diff
@@ -319,87 +393,737 @@ def __repr__(self) -> str:
...
+            "vars": vlabels[csr.indices],
+            "rhs": self._rhs[rows],
+        }
+        if isinstance(self._sign, str):
+            data["sign"] = pl.Series(
+                "sign", [self._sign], dtype=sign_dtype
+            ).new_from_index(0, len(rows))
+        else:
+            data["sign"] = pl.Series("sign", self._sign[rows], dtype=sign_dtype)
+        return pl.DataFrame(data)[["labels", "coeffs", "vars", "sign", "rhs"]]
```
coroa: I think there are some speed-ups still hidden in the to_polars path since the polars format does not distinguish between the different sizes of labels, sign, rhs (which are of size ncons), while coeffs and vars and of size nnz. Putting them into a single polars frame by spreading them out feels like a bad idea.

coroa: But i would not address this in this PR, since that would changes along the whole path and this is supposedly a dying path anyway.

linopy/constraints.py
```diff
@@ -319,87 +393,737 @@ def __repr__(self) -> str:
...
+                self._sign
+                if isinstance(self._sign, str)
+                else self._sign[batch_start:batch_end]
+            )
+            yield CSRConstraint(
+                csr=self._csr[batch_start:batch_end],
+                con_labels=self._con_labels[batch_start:batch_end],
+                rhs=self._rhs[batch_start:batch_end],
+                sign=sign,
+                coords=self._coords,
```
coroa: Don't _coords also need to fit the batch slices?

coroa (commented):
Ok, went over it. Quite some stuff still to clean up

linopy/constraints.py
coroa: Couldn't `data["sign"] = pl.lit(self.sign)` for the str case?

linopy/constraints.py
coroa: Yes, but there are no coords that would be correct.

linopy/constraints.py
coroa: Ah, i was wrong. This IS condensing it onto _dual.

FabianHofmann:
@coroa thanks for the follow up. I just ran the tests in pypsa to confirm the implementation works, all lopf tests pass. I would say let's merge this do not longer create merge conflicts (I would like to get the ball rolling for solver refac). we can always follow up further

linopy/constraints.py
FabianHofmann: done

FBumann:
Amazing that you are progressing 👍

coroa:
Amazing i did not notice, that you did merge it :)

PR #638, opened 2026-04-01
feat: Unify piecewise API behind add_piecewise_formulation (sign + LP dispatch)

FBumann:
## Summary

Follows up on #602. Replaces the descriptor pattern (`PiecewiseExpression`, `PiecewiseConstraintDescriptor`, `piecewise()`) with a single, stateless construction layer: `add_piecewise_formulation`.

`add_piecewise_formulation` becomes the one entry point for piecewise work — equality, one-sided inequality, N-variable linking, per-entity breakpoints, unit-commitment gating, and automatic method dispatch all live behind one call. `tangent_lines()` survives as a low-level helper for users who want to build chord expressions manually.

## The API

Each `(expression, breakpoints[, sign])` tuple links a variable to its breakpoints. All tuples share interpolation weights, coupling them on the same curve piece. The optional 3rd element marks one tuple as **bounded** by the curve instead of pinned to it.

### Equality (default)

```python
# 2-variable: fuel = f(power)
m.add_piecewise_formulation(
    (power, [0, 30, 60, 100]),
    (fuel,  [0, 36, 84, 170]),
)

# N-variable (CHP plant — same pattern, more tuples)
m.add_piecewise_formulation(
    (power, [0, 30, 60, 100]),
    (fuel,  [0, 40, 85, 160]),
    (heat,  [0, 25, 55, 95]),
)

# Disjunctive (disconnected operating regions)
m.add_piecewise_formulation(
    (power, linopy.segments([(0, 0), (50, 80)])),
    (cost,  linopy.segments([(0, 0), (125, 200)])),
)

# Per-entity breakpoints (different curves per generator)
m.add_piecewise_formulation(
    (power, linopy.breakpoints({"gas": [0, 30, 60, 100], "coal": [0, 50, 100, 150]}, dim="gen")),
    (fuel,  linopy.breakpoints({"gas": [0, 40, 90, 180], "coal": [0, 55, 130, 225]}, dim="gen")),
)

# Breakpoints from slopes
m.add_piecewise_formulation(
    (power, [0, 30, 60, 100]),
    (fuel,  linopy.breakpoints(slopes=[1.2, 1.4, 1.7], x_points=[0, 30, 60, 100], y0=0)),
)
```

### Inequality bounds — per-tuple sign

Append `"<="` or `">="` as a third tuple element to mark one expression as bounded by the curve. The other tuples stay pinned. The bounded tuple need not be first — its role is visible at the call site.

```python
# fuel ≤ f(power).  "auto" picks the cheapest correct formulation — a pure
# LP chord formulation when the curvature matches the sign (concave + "<=",
# or convex + ">="), SOS2/incremental otherwise.
m.add_piecewise_formulation(
    (fuel,  [0, 20, 30, 35], "<="),  # bounded
    (power, [0, 10, 20, 30]),         # pinned
)
```

On a matching-curvature curve this dispatches to `method="lp"`: one chord inequality per piece plus a domain bound `x_min ≤ x ≤ x_max`, and **no auxiliary variables**. Mismatched curvature (convex + `"<="` / concave + `">="`) describes the *wrong* region (not just a looser one) — `method="auto"` detects this and falls back to SOS2/incremental with an explanatory info log; `method="lp"` raises with a clear error.

**Restrictions (current):**
- At most one tuple may carry a non-equality sign.
- With 3+ tuples, all signs must be `"=="`.

The N-input bounded case (e.g. `z >= f(x, y)`) is rejected with an error that invites a GitHub issue if anyone has a concrete use case — we don't promise a specific future API shape.

valid or rejectable

### Unit-commitment gating — `active`

```python
# active=1: power in [30, 100], fuel = f(power)
# active=0: power = 0, fuel = 0
m.add_piecewise_formulation(
    (power, [30, 60, 100]),
    (fuel,  [40, 90, 170]),
    active=commit,  # binary variable
)
```

Works with SOS2, incremental, and disjunctive. With a bounded tuple, deactivation only pushes the signed bound to `0` — set `lower=0` on naturally non-negative outputs (fuel, cost, heat) and the combination forces `y = 0` automatically.

### Low-level: `tangent_lines`

```python
# Returns a LinearExpression with one chord per piece — no variables.
# Most users should prefer add_piecewise_formulation with a bounded tuple,
# which builds on this helper and adds domain bounds + curvature checks.
t = linopy.tangent_lines(power, x_pts, y_pts)
m.add_constraints(fuel <= t)
```

### New public API
- `add_piecewise_formulation` — one entry point for all piecewise constructions. Returns a `PiecewiseFormulation` carrying `.method` (resolved formulation), `.convexity` (when well-defined), `.variables`, `.constraints`.
- **Per-tuple sign** — optional 3rd tuple element, default `"=="`. At most one non-equality; 3+ tuples must all be equality.
- `method` parameter — `"auto"` (default), `"sos2"`, `"incremental"`, `"lp"`.
- `active` parameter — binary gating for unit commitment.
- `linopy.breakpoints()` / `linopy.segments()` factories.
- `linopy.slopes_to_points()` utility.
- `linopy.tangent_lines()` low-level helper.

### Auto-dispatch (`method="auto"`)
1. **2-variable inequality on a matching-curvature curve** → `lp` (chord + domain bounds, no aux vars)
2. **Strictly monotonic breakpoints** → `incremental` (deltas + binaries)
3. **Otherwise** → `sos2` (lambdas + SOS2 constraint)
4. **Disjunctive (segments)** → `sos2` with binary segment selection

The resolved method is logged at INFO; when LP is skipped, the log explains why (curvature, NaN layout, tuple count, or `active`).

### Correctness properties
- **`_detect_convexity` is direction-invariant** — ascending and descending x give the same label for the same graph.
- **NaN-padded per-entity breakpoints** are masked everywhere, including the LP chord constraints (padded pieces don't create spurious `y ≤ 0` constraints).
- **`method="lp"` validates curvature + sign + tuple count + `active` absence** upfront; mismatch raises with a clear pointer to `method="auto"`.
- **`PiecewiseFormulation.method` and `.convexity` persist across netCDF round-trip.**

### Architecture (internal)
- `_PwlInputs` dataclass categorizes the user's tuples into `pinned_*` and `bounded_*` slots — explicit role assignment, no positional convention. Carries through the whole dispatch chain.
- `_PwlLinks` is the stacked link representation consumed by SOS2/incremental/disjunctive builders; `_build_links()` produces it from `_PwlInputs`.
- `_try_lp()` / `_resolve_sos2_vs_incremental()` flatten the dispatch — `_add_continuous` is ~10 lines of real logic.
- User tuple order is preserved end-to-end; the bounded one is found via `_PwlInputs`, not by assuming position 0.

### Removed / deprecated
- `PiecewiseExpression`, `PiecewiseConstraintDescriptor`, `linopy.piecewise()` — descriptor pattern gone.
- Old `add_piecewise_constraints` shape that dispatched on `Variable.__le__`/`__ge__`/`__eq__` return types.
- `Variable` operator overloads (`__le__`, `__ge__`, `__eq__`) simplified — no more piecewise dispatch.

## Design principles

- API surface stays minimal: `Model`, `Variables`, `Constraints`, `Expression`, plus the single `PiecewiseFormulation` return type. No user-facing intermediate state.
- Piecewise is a construction layer that produces regular linopy objects.
- One entry point for equality and inequality — per-tuple sign reads as the relation it encodes (`(y, y_pts, "<=")` ⇒ `y ≤ f(...)`).
- Variable names as link coords: the internal `_pwl_var` dimension shows `power, fuel, heat` instead of `0, 1, 2`.

### A) Descriptor pattern (PR #602, previous master)
```python
pw = linopy.piecewise(x, x_pts, y_pts)
m.add_piecewise_constraints(pw == y)
```
**Rejected:** Introduces `PiecewiseExpression` and `PiecewiseConstraintDescriptor` as user-facing state. Breaks the "only Variables, Constraints, Expressions" principle. Structurally limited to 2-variable `x → y`.

### B) Dict of expressions + shared breakpoints DataArray
```python
m.add_piecewise_formulation(exprs={"power": power, "fuel": fuel}, breakpoints=bp)
```
**Considered:** Supports N-variable. But requires string keys to match breakpoint coordinates — an indirection layer.

### C) Global `sign=` keyword (initial implementation, refactored in #664)
```python
m.add_piecewise_formulation((fuel, y_pts), (power, x_pts), sign="<=")
```
The original implementation used a single formulation-level `sign=` kwarg with a positional convention ("the first tuple is the bounded one when `sign != '=='`"). #664 refactored this to per-tuple sign because the role belongs to a specific tuple — encoding it globally + positionally was a wart that required a paragraph of docs to teach. The internal dispatch was rewritten from positional access to explicit categorization (`_PwlInputs` with `bounded_*` and `pinned_*` slots).

### D) Chosen: per-tuple sign on the tuple it applies to
```python
m.add_piecewise_formulation(
    (fuel,  y_pts, "<="),  # bounded (role visible at the call site)
    (power, x_pts),         # pinned
)
```

## Future work this enables

- **2-D triangulated piecewise** (`z ≥ f(x, y)`). The reserved N≥3-inequality slot reads naturally as the epigraph of a bivariate function: `(z, z_pts, ">="), (x, x_pts), (y, y_pts), triangulation=tris`. Vertex coordinates stay as plain 1-D `DataArray`s; the simplex adjacency rides along as one shared sibling kwarg. `_PwlInputs` would grow a single optional field; the formulation builders for the triangulated path are new but consume the same input carrier.
- **Multi-bounded relations** (if a use case appears) — the "at most one bounded tuple" check is a single line; the dispatch chain already handles per-tuple roles.

## Defered Features
- More polished slopes api, potentially with a new class

## Out of scope / follow-ups

Directions the design enables or considered but deliberately not shipping in this PR:

- **2-D triangulated piecewise** (`z ≥ f(x, y)`). The reserved N≥3-inequality slot reads naturally as the epigraph of a bivariate function: `(z, z_pts, ">="), (x, x_pts), (y, y_pts), triangulation=tris`. Vertex coordinates stay as plain 1-D `DataArray`s; the simplex adjacency rides along as one shared sibling kwarg. `_PwlInputs` would grow a single optional field; the formulation builders for the triangulated path are new but consume the same input carrier.
- **Multi-bounded relations** (if a use case appears) — the "at most one bounded tuple" check is a single line; the dispatch chain already handles per-tuple roles. 
- **More polished slopes API, potentially with a new class.** `breakpoints(slopes=..., x_points=..., y0=...)` is shipped today. What's deferred is letting slopes inherit the x grid from a sibling tuple (e.g. `(fuel, Slopes([1.2, 1.4, 1.7], y0=0))` paired with a reference tuple that carries the x grid). Two viable shapes were discussed — extending `breakpoints(slopes=...)` to allow omitted `x_points`, or introducing a small frozen-dataclass `Slopes` value type — each with tradeoffs around type-system visibility and dispatch clarity. `EvolvingAPIWarning` covers either direction as a non-breaking follow-up.

## Notebook examples

`examples/piecewise-linear-constraints.ipynb` (main tutorial):

| # | Section | Feature |
|---|---------|---------|
| 1 | Getting started | Basic 2-variable equality + method inspection via `pwf` |
| 2 | Picking a method | `auto` vs `sos2` vs `incremental` give the same optimum |
| 3 | Disjunctive segments | `segments()` for gaps in operation |
| 4 | Inequality bounds | Per-tuple `"<="` (auto-dispatches LP) |
| 5 | Unit commitment | `active` parameter |
| 6 | N-variable linking | CHP plant (3 variables, joint equality) |
| 7 | Per-entity breakpoints | Fleet with different curves |

`examples/piecewise-inequality-bounds.ipynb` (deep-dive on per-tuple sign / LP):

- Setup with a concave reference curve
- Three methods (LP, SOS2, incremental) verified to produce the same feasible region
- 2-D hypograph feasibility plot
- Curvature × sign rule table — the "wrong region" cases and why
- Auto-dispatch fallback demonstration (non-convex curve, mismatched sign, explicit `method="lp"` raise)

Formulation math lives in the reference page, not in the notebook.

## Test plan

- [x] `pytest test/test_piecewise_constraints.py test/test_piecewise_feasibility.py` — 518 passed
- [x] Full suite (excl. solver/remote) — 1578 passed, 7 skipped
- [x] Both notebooks execute end-to-end against Gurobi
- [x] Ruff lint + format clean, mypy clean on the full repo
- [x] Round-trip tested: `PiecewiseFormulation.method` and `.convexity` persist across `to_netcdf` / `read_netcdf`
- [ ] Review


FBumann:
@FabianHofmann This is pretty urgent to me, as the current piecewise API is already on master and should **NOT** be included in the next release in my opinion.
But i understand that the CSR PR is more important atm.

The current code can be reorganized a bit, but id like to hear your thoughts about the API first.

FBumann:
edit: did some refactoring anyway, grouped in #641

FBumann:
## N-variable disjunctive + follow-up notes

**Done:** Refactored `_add_disjunctive` to use the same stacked N-variable pattern as `_add_continuous`. The 2-variable restriction is removed — disjunctive now works with any number of `(expression, breakpoints)` pairs, using a single `_link` constraint with a `_pwl_var` dimension (same as continuous SOS2/incremental).

FBumann:
## Additional follow-ups

**Single-variable support:** All formulation methods (SOS2, incremental, disjunctive) should support a single `(expression, breakpoints)` pair. Use case: constraining a variable to lie on a set of discrete segments or breakpoints without linking to another variable. The stacking logic should handle `len(pairs) == 1` naturally — just relax the `len(pairs) < 2` check.

**Segments with 2+ breakpoints:** Currently `segments()` only supports pairs of points (start, end) per segment. It should support segments with 2 or more breakpoints, enabling piecewise-within-segment curves (e.g. a nonlinear segment approximated by multiple linear pieces within each disconnected region. Currently, a user could express this with "touching" segments, but its mathematically less efficient).

FBumann:
## Comparison with JuMP's piecewise linear formulations

### JuMP's approach (via PiecewiseLinearOpt.jl)

JuMP's core `piecewiselinear()` returns a single variable representing the PWL output. All auxiliary variables/constraints are added internally. The package offers 8 formulation methods that all share the same convex-combination (λ) core — they differ only in **how SOS2 adjacency is enforced**:

| Method | Binary vars | Notes | linopy |
|---|---|---|---|
| `:SOS2` | 0 | Delegates to solver's native SOS2 branching | ✅ `_add_sos2` |
| `:CC` (Convex Combination) | O(n) | Same as SOS2 but adjacency enforced with explicit binaries instead of SOS2 set | ✅ same formulation as `_add_sos2`but with our internal `sos_reformulation` |
| `:MC` (Multiple Choice) | O(n) | Disaggregated segment copies with binary selection | ✅ essentially `_add_disjunctive` — and we extend it with gap support via `segments()` |
| `:Incremental` | O(n) | Delta chaining | ✅ `_add_incremental` |
| **`:Logarithmic`** (default) | **O(log n)** | Gray code encoding | ❌ not yet |
| `:DisaggLogarithmic` | O(log n) | Disaggregated + Gray codes | ❌ not yet |
| `:ZigZag` | O(log n) | Zig-zag codes | ❌ not yet |
| `:ZigZagInteger` | 1 general int | Most compact | ❌ not yet |

The default is `:Logarithmic` — `ceil(log2(n-1))` binary variables, significantly more efficient for branch-and-bound with many breakpoints.

JuMP also supports **bivariate PWL** via triangulation (UnionJack, K1, Stencil, BestFit patterns) for `z = f(x, y)` with independent inputs.

Reference: Huchette & Vielma, *"Nonconvex piecewise linear functions: Advanced formulations and simple modeling tools"* (2017).

### Two separate concerns: piecewise structure vs. SOS2 enforcement

JuMP mixes these into a single `method` parameter. We can do better by separating them:

**Piecewise formulation structure** (what `add_piecewise_formulation` handles):

| Method | What it does | linopy |
|---|---|---|
| **SOS2** | λ weights + sum=1 + linking constraints. Delegates adjacency enforcement to `add_sos_constraints`. | ✅ `_add_sos2` |
| **Incremental** | Completely different structure — δ deltas + binary chaining. No λ, no SOS2 at all. | ✅ `_add_incremental` |
| **Disjunctive** | Per-segment λ + binary segment selection. Uses SOS2 within each segment, but segment decomposition is piecewise logic. | ✅ `_add_disjunctive` |

**SOS2 adjacency enforcement** (what `add_sos_constraints` / `sos_reformulation.py` handles — orthogonal to piecewise):

| Method | Binary vars | Notes | linopy |
|---|---|---|---|
| Native | 0 | Solver handles branching | ✅ `add_sos_constraints` |
| CC / Big-M | O(n) | `x[i] <= M[i] * (z[i-1] + z[i])` | ✅ `sos_reformulation.py` (auto via `reformulate_sos=True`) |
| Logarithmic (Gray code) | O(log n) | Best trade-off for many breakpoints | ❌ not yet, high value follow op ⭐ |
| DisaggLogarithmic | O(log n) | Disaggregated + Gray codes | ❌ not yet |
| ZigZag | O(log n) | Different code pattern, similar performance | ❌ not yet |
| ZigZagInteger | 1 general int | Most compact | ❌ not yet |

This separation means:
- Piecewise stays simple: 3 structural methods, already complete
- SOS2 reformulations benefit **all** SOS constraints in the model, not just piecewise
- The two choices compose: e.g. disjunctive piecewise + logarithmic SOS2 enforcement

JuMP bundles all 8 as piecewise methods. We separate structure from enforcement — cleaner architecture, broader applicability.

### What linopy has on top of JuMP

| Feature | Description |
|---|---|
| **Disconnected segments** | `segments()` factory + `_add_disjunctive` for forbidden operating zones with gaps (e.g. off or 50–80 MW). JuMP's MC handles contiguous segments only. |
| **N-variable linking** | Link 3+ expressions through shared λ weights (CHP: power/fuel/heat). JuMP's `piecewiselinear()` is 1→1 only. |
| **`active` parameter** (unit commitment) | Binary variable gates the entire PWL — when `active=0`, all aux variables are zero. In JuMP this requires manual formulation. |
| **`tangent_lines()` utility** | Pure LP inequality bounds (no auxiliary variables). Returns a `LinearExpression` per segment for use with regular `add_constraints`. |
| **Per-entity breakpoints** | Different curves per generator via `breakpoints({"gas": [...], "coal": [...]}, dim="gen")` with automatic xarray broadcasting. |
| **Auto method selection** | `method="auto"` detects monotonicity and chooses incremental (faster) vs SOS2 (flexible). JuMP requires explicit method choice. |

### Why our architecture doesn't block any of these

The design in this PR is **method-agnostic by construction**:

1. **The dispatch is a simple `if/else` on a string** in `_add_continuous`. The 3 structural methods are already complete.

2. **All formulation methods reduce to the same primitives**: `add_variables(continuous/binary/integer)` + `add_constraints` + optionally `add_sos_constraints`. Every JuMP method uses exactly these building blocks.

3. **Result tracking is formulation-blind**: `PiecewiseFormulation` snapshots which variable/constraint names were added via before/after diffing. It doesn't care *how* they were structured — any new method's artifacts get captured automatically.

4. **SOS2 reformulations are additive**: Adding logarithmic/zigzag to `sos_reformulation.py` requires zero changes to the piecewise layer — it just gets faster SOS2 enforcement for free.

5. **No formulation-specific assumptions leak into the public API**. The user writes `method="sos2"` and gets back the same `PiecewiseFormulation` with `.variables` and `.constraints`, regardless of how SOS2 is enforced.

### Our N-variable linking vs. JuMP's triangulation

These solve **different problems** (complementary, not competing):

| | Our N-variable (1D curve) | Triangulation (2D+ mesh) |
|---|---|---|
| **Inputs** | 1 hidden parameter (e.g. load) | 2+ independent variables (x, y) |
| **Breakpoints** | 1D sequence | 2D grid, triangulated into simplices |
| **λ weights** | 2 adjacent points (SOS2 along 1 axis) | 3 vertices of one triangle |
| **Degrees of freedom** | 1 | 2+ |
| **Use case** | Coupled outputs of one parameter (CHP, efficiency curves) | `z = f(x, y)` with independent inputs |

Our 3-variable `(power, fuel, heat)` links three outputs coupled by a single operating parameter — you can't set power=60 and heat=25 independently; the curve determines heat given power. Triangulation handles `z = f(x, y)` where x and y move independently (e.g., cost as a function of temperature AND load).

### Follow up: Triangulation

Triangulation is mathematically quite a bit more complex and needs different data. It takes a mesh instead of a 1D breakpoint sequence. (mesh). Its clearly a follow up, probably as a new `Model.add_*()*` method that can reuse the `PiecewiseFormulation` return type, name tracking, and model primitives.

FBumann:
@coroa Can we discuss this together with @FabianHofmann ?

coroa:
> @coroa Can we discuss this together with @FabianHofmann ?

@FabianHofmann is on vacation this week. Can we set up something next week? Or is pressing more urgent than that?

FBumann:
> > @coroa Can we discuss this together with @FabianHofmann ?
> 
> @FabianHofmann is on vacation this week. Can we set up something next week? Or is pressing more urgent than that?

Ah ok i missed that. No, next week is fine :) Could you setup a meeting?

FBumann:
## From call:

- Add a sign parameter to `add_piecewise_constraints()`
- `add_piecewise_constraints()` should forward to tangent lines if possible
- Add new `Slopes` object, drop slopes support from `breakpoints`
- Maybe add a reference variable in `add_piecewise_constraints()` so other variables/slopes can refer to it (especially for tangent lines)

FBumann:
@coroa I thought a lot about the inequality in non lp piecewise relationships.
I added a notebook and plot about it in #662 (599278862d05c668d211a15712332473e26262fb). Essentially, the sign needs to be "translated" to behave like tangent lines. And we need to make a decision about the convention, because givin all options to the user makes a pretty bad UX whie "allowing" cases which probably noone wants.
Here is a visualization for feasible regions for a 2-variable case with SOS2/lambda/delta formulations based on which sign is used for which variable:

fig


Compared to tangent line one:

tangent

We can see that we need to fix one of the variables to the slope! TO be precise, all but one variable needs to be fixed!

FBumann:
@coroa What do you think about the new api?
I think its quite good and clear now. The testing is also already quite good imo.

One thing we should discuss is the notebooks and docs. If you have the time to read it, feedback and potentially changes would be great! 
The one who implemented it usually has a different angle on the docs than the reviewer/user.
https://linopy--638.org.readthedocs.build/en/638/piecewise-linear-constraints.html

FBumann:
@coroa One API detail im not sure about yet, is where to but the sign.
It could be per tuple as a third optional parameter:
```python
m.add_piecewise_formulation(
   (power, xp, "<="),
   (fuel, yp)
)
```
with defaulting to '==' where not specified.

I decided against this api for simplicity, as using more than i single equality produces mathematically weird bounds as far as i understand it (only relevant for the 3 var case). ALso, for the 2 variable LP case, it doestn make any sense to have more than a single sign. So there is a lot of API surface that would lead to errors we need to catch.

We could however also just disallow the 3 variable case with inequalities for now, as i don`t understand the use case that well. This would allow the arguably more obvious notation of having the sing right next to the expression...

With the `EvolvingAPIWarning` however, im fine with keeping it and adjusting it if needed.

Looking forward, this API seems better suited for triangular bounding, where we could actually allow multiple inequalities. But the implementation of this is not part of this PR and quite complex i think, so we can just change it as soon as this potential feature lands

FBumann:
One thing I didn't cover yet, is a dedicated slopes object. Should we add it? Or can we do it in a follow up?

FBumann:
> @coroa One API detail im not sure about yet, is where to but the sign. It could be per tuple as a third optional parameter:
> 

I thought a bit more about it, and its much clearer having the sign next to the variable/factor. THis also extends naturally to triangulation etc, and feels less "magic". I refactored it in #664. This also made the docs cleaner.

I also decided to disallow inequalities for 3+ variables, as im not sure about the mathematical formulation and would rather ship the PR as is without further discussion on the not yet needed 3+ != case.

FBumann:
## Refactor: Disambiguate usage of "segment" by renaming to "piece" in most contexts
Both terms are used in the PWL literature — Wikipedia and the Northwestern optimization wiki use them interchangeably; JuMP's PiecewiseLinearOpt.jl prefers "pieces", Pyomo's API leans "segments". So the rename isn't about correctness -  its about readability.
I argue strongly in favor of **Pieces**, while keeping Segments as a collection of curves, where each curve/segment consist of several pieces (disjunctive). This is clear and mathematically sound.

I added a clear definition into the docs

FBumann:
@coroa @FabianHofmann I think everything is sorted out now and the api is both intuitive, well documented, tested and future proof.

Please take a look

FabianHofmann:
this looks cool, well done @FBumann ! I will have a look in the afternoon!

pyproject.toml
```diff
@@ -110,6 +110,18 @@ norecursedirs = ["dev-scripts", "doc", "examples", "benchmark", "benchmarks"]
...
+    # Silence our own EvolvingAPIWarning inside the test suite so the
+    # piecewise tests don't emit 500+ warnings.  Users still see them.
+    # Match by message prefix on the builtin FutureWarning base class
+    # rather than on ``linopy.EvolvingAPIWarning`` directly — using the
+    # class reference here would force pytest to import linopy at
+    # config-parse time, which loads ``linopy.variables`` from
+    # site-packages and then conflicts with ``--doctest-modules``
+    # collection of ``linopy/variables.py`` in the source tree on
+    # Windows CI.
+    "ignore:piecewise:FutureWarning",
```
FabianHofmann, line 123: ~why do we get futurewarnings on the piecewise implementation? this is not released, we can go wild without any futurewarnings. remove it~

linopy/piecewise.py
```diff
@@ -59,11 +71,113 @@
...
+    )
+
+    def __init__(
+        self,
+        name: str,
+        method: str,
+        variable_names: list[str],
+        constraint_names: list[str],
+        model: Model,
+        convexity: Literal["convex", "concave", "linear", "mixed"] | None = None,
```
FabianHofmann: should use `PWL_CONVEXITIES` from costants.py to have a single source of truth. there is other places like this

linopy/constraints.py
```diff
@@ -733,25 +733,35 @@ def _formatted_names(self) -> dict[str, str]:
...
         return r
 
+    def __repr__(self) -> str:
+        r = "linopy.model.Constraints"
+        line = "-" * len(r)
+        r += f"\n{line}\n"
+        r += self._format_items()
+        return r
+
+    def _repr_filtered(self, exclude: set[str]) -> str:
```
FabianHofmann: drop and use `_format_items` directly in the code

linopy/constraints.py
```diff
@@ -733,25 +733,35 @@ def _formatted_names(self) -> dict[str, str]:
...
                 if ds.coords
                 else ""
             )
             r += f" * {name}{coords}\n"
-        if not len(list(self)):
+        if count == 0:
             r += "<empty>\n"
         return r
 
+    def __repr__(self) -> str:
```
FabianHofmann, line 754:
```suggestion
    def __repr__(self) -> str:
        """
        Return a string representation of the linopy model.
        """
```

linopy/model.py
```diff
@@ -498,17 +503,45 @@ def __repr__(self) -> str:
...
+                for var in pwl.variables.data.values():
+                    for d in var.coords:
+                        if not str(d).startswith("_") and str(d) not in user_dims:
+                            user_dims.append(str(d))
+                dims_str = f" ({', '.join(user_dims)})" if user_dims else ""
+                result += (
+                    f" * {pwl.name}{dims_str}"
+                    f" — {pwl.method}, {n_vars} vars, {n_cons} cons\n"
+                )
+
```
FabianHofmann: this code should go into the piecewise.py and be a one-liner call here, perhaps also `_repr_filtered`, or `_repr_summary`

linopy/expressions.py
```diff
@@ -668,40 +647,13 @@ def __div__(self: GenericExpression, other: SideLike) -> GenericExpression:
...
-    def __eq__(self, rhs: PiecewiseExpression) -> PiecewiseConstraintDescriptor: ...
-
-    @overload
-    def __eq__(self, rhs: SideLike) -> Constraint: ...
-
-    def __eq__(self, rhs: SideLike) -> Constraint | PiecewiseConstraintDescriptor:
-        descriptor = _to_piecewise_constraint_descriptor(self, rhs, "==")
-        if descriptor is not None:
-            return descriptor
+    def __eq__(self, rhs: SideLike) -> Constraint:  # type: ignore
```
FabianHofmann: I would expect that we don't need the `# type: ignore` here anymore

linopy/io.py
```diff
@@ -1147,6 +1147,18 @@ def with_prefix(ds: xr.Dataset, prefix: str) -> xr.Dataset:
     ds = ds.assign_attrs(scalars)
     if m._relaxed_registry:
         ds.attrs["_relaxed_registry"] = json.dumps(m._relaxed_registry)
+    if m._piecewise_formulations:
+        ds.attrs["_piecewise_formulations"] = json.dumps(
+            {
+                name: {
+                    "method": pwl.method,
+                    "variables": pwl.variable_names,
```
FabianHofmann: use `"variable_names"` (same for constraints) for naming consistency?

linopy/model.py
```diff
@@ -498,17 +503,45 @@ def __repr__(self) -> str:
...
+        result += f"\nStatus:\n-------\n{self.status}"
+        return result
+
+    def _piecewise_names(self) -> set[str]:
+        """Return all variable/constraint names belonging to piecewise formulations."""
+        names: set[str] = set()
+        for pwl in self._piecewise_formulations.values():
+            names.update(pwl.variable_names)
+            names.update(pwl.constraint_names)
```
FabianHofmann: piecewise variables and constraints cannot have the same name right?

FabianHofmann (changes requested):
first round of small requests mostly outside of the piecewise module.

pyproject.toml
FabianHofmann: okay, now I actually understand this warning and it might lead to feedback from the community. I find it a bit verbose though. Could we only raise it one time per session and not on all API calls?

FabianHofmann:
@FBumann just went through the notebook, fantastic work. mind if I push some decorative changes to the notebook?

FBumann:
Of course not. Go ahead. You dont have to ask that going forward. If you keep a proper commit history, i can check what you did and complain if i need to :)

linopy/constraints.py
FBumann: I added a repr

linopy/model.py
FBumann: Yes, they can. As regular variables and constraints can share a name too. They do not currently though.

pyproject.toml
FBumann: Did it

FabianHofmann:
<details> 

<summary> Here is the review from my agents: </summary>


## Critical
* **`piecewise.py` — `_lp_eligibility` (1172-1200) and `_try_lp` (1278-1308) duplicate every check.** Two sources of truth for LP eligibility rules. Collapse into one helper returning `(ok, reason)` and raise from `_try_lp` when `method == "lp"`.
* **`__init__.py:60-66` — `__all__` no longer alphabetical.** `tangent_lines` sits between `breakpoints` and `segments`; `align`/`merge`/`options`/`read_netcdf` come after `slopes_to_points`. The latter group was already broken on master, but the new `tangent_lines` insertion compounds it. Likely to trip ruff `RUF022`.
* **No tests for `EvolvingAPIWarning` at all.** `grep -rn "EvolvingAPIWarning" test/` returns zero matches. Per-key dedup, `stacklevel=3`, and the "no double-warn from `_tangent_lines_impl` inside `_add_lp`" claim are all unverified. Module-global dedup set will make tests order-dependent without a reset fixture.
* **NetCDF roundtrip is a single 2-tuple equality case** (`test_formulation_survives_netcdf`, line 2048). Missing: multi-formulation models, disjunctive formulations, bounded `<=`/`>=` tuples, `method="lp"` (empty `variable_names`), and `convexity != None`.

## High

### Code quality / DRY
- **`piecewise.py:648-659`** `_validate_breakpoint_shapes` repeats the BREAKPOINT_DIM error twice for `bp_a`/`bp_b`. Loop the pair.
- **`piecewise.py:1227-1267`** `_build_links` has two near-identical equality-branch builders.
- **`piecewise.py:1038-1059`** `_PwlInputs` two-branch construction differs in 4 fields only.
- **`piecewise.py:127-140`** `PiecewiseFormulation` hand-rolls `__slots__`+`__init__`; use `@dataclass(slots=True)` to drop ~25 lines of boilerplate.
- **`piecewise.py:154-188`** `_user_dims` and `__repr__` duplicate the dim-collection / `_`-prefix-skip loop.

### Type / API
- **`piecewise.py:131`** `method: str` should be `PWL_METHOD` (or narrowed Literal); same for `convexity`.
- **`piecewise.py:797-805`** `add_piecewise_formulation(**kwargs: object)` defeats static type-checking on callers and conflicts with the no-`getattr`-style rule. Use explicit `sign: str | None = None` and raise.
- **`io.py:1265-1269`** Inconsistent fail-fast posture — `convexity` uses `d.get("convexity")` (silent `None`) while every other field uses `d["..."]` (KeyError). Pick one.

### Tests
- **`PiecewiseFormulation.variables` / `.constraints` / `__repr__` / `_user_dims` essentially unexercised.** Only `f.method`/`f.convexity` are read in tests.
- **Auto-method success log never asserted** (only the LP-skip caplog branch at line 1850 exists).
- **`_lp_eligibility` reason strings not individually asserted.** Each branch returns a distinct user-facing diagnostic; parametrize directly.
- **`_add_continuous` resolution** for `method ∈ {"sos2","incremental","lp","auto"}` not parametrized (only `["sos2","incremental"]` at line 1732).

## Medium

### Code quality
- `piecewise.py:62-74` Custom `_emitted_evolving_warnings` set is over-engineered; `warnings.simplefilter("once", EvolvingAPIWarning)` would suffice.
- `piecewise.py:1364` `_resolve_sos2_vs_incremental` final `raise ValueError("unknown method ...")` is dead — `add_piecewise_formulation` validates against `PWL_METHODS` upstream.
- `piecewise.py:1156-1169` `_PwlInputs.all_bps`/`all_coords`/`all_exprs` three near-identical methods; collapse.
- `piecewise.py:1033` link-coord fallback to `str(i)` can collide with a user variable named `"1"`. Use `f"_pwl_{i}"`.
- `model.py:506` indirect `_grouped_names` import → consider exposing as `model._piecewise_grouped_names()`.
- `io.py:1259-1269` move reconstruction into `PiecewiseFormulation.from_dict` for symmetry with `to_netcdf`.

### Type / API
- `piecewise.py:1146` `link_dim = "_pwl_var"` is ad-hoc; promote to `PWL_LINK_DIM` constant alongside `BREAKPOINT_DIM`/`SEGMENT_DIM`.
- `piecewise.py:1448` `piece_dim = f"{dim}_piece"` recomputed in `_add_continuous` instead of reusing `LP_PIECE_DIM` (imported at line 28, used in `_rename_to_pieces` at 236-237). Two sources of truth.
- `piecewise.py:191` `_grouped_names` is opaque naming → `_piecewise_member_names`.
- `piecewise.py` `tangent_lines` (~600-635) and `add_piecewise_formulation` (~900-925) docstrings are long; the `EvolvingAPIWarning` silencing recipe is duplicated verbatim across both sites.
- `variables.py:545` bare `# type: ignore` should be `# type: ignore[override]` (regression — was `[override]` before).
- `piecewise.py:1116-1125` `_stack_along_link` accepts `Sequence[DataArray | xr.Dataset]` but returns `DataArray`; either generic via `TypeVar` or split.

### Tests
- `add_piecewise_formulation` unexpected-kwargs `TypeError` path untested.
- `_breakpoints_from_slopes` 3D slopes path, dict `y0` missing key, per-entity length mismatch — direct tests missing.
- `_validate_breakpoint_shapes` first-arg branch (mirror of the `bp_b` test at 1360) untested.
- Public `breakpoints` / `segments` factories: empty list, single point, all-NaN, ±inf, and duplicate x (zero `dx` → div-by-zero in `tangent_lines` / `_detect_convexity`) all uncovered.
- LP `_add_lp` `piece_count == 1` shortcut (`piece_mask` all-True) not tested.
- `_broadcast_points` `"Cannot broadcast: missing dim"` branch has no negative test.
- SOS2/incremental masking branch of bounded path with NaN not directly verified.

## Low

- `piecewise.py:229-231` `_strip_nan` returns `list[float]`; use `arr[~np.isnan(arr)].tolist()` (faster, type-consistent).
- `piecewise.py:1187, 1287` two `assert ... is not None  # narrowed by is_equality check` comments — drop the trailing rationale comments.
- `piecewise.py:550-551` `_tangent_lines_impl` calls `_coerce_breaks` before the type check; reverse for fail-fast.
- `piecewise.py:216-217` `_repr_summary` uses `len(pwl.variables)` / `len(pwl.constraints)` (constructs views) — `len(pwl.variable_names)` is O(1).
- `piecewise.py:1239` `_PwlLinks.eq_bp = stacked_bp` is redundant in the equality case.
- `constants.py:58,60` `PWL_METHODS` / `PWL_CONVEXITIES` could be `frozenset` for immutability.
- `constants.py:84-107` `EvolvingAPIWarning` ~24-line docstring is verbose; two silencing examples could be one.
- `model.py:808` `add_piecewise_formulation = add_piecewise_formulation` mypy-invisible binding survives refactor (pre-existing).
- `test_piecewise_feasibility.py:274-356` `TestHandComputedAnchors` six near-identical methods — collapse via parametrization.
- `test_piecewise_constraints.py:1823` manual `for sign in signs` loop instead of `pytest.parametrize`.
- `test_piecewise_constraints.py:1933-2034` `from linopy.piecewise import _detect_convexity` repeated in every method — hoist to module scope.
- `test_formulation_survives_netcdf:2048` excludes `_model` from the slot comparison without affirming the back-reference is rebound — that back-reference is what makes `.variables` / `.constraints` work after read.
- `TestTangentLines:370-421` covers var/linexpr/dataarray/constraint paths but no `BreaksLike` size-mismatch or NaN tests.

## Positive observations

- `*pairs` variadic API at `piecewise.py:797` is a clean upgrade over the old descriptor `>=`-overload trickery.
- Per-tuple `(expr, breakpoints, "<=")` sense convention reads naturally and is well-checked.
- `EvolvingAPIWarning` design (subclass `FutureWarning` at `constants.py:84`, per-key dedup at `piecewise.py:66-74` with `stacklevel=3`, message-prefix targeting) is well thought out.
- `PiecewiseFormulation` return value with `.variables` / `.constraints` views is a real improvement over returning a bare `Constraint`.
- **API asymmetry to consider:** `add_piecewise_formulation` is a `Model` method (`model.py:808`) but `tangent_lines` is module-level (`piecewise.py:573`). Either expose `model.tangent_lines(...)` for discoverability or document the split.

---

## Recommended action plan

* Collapse `_lp_eligibility` / `_try_lp` duplication.
* Restore alphabetical `__all__` sort (consider also fixing the pre-existing ordering).
* Add `EvolvingAPIWarning` tests with a fixture clearing `_emitted_evolving_warnings`.
* Expand netCDF roundtrip tests (multi-formulation, bounded, LP, convexity).
* Tighten types: `method: PWL_METHOD`; replace `**kwargs: object` with explicit `sign:` param.
* Make `io.py:1269` convexity loading consistent with other field reads (drop `.get`).
* Cover `PiecewiseFormulation.variables` / `.constraints` / `__repr__`, parametrize `_lp_eligibility` reasons, assert the auto-resolution log.
* DRY refactors in `_validate_breakpoint_shapes` (648-659), `_build_links` (1227-1267), `_PwlInputs` (1038-1059).

**Defer to follow-up cleanup PR (Medium + Low):** dataclass conversion, constants extraction (`PWL_LINK_DIM`, reuse `LP_PIECE_DIM`), docstring trimming, test parametrization cleanups, `_strip_nan` speedup, comment removal.

</details>

I would say nothing of this is really critical, but I think the recommended points should be easy to address. perhaps @FBumann you want to go through this and see what of it you find helpful. I would say nothing of this is a requirement.

FBumann:
@FabianHofmann Seems reasonable.

A few comments on things i disagree on:
1. kwargs: Good catch, but those should just go entirely. They are mere an artifact from iterations during dev...
2. `_build_links` and `_PwlInputs`: I dont see how we could compress this meaningfully.

FabianHofmann (approved):

FBumann:
Great quality improvements @FabianHofmann 🎉

PR #639, opened 2026-04-01
fix: make assert_linequal term-order invariant

FBumann:
## Summary
- `assert_linequal` now sorts both sides by variable labels along `_term` before comparing, so expressions with different term orderings are correctly recognized as semantically equal.
- Fixes false failures when comparing constraints that went through the CSR round-trip (`freeze_constraints=True`) against freshly built expressions, since `csr.sum_duplicates()` sorts terms by variable label.

## Reproducer (from #630 comment)
```python
from linopy import Model
import pandas as pd

m = Model(freeze_constraints=True)
coords = pd.RangeIndex(3, name="time")
x = m.add_variables(coords=[coords], name="x")
y = m.add_variables(coords=[coords], name="y")

m.add_constraints(y - x == 0, name="con")

desired = y == x
assert_linequal(m.constraints["con"].lhs, desired.lhs)  # was failing, now passes
```

## Test plan
- [x] Reproducer passes
- [x] All 2665 existing tests pass
- [x] Linter and pre-commit hooks pass

### DIFF
--- a/linopy/testing.py
+++ b/linopy/testing.py
@@ -1,11 +1,27 @@
 from __future__ import annotations
 
+import numpy as np
 from xarray.testing import assert_equal
 
+from linopy.constants import TERM_DIM
 from linopy.constraints import ConstraintBase, _con_unwrap
 from linopy.expressions import LinearExpression, QuadraticExpression, _expr_unwrap
 from linopy.model import Model
 from linopy.variables import Variable, _var_unwrap
+
+
+def _sort_by_vars_along_term(expr: LinearExpression) -> LinearExpression:
+    """Sort a linear expression's terms by variable labels along _term."""
+    ds = expr.data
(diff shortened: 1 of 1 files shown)


FabianHofmann:
@FBumann would you mind raising this pr from a branch on linopy and not on fork? as for now, the CI is not triggered fully

FBumann:
@FabianHofmann I merged that by accident.
Ill open a PR. Can you revert it? Or maybe leave it if you find it passing...?

FabianHofmann:
> @FabianHofmann I merged that by accident. Ill open a PR. Can you revert it? Or maybe leave it if you find it passing...?

all good, no problem. not important here. but for prs against master it would be

FBumann:
@FabianHofmann definitely...

PR #647, opened 2026-04-15
fix: use xarray.Dataset copy instead of constructor

maurerle:
minimal non-working example:

```python
import pypsa
network = pypsa.examples.ac_dc_meshed()
network.optimize()
```

since the latest xarray version, passing a Dataset as `data_vars` to the Dataset constructor is not supported. Use `ds.copy()` to create a copy of a Dataset.

## Checklist

### DIFF
--- a/linopy/expressions.py
+++ b/linopy/expressions.py
@@ -284,7 +284,7 @@ class LinearExpressionGroupby:
                 index.names = [str(col) for col in orig_group.columns]
                 index.name = GROUP_DIM
                 new_coords = Coordinates.from_pandas_multiindex(index, GROUP_DIM)
-                ds = xr.Dataset(ds.assign_coords(new_coords))
+                ds = ds.assign_coords(new_coords)
 
             ds = ds.rename({GROUP_DIM: final_group_name})
             return LinearExpression(ds, self.model)
@@ -391,8 +391,7 @@ class BaseExpression(ABC):
         coeffs_vars_dict = {str(k): v for k, v in coeffs_vars.items()}
         data = assign_multiindex_safe(data, **coeffs_vars_dict)
 
-        # transpose with new Dataset to really ensure correct order
-        data = Dataset(data.transpose(..., TERM_DIM))
+        data = data.transpose(..., TERM_DIM)
 
         # ensure helper dimensions are not set as coordinates
         if drop_dims := set(HELPER_DIMS).intersection(data.coords):
@@ -2098,7 +2097,7 @@ class QuadraticExpression(BaseExpression):
             raise ValueError(f"Size of dimension {FACTOR_DIM} must be 2.")
 
         # transpose data to have _term as last dimension and _factor as second last
-        data = xr.Dataset(data.transpose(..., FACTOR_DIM, TERM_DIM))
+        data = data.transpose(..., FACTOR_DIM, TERM_DIM)
         self._data = data
 
     @property
--- a/linopy/model.py
+++ b/linopy/model.py
@@ -359,7 +359,9 @@ class Model:
         """
         Set the parameters of the model.
         """
-        self._parameters = Dataset(value)
+        self._parameters = (
+            value.copy() if isinstance(value, Dataset) else Dataset(value)
+        )
 
     @property
     def solution(self) -> Dataset:
--- a/doc/release_notes.rst
+++ b/doc/release_notes.rst
@@ -27,6 +27,7 @@ Upcoming Version
 * Add ``format_labels()`` on ``Constraints``/``Variables`` and ``format_infeasibilities()`` on ``Model`` that return strings instead of printing to stdout, allowing usage with logging, storage, or custom output handling. Deprecate ``print_labels()`` and ``print_infeasibilities()``.
 * Add ``fix()``, ``unfix()``, and ``fixed`` to ``Variable`` and ``Variables`` for fixing variables to values via equality constraints. Supports automatic rounding for integer/binary variables.
 * Add ``relax()``, ``unrelax()``, and ``relaxed`` to ``Variable`` and ``Variables`` for LP relaxation of integer/binary variables. Supports partial relaxation via filtered views (e.g. ``m.variables.integers.relax()``). Semi-continuous variables raise ``NotImplementedError``.
+* Add compatibility to latest xarray version.
 
 
 Version 0.6.6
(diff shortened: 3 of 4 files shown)


FBumann (commented):
Thanks for fixing this! Small suggestion on the `.copy()` calls:

The three cases in `expressions.py` don't need `.copy()` — `transpose()` and `assign_coords()` already return a fresh `Dataset`, so the copy is a pure allocation overhead on every call. `BaseExpression.__init__` runs on every arithmetic operation, so this adds up on models that build many expressions.

Specifically:
- `expressions.py:287`: `ds.assign_coords(new_coords).copy()` → `ds.assign_coords(new_coords)`
- `expressions.py:395`: `data.transpose(..., TERM_DIM).copy()` → `data.transpose(..., TERM_DIM)`
- `expressions.py:2101`: `data.transpose(..., FACTOR_DIM, TERM_DIM).copy()` → `data.transpose(..., FACTOR_DIM, TERM_DIM)`

The xarray changelog recommends `.copy()` specifically as a replacement when you *did* want the old `Dataset(ds)` behavior. But in these spots, the original `Dataset(...)` wrapper wasn't actually doing anything useful — it was silently dropping attrs (which the changelog notes) and adding a redundant materialization. Just removing the wrapper is correct.

For `model.py:362` (the `parameters` setter), keeping `.copy()` is defensible since it gives defensive-copy semantics. Though strictly speaking, the old code didn't provide that either, so dropping `.copy()` there is also fine if you prefer.

linopy/expressions.py
```diff
@@ -2098,7 +2100,7 @@ def __init__(self, data: Dataset | None, model: Model) -> None:
             raise ValueError(f"Size of dimension {FACTOR_DIM} must be 2.")
 
         # transpose data to have _term as last dimension and _factor as second last
-        data = xr.Dataset(data.transpose(..., FACTOR_DIM, TERM_DIM))
+        data = data.transpose(..., FACTOR_DIM, TERM_DIM).copy()
```
FBumann:
```suggestion
        data = data.transpose(..., FACTOR_DIM, TERM_DIM)
```

transpose already creates a shallow copy

linopy/expressions.py
```diff
@@ -391,8 +391,10 @@ def __init__(self, data: Dataset | Any | None, model: Model) -> None:
         coeffs_vars_dict = {str(k): v for k, v in coeffs_vars.items()}
         data = assign_multiindex_safe(data, **coeffs_vars_dict)
 
-        # transpose with new Dataset to really ensure correct order
-        data = Dataset(data.transpose(..., TERM_DIM))
+        # Ensure correct dimension order and materialize a new Dataset.
+        # Wrapping an existing Dataset in Dataset(...) is no longer supported
+        # by newer xarray versions.
+        data = data.transpose(..., TERM_DIM).copy()
```
FBumann:
```suggestion
        data = data.transpose(..., TERM_DIM)
```
transpose already creates a shallow copy

linopy/expressions.py
```diff
@@ -284,7 +284,7 @@ def sum(self, use_fallback: bool = False, **kwargs: Any) -> LinearExpression:
                 index.names = [str(col) for col in orig_group.columns]
                 index.name = GROUP_DIM
                 new_coords = Coordinates.from_pandas_multiindex(index, GROUP_DIM)
-                ds = xr.Dataset(ds.assign_coords(new_coords))
+                ds = ds.assign_coords(new_coords).copy()
```
FBumann:
```suggestion
                ds = ds.assign_coords(new_coords)
```
assign_coords already creates a shallow copy

FBumann (commented):
Im not sure if we should exclude highspy==1.14.0 from linopy deps. This PR gets the CI working, but users will get wring results with highs. Not fully our responsibility, but worth considering

coroa:
> Im not sure if we should exclude highspy==1.14.0 from linopy deps. This PR gets the CI working, but users will get wring results with highs. Not fully our responsibility, but worth considering

Yes, we should, but not in this PR.

linopy/expressions.py
coroa:
```suggestion
        # Ensure dimension order
        data = data.transpose(..., TERM_DIM)
```

test/test_optimization.py
```diff
@@ -897,8 +897,17 @@ def test_modified_model(
...
+    # See https://github.com/PyPSA/linopy/issues/648
+    extra_kwargs: dict = {"presolve": "off"} if solver == "highs" else {}
+
     status, condition = modified_model.solve(
-        solver, io_api=io_api, explicit_coordinate_names=explicit_coordinate_names
+        solver,
+        io_api=io_api,
+        explicit_coordinate_names=explicit_coordinate_names,
+        **extra_kwargs,
     )
```
coroa: Shouldn't be part of this PR

linopy/model.py
```diff
@@ -359,7 +359,10 @@ def parameters(self, value: Dataset | Mapping) -> None:
         """
         Set the parameters of the model.
         """
-        self._parameters = Dataset(value)
+        if isinstance(value, Dataset):
+            self._parameters = value.copy()
+        else:
+            self._parameters = Dataset(value)
```
coroa, line 365:
```suggestion
        self._parameters = value.copy() if isinstance(value, Dataset) else Dataset(value)
```

maurerle:
The CI does not run through because of missing #654

linopy/expressions.py
maurerle: thanks!

linopy/expressions.py
```diff
@@ -391,8 +391,10 @@ def __init__(self, data: Dataset | Any | None, model: Model) -> None:
         coeffs_vars_dict = {str(k): v for k, v in coeffs_vars.items()}
         data = assign_multiindex_safe(data, **coeffs_vars_dict)
 
-        # transpose with new Dataset to really ensure correct order
-        data = Dataset(data.transpose(..., TERM_DIM))
+        # Ensure correct dimension order and materialize a new Dataset.
+        # Wrapping an existing Dataset in Dataset(...) is no longer supported
```
FBumann:
```suggestion
```
Remove this explanation. After we merge this it adds nothing. Its more like a commit message

linopy/expressions.py
```diff
@@ -391,8 +391,10 @@ def __init__(self, data: Dataset | Any | None, model: Model) -> None:
...
+        # by newer xarray versions.
```
FBumann:
```suggestion
```
Remove this explanation. After we merge this it adds nothing. Its more like a commit message

coroa:
Ok, thanks. I'll see if i can get a release out today or latest on Monday.

Issue #645, opened 2026-04-07
Add model diagnostics module


FBumann:
## Describe the feature you'd like to see

Add a `model.diagnostics` module/accessor that provides insights into model structure, quality, and performance characteristics. This module would be **experimental** — the API may change between releases as we learn what's most useful. Users will be warned accordingly.

### Motivation

Users often build large models and have limited visibility into why solves are slow or produce unexpected results. A diagnostics module gives them tools to inspect and understand their models without needing deep solver expertise.

### Ideas and use cases

Below is a non-exhaustive list of possible diagnostics. Not all need to be implemented at once — the module can grow over time.

#### Pre-solve (no solve needed, pure matrix/model analysis)

| Diagnostic | Use case |
|---|---|
| **Scaling report** — coefficient range ratios per row/column, flagging problematic ratios | Identify numerical issues that slow down solvers or cause inaccurate results |
| **Big-M detection** — find constraints where binary variables appear with very large coefficients | Spot big-M formulations that weaken LP relaxation and slow branch-and-bound |
| **RHS/coefficient mismatch** — flag constraints where RHS is on a very different scale than coefficients | Catch common modeling mistakes that hurt solver performance |
| **Model summary/profile** — variable/constraint counts by type, sparsity, density of constraint matrix | Quick overview to sanity-check model size and structure |
| **Coefficient histogram/distribution** — distribution of coefficient magnitudes across the model | Understand overall scaling characteristics at a glance |
| **Duplicate/near-duplicate constraint detection** — find rows of A that are identical or nearly identical | Catch redundant constraints that bloat the model |
| **Empty/trivial constraint detection** — constraints with no variables or that are trivially satisfied | Catch modeling bugs early |
| **Variable bound analysis** — variables with very wide or very tight bounds, or missing bounds | Spot variables that may need tighter bounds for performance |

#### Post-solve (requires a solve)

| Diagnostic | Use case |
|---|---|
| **Constraint activity** — classify constraints as tight, slack, or violated; report slack values | Understand which constraints are binding and which are redundant in the solution |
| **Variables at bounds** — which variables are at their lower/upper bound | Identify bottlenecks in the solution |
| **LP relaxation quality** — solve LP relaxation, report integrality gap and fractionality per integer variable | Assess MIP formulation strength and identify variables with weak relaxation |
| **Dual value analysis** — distribution and extreme values of dual variables | Understand shadow prices and constraint sensitivity |

#### Solver-dependent (requires specific solver backends)

[...]

Issue #561, opened 2026-01-28
Text-based (YAML) math definition interface


brynpickering:
It would be useful to be able to define all the mathematics of a linopy model in human-readable text format. This would help with transparency and shareability (if that's a word) of math, as well as making it easier to define new math without needing to be so familiar with Python.

I have developed something that I think would be suitable [in the Calliope project](https://github.com/calliope-project/calliope/blob/main/src/calliope/math/base.yaml). It is a vectorised syntax that is [documented](https://calliope.readthedocs.io/en/latest/user_defined_math/). The main ideas are:

1. it works with xarray dataarrays, using a [`foreach`](https://calliope.readthedocs.io/en/latest/user_defined_math/syntax/#foreach-lists) directive to define the array dimensions and a [`where`](https://calliope.readthedocs.io/en/latest/user_defined_math/syntax/#where-strings) directive to mask the array so that variables/exprs/constraints can be defined only for a subset.
2. Math expressions do not reference dimensions unless [slicing specific elements in the math](https://calliope.readthedocs.io/en/latest/user_defined_math/syntax/#slicing-data). That is, you could define `p <= p_nom` and it would be applied automatically to all entries in the dimensions defined in `foreach`, or you could define `p[snapshot=2012-01-01] <= p_nom * 0.8` and it would slice `p` to a given snapshot.
3. [helper functions](https://calliope.readthedocs.io/en/latest/user_defined_math/helper_functions/) exist to deal with more complex cases. An obvious use-case is to roll the timeseries dimension forward/backward on one element so that you can get the equivalent of `state_of_charge[t] = state_of_charge[t-1] + p_store[t] - p_discharge[t]` by calling `state_of_charge = roll(state_of_charge, snapshot=1) + p_store - p_discharge`. These are designed as plugins so that power users can add and use new ones if the need arises.
4. It can automatically be rendered as LaTeX math for [use in documentation](https://calliope.readthedocs.io/en/latest/math/base/). This means you only document your math once, when defining it in YAML, and it can then be kept up-to-date in the docs automatically.

It may not be fit for all linopy purposes but the fact that it also relies on multi-dim xarrays I think is already a good starting point. I have been exploring how it could also be used for tidy dataframes (if e.g. updating to using pyoframe) and it seems relatively straightforward; I just need to better understand how to apply the `where` string masking on polars dfs.

[...]

PR #656, opened 2026-04-20
feat: add __weakref__ to Model.__slots__

### LINKED ISSUE #655: Add `__weakref__` to `Model.__slots__`
`linopy.Model` declares `__slots__` but does not include `__weakref__`, which prevents any external registry from tracking `Model` instances.

```python
import linopy, weakref
m = linopy.Model()
weakref.ref(m)
# TypeError: cannot create weak reference to 'Model' object
```

## Use case

Third-party extensions (e.g. accessor-style libraries that add `m.foo.<method>()`) need per-instance storage without polluting `Model`'s own slots or adding `__dict__`. The standard Python pattern is a `WeakKeyDictionary` keyed by the instance, so entries are garbage-collected with the model. That requires the class to be weakref-able.
Im trying to add such in [linopy-yaml](https://github.com/FBumann/linopy-yaml)

## Proposed change

```python
class Model:
    __slots__ = (
        "_variables", "_constraints", ...,
        "__weakref__",   # add
    )
```

## Trade-offs

- Adds one pointer per instance (~8 bytes). No behavioral change.
- No API commitment or new public surface.
- Enables external extension patterns without requiring users to subclass `Model`.

## Compatibility

The change is strictly additive. Existing code that does not touch `weakref` is unaffected. Pickling and `copy`/`deepcopy` are not impacted — `__weakref__` is a runtime-only slot and is not serialized.

One narrow case to flag: any downstream subclass of `linopy.Model` that already declares `__weakref__` in its own `__slots__` (to make that subclass weakref-able) would break at import with:

```
TypeError: __weakref__ slot disallowed: either we already got one, or __itemsize__ != 0
```

The fix on the subclass side is a one-line removal of the redundant `__weakref__` entry. This is unlikely to exist at all i think.

FBumann:
## Summary
- Add `__weakref__` to `Model.__slots__` so `weakref.ref(model)` and `WeakKeyDictionary` keyed by `Model` work.
- Enables third-party accessor-style extensions to attach per-instance state without forcing users to subclass `Model` or reintroduce `__dict__`.

Closes #655

## Test plan
- [x] `test_model_is_weakrefable` — `weakref.ref(Model())` returns a live reference.
- [x] `test_model_weakkeydict_use_case` — entries in a `WeakKeyDictionary[Model, ...]` are GC'd when the model is deleted.
- [x] Full `test/test_model.py` suite still passes (20 tests).

### DIFF
--- a/linopy/model.py
+++ b/linopy/model.py
@@ -231,6 +231,9 @@ class Model:
         "solver_model",
         "solver_name",
         "matrices",
+        # allow weak references to Model instances so third-party extensions
+        # can attach per-instance state via WeakKeyDictionary
+        "__weakref__",
     )
 
     def __init__(
--- a/test/test_model.py
+++ b/test/test_model.py
@@ -6,6 +6,7 @@ Test function defined in the Model class.
 from __future__ import annotations
 
 import copy as pycopy
+import weakref
 from pathlib import Path
 from tempfile import gettempdir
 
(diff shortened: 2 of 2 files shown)


FBumann:
@FabianHofmann I need this to improve the UX of [linopy-yaml](https://github.com/FBumann/linopy-yaml). See #561 . Also probably needed for things like #645.

FabianHofmann (approved):

PR #657, opened 2026-04-20
fix: xarray 2026.4 Dataset-into-Dataset TypeError

FBumann:
## Summary
xarray 2026.4.0 now raises `TypeError: Passing a Dataset as 'data_vars' to the Dataset constructor is not supported` when an existing `Dataset` is passed to `Dataset()`. This broke `linopy.Model()` construction and the RTD docs build (see #656).

Three call sites in `expressions.py` wrapped an already-Dataset return from `.transpose()` / `.assign_coords()` in a redundant `Dataset(...)`. The `Model.parameters` setter also wrapped a potential `Dataset` input — switched to `.copy()` in that branch.

## Changes
- `linopy/expressions.py:287` — drop `xr.Dataset(...)` wrap around `ds.assign_coords(new_coords)`
- `linopy/expressions.py:395` — drop `Dataset(...)` wrap around `data.transpose(..., TERM_DIM)`
- `linopy/expressions.py:2101` — drop `xr.Dataset(...)` wrap around `data.transpose(..., FACTOR_DIM, TERM_DIM)`
- `linopy/model.py:362` — `Model.parameters` setter now uses `.copy()` for Dataset inputs

## Test plan
- [x] `Model()` constructs under xarray 2026.4.0
- [x] Full non-solver test suite passes (1252 passed, 7 skipped) — exercises the previously-broken groupby-with-dataframe path
- [ ] RTD docs build succeeds on the PR preview

### DIFF
--- a/linopy/expressions.py
+++ b/linopy/expressions.py
@@ -284,7 +284,7 @@ class LinearExpressionGroupby:
                 index.names = [str(col) for col in orig_group.columns]
                 index.name = GROUP_DIM
                 new_coords = Coordinates.from_pandas_multiindex(index, GROUP_DIM)
-                ds = xr.Dataset(ds.assign_coords(new_coords))
+                ds = ds.assign_coords(new_coords)
 
             ds = ds.rename({GROUP_DIM: final_group_name})
             return LinearExpression(ds, self.model)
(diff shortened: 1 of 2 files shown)


FBumann:
Duplicate of #647 which is already merged. My local checkout was stale — sorry for the noise.

PR #659, opened 2026-04-20
fix: as_dataarray treating multi-index levels as extra dims

FabianHofmann:

When `as_dataarray` is given a scalar (`np.number`, `int`, `float`, `bool`, `str`, or `list`) together with an `xarray.Coordinates` object whose underlying index is a pandas `MultiIndex`, it inferred dims from `coords.keys()`. That includes the level names, so a scalar with `station` coords containing levels `letter`/`num` was broadcast to shape `(2, 2, 2)` over `('station', 'letter', 'num')` instead of `(2,)` over `('station',)`.

- Use `coords.dims` when `coords` is an `xarray.Coordinates` instance so only real dimensions are used.
- Merge the `np.number` and `int | float | str | bool | list` branches of `as_dataarray` into one, now that both need the same dim resolution.
- Add a regression test in `test_common.py` covering the scalar + multi-index `Coordinates` case.
- Release note added.

## Checklist

### DIFF
--- a/doc/release_notes.rst
+++ b/doc/release_notes.rst
@@ -29,6 +29,7 @@ Upcoming Version
 * Add ``fix()``, ``unfix()``, and ``fixed`` to ``Variable`` and ``Variables`` for fixing variables to values via equality constraints. Supports automatic rounding for integer/binary variables.
 * Add ``relax()``, ``unrelax()``, and ``relaxed`` to ``Variable`` and ``Variables`` for LP relaxation of integer/binary variables. Supports partial relaxation via filtered views (e.g. ``m.variables.integers.relax()``). Semi-continuous variables raise ``NotImplementedError``.
 * Add compatibility to latest xarray version.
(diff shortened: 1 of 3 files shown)


FBumann:
Nice catch — `coords.keys()` was leaking MultiIndex level names as dims; `coords.dims` is the right read.

I added `2c29b87` to harden the test and lock in the behaviour, as im always a bit confused by multiindex:

- Parametrized over arg types (`np.float64`, `int`, `float`, `np.ndarray`) and coord shapes (`xr.Coordinates.from_pandas_multiindex`, plain dict, `DataArray.coords`) — the first coord case exercises the branch where `coords.dims` is a mapping, not a tuple.
- Separate case pinning that explicit `dims=` still wins over inference.

I found the following bug, but declared it out of scope for now: `as_dataarray(pd.Series(...), coords=source.coords)` drops the dim name and returns `dim_0`.

FBumann (approved):

FabianHofmann:
Thanks @FBumann !

PR #672, opened 2026-05-06
feat: add slopes_align to breakpoints()

FabianHofmann:

Adds a new ``slopes_align`` keyword to ``linopy.breakpoints()`` to support an alternative slope-to-breakpoint alignment convention.

- ``slopes_align="pieces"`` (default, unchanged): ``len(slopes) == len(x_points) - 1``; ``slopes[i]`` is the slope between ``x[i]`` and ``x[i+1]``.
- ``slopes_align="leading"``: ``len(slopes) == len(x_points)``; ``slopes[0]`` must be NaN and is dropped, ``slopes[i]`` for ``i>=1`` is the slope between ``x[i-1]`` and ``x[i]``.

The "leading" mode matches the common convention where a marginal value is tabulated alongside each breakpoint with the first row's marginal undefined. It removes the manual ``shift(-1)`` callers (e.g. PyPSA) currently apply before passing slopes to ``breakpoints``.

Validation:
- ``slopes_align`` is rejected outside slopes mode.
- The first slope of each entity must be NaN under ``"leading"``; otherwise a ``ValueError`` is raised.

Per-entity NaN-stripping logic and ragged-input handling are unchanged — the alignment is normalized up front by slicing position 0 along the breakpoint dim.

## Checklist


FBumann:
@FabianHofmann I would probably split breakpoints and slopes into to paths. And introduce a Slopes object, that has a init, alternative constricters via class methods and a to_breakpoints() method. One could use it instead of breakpoints directly. THis would make for better UX. BI planned on doing that in a Follow up

FabianHofmann:
> @FabianHofmann I would probably split breakpoints and slopes into to paths. And introduce a Slopes object, that has a init, alternative constricters via class methods and a to_breakpoints() method. One could use it instead of breakpoints directly. THis would make for better UX. BI planned on doing that in a Follow up

good idea!

FBumann:
WOuld you still do this PR, and then remove the breakpoints() path for Slopes?

FabianHofmann:
let's merge this one, then replace the functionality entirely. fine with that?

PR #677, opened 2026-05-11
docs(piecewise): rewrite reference page + tutorials for correctness and pedagogy

FBumann:
## What was wrong

The piecewise documentation had grown organically and accumulated several precision, structural, and pedagogical problems.

### Correctness bugs

- **Disjunctive auxiliary-variables row** in the comparison table said "Binary + SOS2", but `_add_disjunctive` actually creates continuous `lambda_var`, binary segment selectors, *and* an SOS2 constraint — three kinds, not two.
- **LP domain bound math** was written `x_0 ≤ x ≤ x_n`, implicitly assuming ascending breakpoints. `_lp_eligibility` accepts strict monotonicity in either direction; the math should be `x_min ≤ x ≤ x_max`.
- **Inequality Quick Start example** used `(fuel, [0, 20, 30, 35], "<=")` with the misleading comment "pinned to the curve" on the other tuple. Bounding fuel (a consumption-side variable) with `<=` was the *anti-pattern* the doc's own "Choice of bounded tuple" section warned against.
- **(0, 0) "off" segment** in the disjunctive example used a degenerate single-point segment to encode on/off — exactly what `active=...` (Advanced Features) is for. The example duplicated a pattern that has a proper API and taught a bad habit.
- **Disjunctive table caveat** that "Disjunctive (segments) → always sos2" appeared in the prose before the table, where the apparent contradiction (`.method == "sos2"` but a separate column) was confusing.
- **Implementation-detail leakage**: a "Formulation math" subsection named internal constraint suffixes (`*_link`, `*_output_link`) while a different part of the page said names "are an implementation detail and may evolve".
- **"Pinned" was geometrically misleading**: the doc claimed pinned tuples "lie on the curve", but for 2-tuple inequalities the pinned axis's marginal feasible set is just its breakpoint domain — a single coordinate can't locate a curve point.
- **Tutorial 2 was structured around `fuel ≤ f(power)` with a `maximise -fuel` objective.** The artificial objective hid the fact that bounding fuel from above admits `fuel = 0`, which makes no physical sense.

### Structural problems

- **"Per-tuple sign" was a chapter masquerading as a section** — six concepts under one heading (roles, restrictions, joint geometry across 1/2/3+ tuples, formulation math, sign-choice motivation, deactivation warning). Readers looking up "how do I supply breakpoints" had to scroll past all of it.
- **Two breakpoint sections** — "Breakpoint inputs" under API (overview) and "Breakpoint Construction" as its own section (tutorial). Same three building blocks, overlapping content.
- **`tangent_lines`** was a useful standalone helper buried in a paragraph inside the LP method section.
- **Tutorial 1 §2's method-comparison table** appeared before §3 and §4 had defined the terms it referenced (`sign`, curvature requirements).
- **Tutorial 2's standalone setup plot** showed the same curve that the hypograph/epigraph visualisation later in the notebook would show again with operating points.

### Tutorial-execution noise

- Every `m.solve()` printed the HiGHS banner and the once-per-session `EvolvingAPIWarning`, drowning the actual lesson.

### `doc/piecewise-linear-constraints.rst`

- **Restructured**: moved `Breakpoint Construction` ahead of `Per-tuple sign`, split the latter into four focused subsections, folded the duplicate "Breakpoint inputs" overview into it, promoted `tangent_lines` to its own subsection under LP, and shortened the up-front terminology block.
- **Removed the abstract "Formulation math" subsection**; the load-bearing `active=0` warning moved into `Active parameter`.
- **Fixed factual content**: corrected the Disjunctive aux-vars row, rewrote the LP domain bound as `x_min ≤ x ≤ x_max`, referenced `:ref:`sos-reformulation`` for the Big-M fallback, generalised "Choice of bounded tuple" to both signs, and explained *why* `active` isn't supported with `method="lp"`.
- **Replaced misleading examples**: Quick Start inequality now uses `fuel ≥ f(power)` on the same curve as the equality example; disjunctive snippet replaces the `(0, 0)` on/off hack with a stepped pump and points to `active=...` for on/off.

### `examples/piecewise-linear-constraints.ipynb`

- New intro with an 8-section roadmap; heading renamed to "Creating Piecewise Linear Constraints".
- §2 method comparison trimmed to a forward pointer.
- **§3 rewritten** as two pumps in parallel with two operating bands each (low / high) and a forbidden zone. Demand `[30, 75, 150]` makes every timestep single-pump-infeasible. Output is a flat 4-column DataFrame.
- §4 inequality now uses `fuel ≥ f(power)` on §1's curve.
- New closing "When to use what" table with cross-links.

### `examples/piecewise-inequality-bounds.ipynb`

- Heading renamed to "Creating Piecewise Inequality Bounds".
- Intro rewritten (motivation, then snippet, then a precise tuple-roles table); standalone setup plot removed.
- **Reframed end-to-end around `fuel ≥ f(power)`** with breakpoints aligned to the rest of the doc; hypograph visualisation replaced with the epigraph variant; fallback cell exercises the dual mismatched-curvature cases.
- New summary cross-linking back to the linear-constraints tutorial.

### Both tutorials

- `m.solve(...)` calls now pass `solver_name="highs"`, `reformulate_sos="auto"`, and `output_flag=False`.
- Imports cell silences `EvolvingAPIWarning`.

### Pedagogical additions (beyond the stated bug list)

A few additions go beyond fixing the documented problems and into making the docs a better read.  Calling them out explicitly so they're easy to review or revert:

- **Tutorial 1 intro roadmap** (8-section list).
- **Tutorial 1 "When to use what" closing table** (cheat-sheet for picking the right API pattern).
- **Tutorial 2 "Tuple roles" table** (replaces and expands the old "Key points" table with role + what-it-constrains semantics).
- **Tutorial 2 closing summary** with cross-links back to the linear-constraints tutorial.
- **`tangent_lines` promoted to its own subsection** under LP (discoverability — the function existed before but was buried).

## Test plan

- [ ] `sphinx-build` of `doc/` renders without warnings.
- [ ] CI's "Test documentation notebooks" job passes on both notebooks.
- [ ] `:ref:`sos-reformulation`` link resolves correctly.


FBumann:
@FabianHofmann I would argue that we could try to remove the solver logs from the tutorials in the docs overall...?
WOuld improve readability and reduce noise i think

FabianHofmann:
> @FabianHofmann I would argue that we could try to remove the solver logs from the tutorials in the docs overall...? WOuld improve readability and reduce noise i think

yes, totally

doc/piecewise-linear-constraints.rst
```diff
@@ -288,17 +171,26 @@ over remaining dimensions (e.g. ``time``).
...
     )
 
-The disjunctive formulation is selected automatically when breakpoints have a
-segment dimension.  A bounded tuple (``"<="`` / ``">="``) also works here.
+The disjunctive formulation is selected automatically when breakpoints
+have a segment dimension.  A bounded tuple (``"<="`` / ``">="``) also
+works here.
+
+For a single on/off gate on one continuous curve, prefer ``active=...``
+(see *Advanced Features* below) — using a degenerate ``(0, 0)`` segment
```
FabianHofmann: "advanced features"  should likely be a ref link

doc/piecewise-linear-constraints.rst
```diff
@@ -314,11 +206,115 @@ Link any number of variables through shared breakpoints (joint equality):
...
+  equality.
+- ``"<="`` / ``">="`` — **bounded**: the expression undershoots /
+  overshoots the curve.
+
+**Restrictions (current):**
+
+- At most one tuple may carry a non-equality sign — a single bounded side.
+- With **3 or more** tuples, all signs must be ``"=="``.
+
+Multi-bounded and N≥3-inequality use cases aren't supported yet.  If
```
FabianHofmann: this could go into a info block

doc/piecewise-linear-constraints.rst
```diff
@@ -69,152 +69,35 @@ API
...
 -----------------------
 
+Three building blocks provide breakpoint data:
+
+- :func:`~linopy.breakpoints` — values along a single **connected**
+  curve.
+- :func:`~linopy.segments` — **disjoint** operating regions with gaps
+  between them (e.g. forbidden zones); selects the disjunctive
+  formulation automatically.
+- :class:`~linopy.Slopes` — per-piece slopes plus an initial ``y0``,
```
FabianHofmann: I don't see a more detailed explanation on their usages. it would be helpful to know how they show up in the code. add a paragraph what their individual purpose is (either individually or combined) or ref examples (like the notebook)

FabianHofmann (changes requested):

doc/piecewise-linear-constraints.rst
FBumann: Totally

doc/piecewise-linear-constraints.rst
FBumann: Now links directly to the active parameter

doc/piecewise-linear-constraints.rst
FBumann: I rewrote the section as a whole. now starts with a brief intro thats closely related to our data mdoel (breakpoints or segments), followed by how to pass breakpoints or segments. Slopes is framed as a convenience for breakpoints (which it is. We resolve it to breakpoints)

FBumann:
@FabianHofmann Its ready.

PR #680, opened 2026-05-13
docs: fix broken toctree, refresh API reference, and clean up references

FBumann:
## Summary

A docs cleanup pass covering correctness issues found during a thorough review:

- **Broken toctree** — `index.rst` listed `coordinate-alignment` but no `.rst`/`.nblink` existed; added `doc/coordinate-alignment.nblink` pointing to the existing example notebook.
- **API reference drift** (`api.rst`):
  - Added the five solver classes that were missing despite being advertised: `COPT`, `Knitro`, `MindOpt`, `PIPS`, `cuPDLPx`.
  - Exposed top-level helpers: `align`, `merge`, `options`, `EvolvingAPIWarning`, `PerformanceWarning`.
  - Added recently-introduced Model methods (`add_sos_constraints`, `reformulate_sos_constraints`, `compute_infeasibilities`, `format_infeasibilities`) and Variable methods (`to_linexpr`, `fix`/`unfix`, `relax`/`unrelax`).
  - New sections for `QuadraticExpression`, `Objective`, and `RemoteHandler`.
  - Removed duplicate `Variables.integers` entry; fixed `"under the hook"` → `"under the hood"`.
- **`contributing.rst`** — replaced the stale Black reference with ruff, corrected the `.nblink` example (proper JSON, right path, indentation that was breaking pygments), and used `pre-commit run --all-files`.
- **`benchmark.rst`** — rendered objective read as a product of two variables (nonlinear); fixed to the actual linear benchmark `2x + y` with `x − y ≥ i−1`, verified against `benchmark/scripts/benchmark_linopy.py`.
- **`prerequisites.rst`** — added SCIP, gave MOSEK a description, removed dangling `" -"` after MindOpt, dropped the outdated HiGHS-platforms claim, and clarified what `linopy[solvers]` actually pulls in (verified against `pyproject.toml`).
- **`conf.py` + `index.rst`** — bumped copyright to 2026, fixed `"contnuous"` typo.

Also bundled (already on the branch): silencing HiGHS console output in the tutorial notebooks.

## Test plan

- [x] `sphinx-build -b html doc /tmp/out` runs clean for the changed files (no new warnings introduced).
- [ ] Render the built site locally and spot-check the new API sections (`Top-level helpers`, `QuadraticExpression`, `Objective`, `Remote solving`, and the additional solver pages).
- [ ] Confirm `coordinate-alignment` shows up in the rendered toctree.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[...]

PR #681, opened 2026-05-13
docs: reorganize toctree into basic→advanced sections, rewrite user-guide landing

FBumann:
Stacked on top of #680.

## What this PR does

A second pass on the docs that reorganizes the top-level structure for a basic→advanced flow, tightens the boundary between Getting Started and the User Guide, restructures the API reference, and cleans up a handful of latent rendering issues that were dragging the build.

## Toctree reorganization

`index.rst` had a 16-item flat *User Guide* bag mixing core mechanics, advanced features, troubleshooting, remote solving, migration, and FAQ-style pages. Split into focused sections with a clear reading order:

```
Getting Started → User Guide → Examples → Advanced Features →
Solving → Troubleshooting → Comparisons → Reference
```

- Renamed *Benchmarking* → *Comparisons* (both child pages are comparisons with JuMP / Pyomo, not benchmarks in the regression-tracking sense). `benchmark.rst` H1 became *Performance comparison* so the page title matches the section framing — `syntax.rst` was already titled *Syntax comparison*.
- Renamed *Tutorials* → *Examples* (every notebook in the docs is tutorial-style, the name was overloaded).
- Examples sits immediately under User Guide — reader finishes the core building blocks, then sees an end-to-end worked problem.

## Getting Started → User Guide bridge

Three small but real gaps before this PR:

- The coordinates notebook ended without acknowledging the next step.
- `user-guide.rst` was a one-paragraph stub.
- The User Guide landing's own H1 was *User Guide*, duplicating the sidebar caption.

Fixed by:

- Adding a *Where to next* cell at the end of `create-a-model-with-coordinates` that links forward to the User Guide overview. (Initially tried `:doc:` cross-refs — those only resolve inside RST contexts; switched to markdown links pointing at `.ipynb` files, which nbsphinx rewrites to the rendered `.html` targets.)
- Rewriting `user-guide.rst` as a roadmap that explicitly references the Getting Started content and frames the User Guide as the depth pass on the same surface.
- Renaming the landing H1 to *Overview*.

## API reference restructure (`api.rst`)

Replaced the alphabetic flat dump per class with a layered structure that puts the surface 90% of users reach for at the top, supporting classes in the middle, and genuinely rare-use surface at the bottom.

Top sections are task-oriented (Model methods grouped by what the user is doing):

- *Creating a model*, *Inspecting a model*, *Modifying a model*, *Solving*, *Post-solve access*, *Diagnostics*, *IO*, *Top-level helpers*

*Classes under the hood* keeps the supporting types together (Variable, Variables, LinearExpression, Constraint, Constraints, Objective, Piecewise), each with thematic subgroups (*Attributes / Operations / Conversion / Post-solve* etc.) rather than alphabetic dumps. `autosummary` renders each subgroup as a name + first-line-of-docstring table, so the page becomes scannable.

A new *Advanced* umbrella at the bottom for surface most users don't reach for:

- `QuadraticExpression`, `CSRConstraint`, bulk variable operations (`Variables.fix/unfix/relax/unrelax`), auto-reformulation (`Model.reformulate_sos_constraints`), remote solving (`RemoteHandler`), warning classes.

Curated to drop internal escape hatches and helpers:

- `Model.to_gurobipy / to_highspy / to_mosek / to_cupdlpx` — power-user escape hatches.
- `Model.linexpr` — arithmetic on `Variable` is the natural path.
- `Constraints.format_labels`, `Constraints.coefficientrange` — internal diagnostics.
- `LinearExpression.from_rule`, `LinearExpression.from_constant` — niche constructors.
- `Constraint.freeze / .mutable`, `CSRConstraint.freeze / .mutable` — CSR backend plumbing.
- `ScalarVariable`, `ScalarLinearExpression` — internal types.
- `solvers.PIPS` — stub that raises `NotImplementedError`.

Kept: `Objective` (with the post-solve fields a reader cares about: `value`, `sense`, `expression`, `is_linear`, `is_quadratic`), `EvolvingAPIWarning` / `PerformanceWarning` (called out under *Warnings* as `filterwarnings` targets).

`PiecewiseFormulation` is the return type of `add_piecewise_formulation`, not a construction helper — moved out of *Creating a model* into its own *Piecewise* subsection under *Classes under the hood*, grouped with the lower-level `tangent_lines` helper.

## Source-side fixes that fell out of the restructure

The new api.rst structure exposed 14 properties whose autosummary table cell was blank because they had no docstring. Added a single-line description to each:

- `Model.is_linear` / `is_quadratic` / `type`
- `BaseExpression.vars` / `coeffs` / `const` (fixes both `LinearExpression` and `QuadraticExpression` entries through inheritance)
- `Objective.is_linear` / `is_quadratic`
- `PiecewiseFormulation.method` / `convexity` (inline attribute docstrings on the dataclass fields)
- `OptionSettings` class docstring (so the top-level `options` instance picks up a description)

While in `piecewise.py`, also dropped the duplicated literal value lists from the class-level `Attributes` block — now refers to the `PWL_METHOD` / `PWL_CONVEXITY` type aliases so there's a single source of truth for the allowed values.

To make those references resolve as actual links (rather than rendering as plain `<code>`):

- Added docstrings to `PWL_METHOD`, `PWL_METHODS`, `PWL_CONVEXITY`, `PWL_CONVEXITIES` in `linopy/constants.py`.
- Listed all four in `api.rst` under the *Piecewise* subsection so `autosummary` generates dedicated pages.
- Switched the cross-references in `PiecewiseFormulation.method` / `.convexity` and in `constants.py` to the fully-qualified form (`:data:\`~linopy.constants.PWL_METHOD\``) so they resolve from whichever module they're rendered in.

## Notebook rendering fixes

- Three RST admonitions in markdown cells (`creating-expressions` cells 13 + 17, `creating-variables` cell 41) had a blank line between the directive and a tab-indented body. CommonMark ate the indented block as a code block, so nbsphinx saw an empty directive and Sphinx emitted *"Content block expected for the … directive"* build errors. Fixed by using the convention already working elsewhere in the same notebooks (no blank line, 3-space indent).
- The first cell of `solve-on-remote.ipynb` had literal `\n` sequences in its source instead of real newlines (plus a stray trailing quote). The markdown rendered as one run-on line, Sphinx emitted four *"SSH:nbsphinx-math"* file-not-found and inline-interpreted-text warnings. Cell rewritten with proper line breaks.

## intersphinx

The `sphinx.ext.intersphinx` extension was loaded in `conf.py` but had no `intersphinx_mapping` — cross-refs to `xarray`, `numpy`, `pandas`, etc. were silently unresolved. Added mappings for python, numpy, pandas, xarray, scipy, dask.

## Build

`sphinx-build` runs clean. The remaining warnings are pre-existing docstring formatting issues in `linopy/` source (`CSRConstraint`, `Model.solve`) — separate cleanup.


doc/conf.py
```diff
@@ -18,7 +18,7 @@
 # -- Project information -----------------------------------------------------
 
 project = "linopy"
-copyright = "2021, Fabian Hofmann"
+copyright = "2021-2026, Fabian Hofmann"
 author = "Fabian Hofmann"
```
FabianHofmann, line 22: please also add yourself here, you deserve the credit

FabianHofmann (commented):
I recognized that it would have been easier and cleaner to make a global linopy option `suppress_solver_output` instead of the repeating the `output_flag=False`. 

Another thing:

* A running installation of paramiko on your local machine (`pip install paramiko`)

in the ssh section. we should likely make that a pyproject toml env extension, so prerequisit "remote" is `uv sync --extra remote`

linopy/constants.py
```diff
@@ -60,9 +60,16 @@ class PerformanceWarning(UserWarning):
 PWL_DOMAIN_HI_SUFFIX = "_domain_hi"
 
 PWL_METHOD: TypeAlias = Literal["sos2", "lp", "incremental", "auto"]
+"""Allowed values for the ``method`` argument of :func:`Model.add_piecewise_formulation`."""
```
FabianHofmann: is that a thing? this is a string for developers or can you see that as a user? if not, I would turn that into comments

doc/conf.py
FBumann: Thanks!

FBumann:
> I recognized that it would have been easier and cleaner to make a global linopy option `suppress_solver_output` instead of the repeating the `output_flag=False`.

Possible, but output_flag is the parameter for highs. other solvers use other parameter names.
So might be better to do in a follow up PR for the solver refactor? Its easy to refactir the notebooks accordingly...

doc/conf.py
FBumann: @FabianHofmann copyright or author? Or both?

FBumann:
> * A running installation of paramiko on your local machine (`pip install paramiko`)
> 
> in the ssh section. we should likely make that a pyproject toml env extension, so prerequisit "remote" is `uv sync --extra remote`

Im not really sure if i get what you mean. 
Simply add remote = ["paramiko"] under [project.optional-dependencies] ?
And use it in ci?

linopy/constants.py
FBumann: The docstring shows in the api reference in the docs. As its a type hint used in publix methods, i think it deserves a public docstring.

FabianHofmann:
> > I recognized that it would have been easier and cleaner to make a global linopy option `suppress_solver_output` instead of the repeating the `output_flag=False`.
> 
> Possible, but output_flag is the parameter for highs. other solvers use other parameter names. So might be better to do in a follow up PR for the solver refactor? Its easy to refactir the notebooks accordingly...

ah right. let's leave it as it then.

FabianHofmann:
> > * A running installation of paramiko on your local machine (`pip install paramiko`)
> > 
> > in the ssh section. we should likely make that a pyproject toml env extension, so prerequisit "remote" is `uv sync --extra remote`
> 
> Im not really sure if i get what you mean. Simply add remote = ["paramiko"] under [project.optional-dependencies] ? And use it in ci?

yes, that is what I would suggest (similar to `oetc`) but I don't think you need it in the CI (only in the docs). `pip install ...` calls should be replaced with proper `uv` commands

FBumann:
## Follow up after we merge this into master

Add new solver surface from #682 to api ref, as well as a notebook: #685

doc/conf.py
FabianHofmann: honestly, not sure what the convention is here. probably both?

linopy/constants.py
FabianHofmann: yep

FabianHofmann (approved):

FabianHofmann:
@FBumann feel free to merge

doc/conf.py
FBumann: Ill check it. Separate pr.

FBumann:
> > > * A running installation of paramiko on your local machine (`pip install paramiko`)
> > > 
> > > in the ssh section. we should likely make that a pyproject toml env extension, so prerequisit "remote" is `uv sync --extra remote`
> > 
> > 
> > Im not really sure if i get what you mean. Simply add remote = ["paramiko"] under [project.optional-dependencies] ? And use it in ci?
> 
> yes, that is what I would suggest (similar to `oetc`) but I don't think you need it in the CI (only in the docs). `pip install ...` calls should be replaced with proper `uv` commands

SHould i migrate everything to uv right away?

PR #686, opened 2026-05-18
deps: add `remote` extra for paramiko and migrate install docs to uv

FBumann:
## Summary

Follow-up on the discussion in #681 about the SSH prerequisites.

- Add `remote = ["paramiko"]` to `[project.optional-dependencies]` (mirrors the `oetc` extra convention) so the SSH backend can be installed via `linopy[remote]` instead of telling users to `pip install paramiko` separately.
- Replace `pip install paramiko` in `examples/solve-on-remote.ipynb` with `uv pip install "linopy[remote]"`.
- Migrate the remaining `pip install` snippets in `README.md`, `doc/prerequisites.rst`, `doc/contributing.rst`, and `doc/gpu-acceleration.rst` to their `uv pip install` / `uv sync --extra ...` equivalents, per FabianHofmann's suggestion in #681 that "`pip install ...` calls should be replaced with proper uv commands".

Kept the `conda install -c conda-forge ...` alternatives untouched since those are real alternative install paths, not pip/uv variants.

The `dev` extra is intentionally left as-is in this PR — `paramiko` and `types-paramiko` still ship there so the existing developer workflow does not change. Happy to dedupe (drop `paramiko` from `dev` and rely on `--extra remote`) as a follow-up if preferred.

### Comment

I migrated everything from pip to `uv pip ...`. DO you agree?

### DIFF
--- a/doc/contributing.rst
+++ b/doc/contributing.rst
@@ -16,7 +16,7 @@ Development Setup
 For linting and formatting, we use `ruff <https://docs.astral.sh/ruff/>`_
 and run it via `pre-commit <https://pre-commit.com/index.html>`_:
 
-1. Installation ``conda install -c conda-forge pre-commit`` or ``pip install pre-commit``
+1. Installation ``conda install -c conda-forge pre-commit`` or ``uv pip install pre-commit``
 2. Usage:
     * To automatically activate ``pre-commit`` on every ``git commit``: Run ``pre-commit install``
     * To manually run it: ``pre-commit run --all-files``
@@ -34,7 +34,7 @@ To run the test suite:
 .. code-block:: bash
 
     # Install development dependencies
-    pip install -e .[dev,solvers]
+    uv sync --extra dev --extra solvers
 
     # Run all tests
     pytest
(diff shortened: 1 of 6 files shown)


doc/contributing.rst
```diff
@@ -16,7 +16,7 @@ Development Setup
 For linting and formatting, we use `ruff <https://docs.astral.sh/ruff/>`_
 and run it via `pre-commit <https://pre-commit.com/index.html>`_:
 
-1. Installation ``conda install -c conda-forge pre-commit`` or ``pip install pre-commit``
+1. Installation ``conda install -c conda-forge pre-commit`` or ``uv pip install pre-commit``
```
FabianHofmann: could you add `pre-commit` to the dev extension, then we don't need this step anymore

doc/prerequisites.rst
```diff
@@ -51,15 +51,15 @@ required:
...
+    uv pip install "linopy[solvers]"
 
 
 We recommend installing the HiGHS solver, which is free, open source, and
 fast across a wide range of problem sizes:
 
 .. code:: bash
 
-    pip install highspy
+    uv pip install highspy
```
FabianHofmann: highspy is already in `dev` and `solvers`, so no need for this step anymore

FabianHofmann (changes requested):
wonderful. while we are on it there are two other items we should address in this pr

doc/contributing.rst
FBumann: Added the setup step instead

FabianHofmann (approved):

Issue #688, opened 2026-05-18
Using SOS on masked variables breaks


FBumann:
### Summary

Direct-API SOS construction is broken for both solvers that implement it (Gurobi and Xpress, after #684). Both pass linopy variable labels as if they were solver column positions; the two are only equal when no variable in the model is masked. With a single masked entry anywhere, the SOS set is built against the wrong columns — and the solvers happen to raise rather than silently mis-build, so the bug has been invisible until you try a masked SOS variable.

### Repro

```python
# dev-scripts/repro_sos_minus1_both.py
import pandas as pd
from linopy import Model


def build():
    m = Model()
    coords = pd.Index([0, 1, 2, 3], name="i")
    mask = pd.Series([True, True, False, True], index=coords)  # one slot masked
    sos_var = m.add_variables(
        lower=0, upper=1, coords=[coords], mask=mask, name="sos_var"
    )
    m.add_sos_constraints(sos_var, sos_type=1, sos_dim="i")
    m.add_objective(-sos_var.sum())
    return m


for solver in ("gurobi", "xpress"):
    print(f"\n=== {solver} ===")
    try:
        build().solve(solver_name=solver, io_api="direct")
    except Exception as e:
        print(f"  RAISED {type(e).__name__}: {e}")
```

Output (against the `xpress-direct` branch from #684, on master for Gurobi):

```
=== gurobi ===
  RAISED IndexError: index 3 is out of bounds for axis 0 with size 3

=== xpress ===
  ?404 Error: Invalid column number passed to XPRSaddsets.
  Element 2 of your array has invalid column number 3
  RAISED SolverError
```

### Root cause

In both `_build_solver_model` implementations, the SOS code passes `var.labels` values straight into the solver's `addSOS` as if they were column indices:

- **Gurobi** (`linopy/solvers.py:1530-1532`): `gm.addSOS(sos_type, x[indices].tolist(), weights)` — `x` is `gm.addMVar(M.vlabels.shape, ...)`, indexed by *position*. The code passes linopy labels as positions.
- **Xpress** (#684, `linopy/solvers.py:2168-2170`): `problem.addSOS(indices, weights, type=sos_type)` — `problem` was built via `loadproblem`, where column *position* in the solver matches position in `M.vlabels`. Same mistake.

[...]

PR #684, opened 2026-05-18
feat(xpress): add direct API support via loadproblem

FabianHofmann:
closes #596 

## Changes proposed in this Pull Request

Adds an ``io_api="direct"`` path for the FICO Xpress solver, on par with Gurobi / HiGHS / Mosek / cuPDLPx.

- **``Xpress._build_solver_model``**: builds an ``xpress.problem`` from a linopy ``Model`` via the universal ``loadproblem`` native array API (LP / QP / MIQP in a single call), with SOS constraints attached via ``addSOS``. Compatible with xpress 9.4 and newer (verified the API surface on both 9.4.6 and 9.8.1).
- **``Xpress._build_direct`` / ``_run_direct``**: standard direct-API hooks. ``_run_file`` and ``_run_direct`` now share a new ``_solve`` helper that handles options, log/warmstart/basis/sol I/O, status mapping, and primal/dual extraction (label-indexed for direct, name-indexed when reading back from a file). Mirrors the Gurobi layout.
- **Feature flags**: added ``DIRECT_API`` and ``SOS_CONSTRAINTS`` to ``Xpress.features`` — the parametrized solver test sweep auto-picks them up.
- **``model.to_xpress()``**: new helper matching ``to_gurobipy`` / ``to_highspy`` / ``to_mosek``, exposed on the ``Model`` class.
- **Tests**: ``test_to_xpress`` and ``test_to_xpress_no_names`` in ``test/test_io.py``; the existing ``test_optimization.py`` parametrized sweep now also exercises ``xpress-direct`` (295 xpress tests pass on 9.4.6, 2 skipped).
- **Release notes**: entry under Performance; updated the Internal "file-only solvers" list to drop Xpress.

## Checklist

- [x] Code changes are sufficiently documented; i.e. new functions contain docstrings and further explanations may be given in ``doc``.
- [x] Unit tests for new features were added (if applicable).
- [x] A note for the release notes ``doc/release_notes.rst`` of the upcoming release is included.
- [x] I consent to the release of this PR's code under the MIT license.

PR #689, opened 2026-05-18
fix(sos): refuse masked SOS variables with a clear error; fix `reformulate_sos=True` no-op on native SOS solvers

FBumann:
Stacked on #684. Tracks #688.

## Summary

Two related fixes around SOS handling:

1. **Refuse masked SOS variables up front.** The SOS plumbing (direct-API solver builds + LP-file writer) treats linopy variable labels as solver column indices, which breaks as soon as any variable in the model is masked. Today's symptoms are solver-specific and confusing:
   - Gurobi direct: `IndexError`
   - Xpress direct (after #684): `?404 Invalid column number`
   - LP file: parse errors on Xpress/CPLEX, silent SOS-set corruption on Gurobi (writes phantom `x-1` member)

   Until the proper fix (#688), surface a single clear `NotImplementedError` instead.

2. **Fix `reformulate_sos=True` to actually reformulate on native SOS solvers.** It was silently no-op'd with a warning ("supports SOS natively, ignored"). The 3 options (True, 'auto', False) now each have a distinct use case.

This enables a clean workaround for the masked-SOS bug: pass `reformulate_sos=True` and the SOS is converted to binary+linear before reaching the masked-SOS guard.

## Where the guard lives

In `Solver._build` (covers both `Model.solve` and the 2-step `Solver.from_model(...).solve()` API) and inside `sos_to_file` (covers standalone `m.to_file()`).

## Affected piecewise test

`test_lp_per_entity_nan_padding` exercised both `method='lp'` and `method='sos2'` on NaN-padded breakpoints. The sos2 case relied on the silent corruption — split into:
- `test_lp_per_entity_nan_padding` — keeps the LP-path regression coverage (the original purpose)
- `test_sos2_per_entity_nan_padding_errors` — asserts the new guard fires

Users on piecewise with heterogeneous breakpoints should pick `method='lp'` (or `'incremental'`) until #688.

## Test plan

- [x] `test/test_sos_constraints.py` — 5 new tests (Gurobi/Xpress direct raise, LP writer raise, reformulate_sos=True workaround works, reformulate_sos=True actually reformulates on Gurobi)
- [x] `test/test_sos_reformulation.py` — all 60 existing tests still pass
- [x] `test/test_piecewise_constraints.py` — all 242 tests pass (1 updated, 1 added)
- [x] `pre-commit run --all-files` clean

@FabianHofmann I would like to merge this, creating a baseline for masked sos constraints. Then working on proper support for them with tests etc. SO much of this code will be reverted.


test/test_sos_constraints.py
```diff
@@ -177,6 +177,72 @@ def test_sos2_xpress_direct() -> None:
     assert np.isclose(m.objective.value, 5)
 
 
+def _masked_sos_model(sos_type: int = 1) -> Model:
```
FabianHofmann: let's make this a fixture

test/test_sos_constraints.py
```diff
@@ -177,6 +177,72 @@ def test_sos2_xpress_direct() -> None:
...
+def test_xpress_direct_raises_on_masked_sos() -> None:
+    m = _masked_sos_model()
+    with pytest.raises(NotImplementedError, match="masked"):
+        m.solve(solver_name="xpress", io_api="direct")
+
+
+def test_lp_writer_raises_on_masked_sos(tmp_path: Path) -> None:
+    m = _masked_sos_model()
+    with pytest.raises(NotImplementedError, match="masked"):
+        m.to_file(tmp_path / "sos.lp", io_api="lp")
```
FabianHofmann: pack that into a parameter matrix

FabianHofmann (changes requested):
looks good and the error message makes things quite clear

test/test_sos_constraints.py
FBumann: Done

test/test_sos_constraints.py
FBumann: Done

FBumann:
@FabianHofmann ready for approval

FabianHofmann (approved):

Issue #683, opened 2026-05-16
Refactor OETC as a `Solver` subclass


FabianHofmann:
Follow-up to #682. That PR introduces the stateful `Solver` interface (`Solver.from_name(...).solve() → Result`, `model.apply_result(...)`) but left OETC on the old `remote=OetcHandler(...)` branch in `Model.solve`. This issue tracks folding OETC into the new shape.

## Motivation

- `Model.solve` still has a special-case for `isinstance(remote, OetcHandler)` (linopy/model.py:1626–1644) that bypasses the new `Solver` pipeline.
- OETC users can't say `m.solve("oetc", ...)`; they have to construct an `OetcHandler` and pass it via `remote=`.
- The returned solution is patched in field-by-field rather than going through `apply_result` / label-indexed `Solution`.
- Aligning OETC with the `Solver` interface sets up the async-job seam coroa flagged in #682 (Gurobi batch, OETC).

## Proposed design (TBD)

Add `class Oetc(Solver[OetcSettings])` in `linopy/solvers.py`. Treat netcdf-over-GCP as `io_api="direct"` — the "native model" is the linopy `Model` shipped as netcdf; don't introduce a new `IO_APIS` entry.

**Dataclass fields** (on top of inherited `model`/`io_api`/`options`):
- `settings: OetcSettings | None = None` (resolve via `OetcSettings.from_env()` if absent)
- Runtime: `_handler: OetcHandler | None`, `_job_uuid: str | None`, `_solved_model: Model | None`

**Lifecycle:**
- `_build_direct(...)` — instantiate `OetcHandler(self.settings)`, serialize `self.model` to a temp netcdf, upload to GCP, cache `_vlabels`/`_clabels` via `_cache_model_labels(self.model)`. **Do not submit the job here.**
- `_run_direct(...)` — submit, poll, download, `read_netcdf` into `self._solved_model`, assemble `Solution`/`Status`/`SolverReport`, return via `self._make_result(...)`.
- `_run_file` — not implemented.

**Solution assembly.** The round-tripped model shares labels with the source. Build dense label-indexed arrays of size `_n_vars` / `_n_cons` by iterating local `self.model.variables` / `constraints`, reading `labels` per name, and indexing `primal[labels.ravel()] = solved.variables[name].solution.values.ravel()` (and same for duals). Objective from `solved.objective.value`. Missing entries stay `NaN`.

[...]

PR #690, opened 2026-05-18
refactor(sos): add Model.apply/undo_sos_reformulation methods

FBumann:
## Summary

**On top of the following this PR includes changes from #691** 

Stateful pair of methods on `Model` that own the SOS reformulation lifecycle. `Model.solve(reformulate_sos=...)` delegates to them; same external behavior.

- **`Model.apply_sos_reformulation()`** — stashes the token on `_sos_reformulation_state`. Raises `RuntimeError` if already applied.
- **`Model.undo_sos_reformulation()`** — restores the original SOS form. No-op if nothing is applied.
- **`Solver._build()`** — raises `ValueError` when an SOS-bearing model meets a solver without `SolverFeature.SOS_CONSTRAINTS`, so the low-level `Solver.from_model(...).solve()` path is safe without going through `Model.solve()`.

Persistence:
- `Model.copy()` / `copy.copy` / `copy.deepcopy` deep-copy the token; the copy is independently undoable.
- `to_netcdf()` raises if reformulation is active — undo first.

## Why the lifecycle lives on `Model`, not on `Solver`

I considered three alternatives and rejected all of them:

**On `Solver.from_model()` with undo on `solver.close()`.** Tempting because Solver already manages env lifecycle. Rejected: reformulation is a *model* transformation, and the Solver only queries a feature flag — putting the lifecycle there makes Solver responsible for work it doesn't perform.

**Build-time reformulation inside `_build_*`, no model mutation.** Cleanest in theory. Rejected because the mutation is a feature: users can print the reformulated MILP, export it to LP/MPS, or write additional constraints referencing the aux binaries. Hiding it inside the build step strips all of that.

**Standalone context manager.** Rejected: it only handles the transient (auto-undo) case. The method pair supports transient, permanent, and decide-later workflows; a 3-line context manager over the methods can be added later if wanted.

What's left on `Solver`: one feature-flag check in `_build()`. Layering:

| Layer | Role |
|---|---|
| `sos_reformulation.py` | implementation primitives |
| `Model` | owns transformation state — apply / undo / inspect |
| `Solver` | declares feature support; raises on incompatible state |
| `Model.solve()` | bracketing for the one-shot ergonomic |

## Context

Same thread as #688. While reviewing the two-step `Solver.from_model().solve()` API from #682, I noticed `Model.solve()` was doing orchestration the Solver path couldn't reproduce. This PR addresses the SOS slice; validation / sanitization lifting can follow separately.

## Test plan

- [x] `pytest test/test_sos_reformulation.py` — 69 tests, 9 new (apply/undo primitives, copy across all three protocols, netcdf guard, Solver-path raise).
- [x] `pytest test/test_sos_constraints.py test/test_optimization.py test/test_io.py` — 1544 pass.
- [x] `ruff check`, `ruff format`, `mypy linopy/{model,solvers,io}.py` clean.

### DIFF
--- a/linopy/io.py
+++ b/linopy/io.py
@@ -857,8 +857,7 @@ def to_netcdf(m: Model, *args: Any, **kwargs: Any) -> None:
         raise RuntimeError(
             "Cannot serialize a model with an active SOS reformulation. "
             "Call `model.undo_sos_reformulation()` first to restore the "
-            "original SOS form, or save the model in its reformulated form "
-            "after explicitly clearing `model._sos_reformulation_state`."
+            "original SOS form before saving."
         )
 
     def with_prefix(ds: xr.Dataset, prefix: str) -> xr.Dataset:
--- a/linopy/io.py
+++ b/linopy/io.py
@@ -10,6 +10,7 @@ import json
 import logging
 import shutil
 import time
+import warnings
 from collections.abc import Callable, Iterable
 from io import BufferedWriter
 from pathlib import Path
@@ -847,17 +848,27 @@ def to_netcdf(m: Model, *args: Any, **kwargs: Any) -> None:
     **kwargs : TYPE
         Keyword arguments passed to ``xarray.Dataset.to_netcdf``.
 
-    Raises
-    ------
-    RuntimeError
-        If the model has an active SOS reformulation. Call
-        :meth:`Model.undo_sos_reformulation` before serializing.
+    Notes
+    -----
+    The SOS reformulation lifecycle token lives only on the in-memory
+    Model and is not persisted. If the model has an active SOS
+    reformulation at serialization time, the netcdf contains the
+    reformulated MILP form (aux binaries and cardinality constraints)
+    and a :class:`UserWarning` is emitted to flag that the deserialized
+    copy will not be able to undo the reformulation.
+
+    ``Model.solve(remote=...)`` invokes ``to_netcdf`` internally on the
+    reformulated model and suppresses this warning.
     """
     if m._sos_reformulation_state is not None:
-        raise RuntimeError(
-            "Cannot serialize a model with an active SOS reformulation. "
-            "Call `model.undo_sos_reformulation()` first to restore the "
-            "original SOS form before saving."
+        warnings.warn(
+            "Serializing a model with an active SOS reformulation. The "
+            "netcdf will contain the reformulated MILP form; the "
+            "reformulation lifecycle token is not persisted, so a "
+            "reader cannot undo it. Call `model.undo_sos_reformulation()` "
+            "first if you want the original SOS form on disk.",
+            UserWarning,
+            stacklevel=2,
         )
 
     def with_prefix(ds: xr.Dataset, prefix: str) -> xr.Dataset:
@@ -929,6 +940,13 @@ def read_netcdf(path: Path | str, **kwargs: Any) -> Model:
     Returns
     -------
     m : linopy.Model
+
+    Notes
+    -----
+    The SOS reformulation lifecycle token is not persisted by
+    :func:`to_netcdf`. If the saved model was in reformulated form,
+    the deserialized Model is too, but
+    :meth:`Model.undo_sos_reformulation` is a no-op on it.
     """
     from linopy.constraints import (
         Constraint,
--- a/linopy/io.py
+++ b/linopy/io.py
@@ -5,6 +5,7 @@ Module containing all import/export functionalities.
 
 from __future__ import annotations
 
+import copy as _copy
 import json
 import logging
 import shutil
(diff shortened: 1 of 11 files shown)


FBumann:
@FabianHofmann This is probably the heavy lifting of #688

FBumann:
@FabianHofmann Look at the reasoning in #691, where the guideline "Mutating the Model is done deliberately before using a Solver" also affects sanitization etc

FabianHofmann:
yes, makes sense too. should we merge into #691 since they both go in the same direction? would make it easier for reviewing I'd say

FBumann:
@FabianHofmann #691 is stacked on top anyway. So we can just clos this and point the other to master. But i thought this deserves its own deliberate PR with description and reasoning etc.
But feel free to do the review in #691.
Id like to merge them sequentially into master though.

FabianHofmann:
> @FabianHofmann #691 is stacked on top anyway. So we can just clos this and point the other to master. But i thought this deserves its own deliberate PR with description and reasoning etc. But feel free to do the review in #691. Id like to merge them sequentially into master though.

was confused with the references. that makes sense. I am reviewing #691 quickly and if you don't mind we merge #691 to here and I have another look at the combined diff. would that be fine?

FBumann:
I would rather merge them both into master sequentially, so we keep the PR's in master. But Its not really important to be honest. Do as you please

FabianHofmann:
### Code review

Found 1 issue:

1. `Model.solve(reformulate_sos=True|"auto")` leaks `_sos_reformulation_state` if `Solver.from_name()` or `solver.solve()` raises. The undo lives in a `finally` around `assign_result()` (lines 1830–1832), but `from_name` and `solve` are wrapped by a separate `try/finally` whose only job is file cleanup (lines 1788–1825). On any build/solve failure the second `try` is never reached, the model stays in reformulated form, and the next `Model.solve()` call hits `RuntimeError: SOS reformulation has already been applied` from `apply_sos_reformulation` (lines 1243–1245). Suggest a single `try`/`finally` wrapping both blocks so the auto-undo contract documented on `apply_sos_reformulation` actually holds for the transient case.

https://github.com/PyPSA/linopy/blob/68d75cd8fee240d85443d8165b366536a3b78f16/linopy/model.py#L1786-L1833

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

linopy/io.py
FBumann, line 1: I thought someone might need it to use sos reformulation with remote/oetc. Doesnt this go through netcdf?

FabianHofmann: that is totally true. so would be nice to support it

FabianHofmann: @FBumann should I quickly do that?

FBumann: Ill do it

FabianHofmann: after thinking about it, I would actually like to have that as a follow up where we touch more on logic how to store sos attributes. so this should not be a blocker for now

FBumann: My plan: handle it in solve(), reataching it after remote finishes. Remote doesnt see it. Warn on to_netcdf() when sosreformulationstate is atached.

FabianHofmann: I see, feels a bit hacky though, as we are effectively changing the model on the way to remote. I would rather prefer supporting the attributes in netcdf IO, but this needs another pr I'd say

FabianHofmann: ah i came too late

FBumann: For me it doesnt feel hacky. We treat oetc/remote like we do any solver: **Mutate the Model**,  the solver gets only what he needs. This perfectky trasitions to #683, with OETC behaving like a regular solver.
IO Is now not needed, but can be added anyway (dropping data on IO is never the best option)

FBumann: @FabianHofmann If you disagree, we can discuss it. Im not using oetc...

FabianHofmann: fine, let's me add a commit to streamline the code in solve now. many lines are they only for the purpose of reformulation awareness. I'll draft something and likely move the reformulation hanlding into the functions solve_on_* with a context manager

FBumann: I would maybe defer that after we close #683 ? Then we cleanly refactor? But we can also do it twice, i dont mind

FabianHofmann: #683 is way more invasive than the small refactor above, have it ready. just reviewed locally and pushing it here. improves the readability of solve significantly

FabianHofmann (approved):

FabianHofmann:
@FBumann ready to go. if you want take a look at my latest commit. if you agree, feel free to merge

FBumann:
@FabianHofmann Looks good.

PR #685, opened 2026-05-18
docs: add working-with-solvers notebook for the stateful Solver API

FBumann:
## Summary
- Adds a `working-with-solvers` notebook that walks through the stateful `Solver` instance and the two-step solve API introduced in #682.
- Wires the notebook into the docs via `doc/working-with-solvers.nblink` and `doc/index.rst`.
- Updates `examples/create-a-model.ipynb` so the basics intro lines up with the new solver flow.
- Drops the hardcoded `SolverReport` coverage list from the notebook (auto-derived now), strips stray outputs, trims redundant cells, and silences HiGHS logs in the example runs.
- Adds a **Model Transformations** section covering SOS reformulation and constraint sanitization, with the explicit layering from #690: transformations live on `Model`, `Model.solve(...)` brackets them for the one-shot path, and the two-step `Solver` API requires the caller to invoke (and undo) them themselves. The example uses a `try/finally` around `apply_sos_reformulation` / `undo_sos_reformulation`.
- Merges `master` into the branch (#684 Xpress direct API, #686 `remote` extra + uv install docs, #689/#690 SOS work, #698 remote tempfile fix). Conflicts resolved in `doc/index.rst`, `doc/api.rst`, and `examples/solve-on-remote.ipynb`.

## Follow-up
A small public context-manager wrapper over the SOS apply/undo pair (e.g. `Model.sos_reformulated()`, or a module-level `linopy.sos_reformulation.reformulated_sos(model)`) would let the two-step example collapse the `try/finally` into one line. PR #690 already left the door open for this — punted to a separate PR so this one stays a docs change.

## Test plan
- [ ] Build the docs locally and confirm the new notebook renders end-to-end without execution errors
- [ ] Verify nav/toctree entry for "Working with solvers" appears in the rendered site
- [ ] Re-run the notebook to confirm cells execute cleanly with the bundled solver (HiGHS)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #691, opened 2026-05-18
refactor(solver): validation, sanitize kwargs, and result wiring on Solver path

FBumann:
**Stacked on #690.** Closes the gap between `Model.solve()` and `Solver.from_name(...).solve()`.

## Core principle

`Solver.from_model()` does **not** mutate the model. Model-state transformations are Model-level primitives, called explicitly:

```python
model.constraints.sanitize_zeros()        # drop zero-coefficient terms
model.constraints.sanitize_infinities()   # drop trivial <= +inf / >= -inf rows
model.apply_sos_reformulation()           # rewrite SOS as binary + linear (#690)
model.undo_sos_reformulation()            # restore the original SOS form (#690)
```

`Model.solve()` calls these internally as orchestration. The low-level Solver path expects the caller to apply them — no silent mutation on build.

Canonical low-level pattern:

```python
model.constraints.sanitize_zeros()   # optional preprocessing
solver = Solver.from_name("highs", model, io_api="direct")
result = solver.solve()
model.assign_result(result, solver=solver)
```

## What lives where after this PR

`Port?` = should it move to `Solver`. `Status` = ✅ done · ⬜ todo · — n/a.

| Item | Port? | Status | Where (on `Solver` or why not) |
|---|---|---|---|
| Quadratic / semi-continuous feature gates | yes | ✅ | `Solver._validate_model()` (build-time, called from `_build()`) |
| SOS feature gate | yes | ✅ (#690) | `Solver._validate_model()` (build-time) |
| Empty-objective check | yes | ✅ | `Solver.solve()` (run-time); scoped dup in `Model.solve()`'s remote branch, removed by #683 |
| Result wiring (`model.solver` + primal/dual/status) | yes | ✅ | `Model.assign_result(result, solver=...)` (post-solve) |
| Build-time kwargs: `io_api`, `explicit_coordinate_names`, `set_names`, `problem_fn`, `slice_size`, `progress`, `**solver_options` | yes | ✅ | `Solver.from_model(...)` |
| Run-time kwargs: `basis_fn`, `warmstart_fn` | yes | ✅ | `Solver.solve(...)` |
| `env`, `log_fn` | yes | ⬜ | accepted on **both** `from_model()` and `solve()` today; PR3 consolidates to `from_model()` only (13 subclasses) |
| `sanitize_zeros` / `sanitize_infinities` | no | — | Model-level: `model.constraints.sanitize_*()`. Mutates — caller-owned. |
| SOS apply/undo lifecycle | no | — | Model-level: `model.apply_sos_reformulation()` / `undo_sos_reformulation()` (#690). Mutates — caller-owned. |
| `reformulate_sos` (`True`/`False`/`"auto"`) | no | — | Model.solve()-side convenience wrapping `model.apply_sos_reformulation()`. Solver-path users call apply/undo directly. |
| OETC remote (`remote=OetcHandler(...)`) | yes | ⬜ (#683) | refactor into a `Solver` subclass |
| Generic `remote=RemoteHandler(...)` | ? | — | could follow #683 |
| `mock_solve=True` | ? | — | could become a `Mock` `Solver` subclass |
| "Any solver installed" raise | ? | — | `Solver.__post_init__` covers "this-installed"; "any-installed" is module-level |
| `solver_name` auto-pick, default `solution_fn`, `keep_files` cleanup, info logs, `reset_solution()` | no | — | orchestration: belongs above the Solver layer |

> **Docs follow-up.** Keeping sanitize off the Solver kwargs makes it less discoverable for users coming from `Model.solve()`. The user guide / example notebooks should surface the model-level primitives — separate doc PR #685, not in scope here.

### DIFF
--- a/linopy/solvers.py
+++ b/linopy/solvers.py
@@ -507,18 +507,44 @@ class Solver(ABC, Generic[EnvType]):
         """Dispatch to direct or file build based on ``io_api``."""
         if self.model is None:
             raise RuntimeError("Solver has no model attached; cannot build.")
-        if self.model.variables.sos and not type(self).supports(
-            SolverFeature.SOS_CONSTRAINTS
-        ):
-            raise ValueError(
-                f"Solver {self.solver_name.value} does not support SOS constraints. "
-                "Call `model.apply_sos_reformulation()` first, or use a solver that "
-                "supports SOS."
-            )
+        sanitize_zeros = build_kwargs.pop("sanitize_zeros", True)
+        sanitize_infinities = build_kwargs.pop("sanitize_infinities", True)
+        self._validate_model()
+        if sanitize_zeros:
+            self.model.constraints.sanitize_zeros()
+        if sanitize_infinities:
+            self.model.constraints.sanitize_infinities()
         if self.io_api == "direct":
             self._build_direct(**build_kwargs)
         else:
             self._build_file(**build_kwargs)
+
+    def _validate_model(self) -> None:
+        """Pre-build checks on whether this solver can handle ``self.model``."""
+        model = self.model
+        assert model is not None
+        solver_name = self.solver_name.value
+        cls = type(self)
+
+        if model.is_quadratic and not cls.supports(SolverFeature.QUADRATIC_OBJECTIVE):
+            raise ValueError(
+                f"Solver {solver_name} does not support quadratic problems."
+            )
+
+        if model.variables.semi_continuous and not cls.supports(
+            SolverFeature.SEMI_CONTINUOUS_VARIABLES
+        ):
+            raise ValueError(
+                f"Solver {solver_name} does not support semi-continuous variables. "
+                "Use a solver that supports them (gurobi, cplex, highs)."
+            )
+
+        if model.variables.sos and not cls.supports(SolverFeature.SOS_CONSTRAINTS):
+            raise ValueError(
+                f"Solver {solver_name} does not support SOS constraints. "
+                "Call `model.apply_sos_reformulation()` first, or use a solver that "
+                "supports SOS."
+            )
 
     def _build_direct(self, **build_kwargs: Any) -> None:
         """Build the native solver model from ``self.model``. Override per-solver."""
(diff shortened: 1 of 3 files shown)


linopy/solvers.py
```diff
@@ -560,7 +589,30 @@ def _build_file(self, **build_kwargs: Any) -> None:
...
+
+        The canonical low-level pattern is::
+
+            solver = Solver.from_name("gurobi", model, io_api="direct")
+            result = solver.solve()
+            model.assign_result(result, solver=solver)
+
+        Passing ``solver=`` to :meth:`Model.assign_result` wires
+        ``model.solver`` so post-solve helpers like
+        :meth:`Model.compute_infeasibilities` keep working.
```
FabianHofmann, line 604: very good catch!

test/test_solvers.py
```diff
@@ -464,3 +464,112 @@ def test_xpress_gpu_feature_reflects_installed_version() -> None:
...
         SolverFeature.GPU_ACCELERATION
     ) == _installed_version_in("xpress", ">=9.8.0")
+
+
+class TestValidateModelOnBuild:
+    """Solver._build() runs solver-feature checks regardless of entry point."""
+
+    @pytest.mark.skipif(
+        "highs" not in solvers.available_solvers, reason="HiGHS not installed"
+    )
```
FabianHofmann: we definitely need a helper function that makes these skips snappy

linopy/solvers.py
```diff
@@ -560,7 +589,30 @@ def _build_file(self, **build_kwargs: Any) -> None:
...
+        Raises
+        ------
+        ValueError
+            If the attached model has no objective set. Submit-time check
+            shared by both ``Model.solve()`` and direct-Solver callers.
+        """
+        if self.model is not None and self.model.objective.expression.empty:
+            raise ValueError(
+                "No objective has been set on the model. Use `m.add_objective(...)` "
+                "first (e.g. `m.add_objective(0 * x)` for a pure feasibility problem)."
```
FabianHofmann, line 615: that should covered by the check on model level in `Model.solve`. why also have it here?

FabianHofmann (commented):
this is extremely sensible! I did not think about model mutations in the solver refactor pr. thanks for addressing this. let me know when this is ready to review

linopy/solvers.py
FBumann: I migrated it here. Its gone from Model.solve(). (Actually, currently its still repeated for remote solving, but this will go eventually as soon as OETC is a Solver instance)

FBumann:
> this is extremely sensible! I did not think about model mutations in the solver refactor pr. thanks for addressing this. let me know when this is ready to review

It is ready, but i think its less about correctness, but more about the conceptual split between Model and Solver, and the ergonomics for users using the new Solver class

linopy/solvers.py
FabianHofmann: gotcha, forgot about the condition in Model.solve

PR #692, opened 2026-05-18
fix(sos): SOS constraints on masked variables (#688)

### LINKED ISSUE #688: Using SOS on masked variables breaks

FBumann:
**Stacked on #690.** Fixes #688.

Closes the SOS-on-masked-variables bug at its root rather than guarding against it. The defensive `_check_sos_unmasked` workaround introduced by #689 is removed; the real fix covers both the direct-API path (Gurobi, Xpress) and the LP file writer.

## Root cause

Two related failure modes, both rooted in conflating linopy *labels* with solver-side *positions / names*:

1. **Direct API**: `gm.addSOS(...)` / `problem.addSOS(...)` were given linopy labels as if they were 0-based column positions in the active-variable array. Labels equal positions only when no variable in the model is masked anywhere.
2. **LP writer**: `sos_to_file` emitted variable names directly from the raw label array. A masked entry (label = -1) became `x-1`, which LP parsers reject or (gurobi LP reader) silently corrupt. The set identifier `labels.isel(sos_dim=0)` was -1 whenever the first slot was masked.

## Fix

- **Direct API** (Gurobi, Xpress): shared `_sos_set_positions(labels, weights, label_to_pos)` helper drops masked entries and resolves linopy labels to active-variable positions via `model.variables.label_index.label_to_pos`. Both direct builders route through it; empty sets are skipped.
- **LP writer** (`sos_to_file`): polars pipeline gets a `labels != -1` filter before `group_by`; per-set id is the `max` of non-masked labels (always a valid label; fully-masked sets get id -1 and are filtered out). Keeps polars vectorization.

The two paths stay separate by design — LP polars vs direct numpy is the right idiom split for each path's throughput needs; unification would force LP to Python-level iteration with no maintenance win for a ~10-line shared concept.

## Coverage

- **`test/test_sos_masked.py`** (new) — 11 fixture configs (1D/2D × all mask placements × with/without unrelated-var mask) × 2 SOS types × every SOS-capable solver × {direct, lp} io_apis ≈ 110 cases. Three-layer oracle per test: `status == "ok"` + `objective == expected` + element-wise `solution == expected`. Asymmetric objective coefficients break permutation symmetry; analytical optimizer in the fixture provides exact oracles using list-position adjacency for SOS2.
- Companion sos2 piecewise NaN-padding test converted to a positive assertion (same `y_b = 12.5` oracle as the lp variant).
- Existing reformulation-with-masked test moved to `test_sos_reformulation.py`, reworded — reformulation isn't a "workaround" anymore.

## TDD discipline

5 commits, each fix preceded by its failing test:

1. `test(sos): regression matrix for masked SOS variables` — drops defensive guard so tests surface the real bug. **Without fix:** 28 LP-path masked cases fail with `Name 'x-1' does not exist`.
2. `fix(sos): resolve linopy labels to solver column positions in direct builds` — direct-path passes.
3. `fix(sos): filter masked SOS members + robust set id in LP writer` — LP-path passes.
4. `test(pwl): convert masked-sos2 NaN-padding test to positive assertion`.
5. `test(sos): move masked-reformulation test to test_sos_reformulation`.

## Test plan

- [x] `pytest test/test_sos_masked.py` — 66 active cases pass, 44 skipped.
- [x] Full suite passes.
- [x] `ruff check`, `ruff format`, `mypy linopy/{model,solvers,io}.py` clean.

### DIFF
--- a/linopy/io.py
+++ b/linopy/io.py
@@ -424,8 +424,8 @@ def sos_to_file(
 
     for name in names:
         var = m.variables[name]
-        sos_type = var.attrs[SOS_TYPE_ATTR]
-        sos_dim = var.attrs[SOS_DIM_ATTR]
+        sos_type = int(var.attrs[SOS_TYPE_ATTR])  # type: ignore[call-overload]
+        sos_dim = str(var.attrs[SOS_DIM_ATTR])
 
         other_dims = [dim for dim in var.labels.dims if dim != sos_dim]
         for var_slice in var.iterate_slices(slice_size, other_dims):
--- a/linopy/io.py
+++ b/linopy/io.py
@@ -593,8 +593,6 @@ def to_file(
     """
     Write out a model to a lp or mps file.
     """
-    m._check_sos_unmasked()
-
     if fn is None:
         fn = Path(m.get_problem_file())
     if isinstance(fn, str):
--- a/linopy/io.py
+++ b/linopy/io.py
@@ -430,9 +430,19 @@ def sos_to_file(
         other_dims = [dim for dim in var.labels.dims if dim != sos_dim]
         for var_slice in var.iterate_slices(slice_size, other_dims):
             ds = var_slice.labels.to_dataset()
-            ds["sos_labels"] = ds["labels"].isel({sos_dim: 0})
+            # Per-set id: max of labels along the SOS dim. Real labels are
+            # non-negative and globally unique, so max yields a valid,
+            # unique-per-set id whenever the set has any unmasked slot.
+            # Fully-masked sets get id -1 and are filtered out below.
+            ds["sos_labels"] = ds["labels"].max(sos_dim)
             ds["weights"] = ds.coords[sos_dim]
             df = to_polars(ds)
+
+            # Drop masked member rows so the LP file never emits `x-1`, and
+            # drop any rows belonging to a fully-masked set (sos_labels == -1).
+            df = df.filter((pl.col("labels") != -1) & (pl.col("sos_labels") != -1))
+            if df.is_empty():
+                continue
 
             df = df.group_by("sos_labels").agg(
                 pl.concat_str(
--- a/linopy/io.py
+++ b/linopy/io.py
@@ -430,16 +430,13 @@ def sos_to_file(
         other_dims = [dim for dim in var.labels.dims if dim != sos_dim]
         for var_slice in var.iterate_slices(slice_size, other_dims):
             ds = var_slice.labels.to_dataset()
-            # Per-set id: max of labels along the SOS dim. Real labels are
-            # non-negative and globally unique, so max yields a valid,
-            # unique-per-set id whenever the set has any unmasked slot.
-            # Fully-masked sets get id -1 and are filtered out below.
+            # Per-set id = max member label: unique per set (labels are globally
+            # unique); a fully-masked set reduces to -1 and is dropped below.
             ds["sos_labels"] = ds["labels"].max(sos_dim)
             ds["weights"] = ds.coords[sos_dim]
             df = to_polars(ds)
 
-            # Drop masked member rows so the LP file never emits `x-1`, and
-            # drop any rows belonging to a fully-masked set (sos_labels == -1).
+            # Drop masked members
             df = df.filter((pl.col("labels") != -1) & (pl.col("sos_labels") != -1))
             if df.is_empty():
                 continue
--- a/linopy/solvers.py
+++ b/linopy/solvers.py
@@ -106,6 +106,41 @@ def _solution_from_labels(
     return values_to_lookup_array(np.asarray(values, dtype=float), labels, size=size)
 
 
+def _sos_set_positions(
+    labels: np.ndarray, weights: np.ndarray, label_to_pos: np.ndarray
+) -> tuple[list[int], list[float]]:
+    """
+    Convert a SOS set's linopy labels to solver column positions.
+
+    Direct-API solvers (gurobi, xpress) accept SOS members as 0-based column
+    positions in the solver's variable array, which corresponds to the active
+    (non-masked) variable order — i.e., the order of
+    ``model.variables.label_index.vlabels``. Masked entries (label ``-1``) are
+    dropped along with their weights.
+
+    Parameters
+    ----------
+    labels : np.ndarray
+        Flat array of linopy labels for the SOS members.
+    weights : np.ndarray
+        Matching weights; same length as ``labels``.
+    label_to_pos : np.ndarray
+        ``model.variables.label_index.label_to_pos`` — lookup of label →
+        active-variable position.
+
+    Returns
+    -------
+    tuple[list[int], list[float]]
+        Solver column positions and matching weights, with masked entries
+        removed.
+    """
+    mask = labels != -1
+    return (
+        label_to_pos[labels[mask]].tolist(),
+        weights[mask].tolist(),
+    )
+
+
 class SolverFeature(Enum):
     """Enumeration of all solver capabilities tracked by linopy."""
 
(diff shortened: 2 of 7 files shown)


FabianHofmann:
@FBumann would you mind updating this, I would then take another look

FBumann:
@FabianHofmann Ready. Sorry, took a minute

linopy/io.py
```diff
@@ -424,16 +424,26 @@ def sos_to_file(
...
+        sos_dim = str(var.attrs[SOS_DIM_ATTR])
 
         other_dims = [dim for dim in var.labels.dims if dim != sos_dim]
         for var_slice in var.iterate_slices(slice_size, other_dims):
             ds = var_slice.labels.to_dataset()
-            ds["sos_labels"] = ds["labels"].isel({sos_dim: 0})
+            # Per-set id: max of labels along the SOS dim. Real labels are
+            # non-negative and globally unique, so max yields a valid,
+            # unique-per-set id whenever the set has any unmasked slot.
+            # Fully-masked sets get id -1 and are filtered out below.
```
FabianHofmann: make that comment more concise

linopy/io.py
```diff
@@ -424,16 +424,26 @@ def sos_to_file(
...
+            ds["sos_labels"] = ds["labels"].max(sos_dim)
             ds["weights"] = ds.coords[sos_dim]
             df = to_polars(ds)
 
+            # Drop masked member rows so the LP file never emits `x-1`, and
+            # drop any rows belonging to a fully-masked set (sos_labels == -1).
```
FabianHofmann: remove that

linopy/solvers.py
```diff
@@ -106,6 +106,75 @@ def _solution_from_labels(
     return values_to_lookup_array(np.asarray(values, dtype=float), labels, size=size)
 
 
+def _sos_set_positions(
```
FabianHofmann: this function only used once and has to lines of code, drop it and directly use the code.

linopy/solvers.py
```diff
@@ -106,6 +106,75 @@ def _solution_from_labels(
...
+
+    Returns
+    -------
+    tuple[list[int], list[float]]
+        Solver column positions and matching weights, with masked entries
+        removed.
+    """
+    mask = labels != -1
+    return (
+        label_to_pos[labels[mask]].tolist(),
```
FabianHofmann: this `to_list()` looks suspcisous. can we stay in the performent numpy world?

linopy/solvers.py
```diff
@@ -106,6 +106,75 @@ def _solution_from_labels(
...
+            stacked = var.labels.stack(_sos_group=others)
+            sets = (s.unstack("_sos_group") for _, s in stacked.groupby("_sos_group"))
+
+        for s in sets:
+            s = s.squeeze()
+            labels = s.values.flatten()
+            weights = s.coords[sos_dim].values
+            positions, kept_weights = _sos_set_positions(labels, weights, label_to_pos)
+            if positions:
+                yield sos_type, positions, kept_weights
```
FabianHofmann: I don't think this is performant with the xarray stack, groupby, and unstack. we could potentially go to numpy only:

```python
def _iter_sos_sets(model: Model) -> Iterator[tuple[int, list[int], list[float]]]:
    label_to_pos = model.variables.label_index.label_to_pos
    for var_name in model.variables.sos:
        var = model.variables.sos[var_name]
        sos_type = int(var.attrs[SOS_TYPE_ATTR])
        sos_dim = str(var.attrs[SOS_DIM_ATTR])

        labels = var.labels.transpose(sos_dim, ...)        # put SOS dim first
        weights = labels.coords[sos_dim].values            # read once
        arr = labels.values.reshape(labels.shape[0], -1)  # (n_sos, n_sets)

        for i in range(arr.shape[1]):
            positions, kept_weights = _sos_set_positions(arr[:, i], weights, label_to_pos)
            if positions:
                yield sos_type, positions, kept_weights
```

test/test_sos_masked.py
```diff
@@ -0,0 +1,313 @@
...
+
+    return _build
+
+
+# ---------------------------------------------------------------------------
+# Test matrix: 11 fixture configs × 2 SOS types × (solver, io_api)
+# ---------------------------------------------------------------------------
+
+# Each entry: (sos_var_2d, mask_on_sos, mask_on_other)
+FIXTURE_CONFIGS = [
```
FabianHofmann, line 246: I like having a fixture config

FabianHofmann (changes requested):

FBumann:
@FabianHofmann thanks for the honest review. Il fix it up

linopy/solvers.py
FBumann: yes

linopy/io.py
FBumann: I made it shorter. We can drop fully if you prefer

linopy/io.py
FBumann: Done

FabianHofmann (approved):
wonderful @FBumann, feel free to merge

PR #694, opened 2026-05-19
fix(test): only skip GPU-only solvers without --run-gpu (#693)

### LINKED ISSUE #693: Conftest GPU auto-skip incorrectly skips CPU xpress tests
## Problem

`test/conftest.py::pytest_collection_modifyitems` auto-marks any parametrized test where the `solver` param is a backend with `SolverFeature.GPU_ACCELERATION` as a GPU test and skips it without `--run-gpu`.

Xpress carries `GPU_ACCELERATION=True` because it has an optional GPU pathway, but the vast majority of xpress test runs are pure CPU and don't need a GPU. The current rule treats CPU xpress and GPU xpress as the same thing, so any `parametrize("solver", [..., "xpress", ...])` silently drops the xpress cell from the default test run.

## Reproduction

```bash
pytest test/test_sos_masked.py -k xpress -rs
# every xpress case → SKIPPED [reason: need --run-gpu option to run GPU tests]

pytest test/test_sos_masked.py -k xpress --run-gpu
# every xpress case → PASSED
```

Same pattern hits `test_piecewise_constraints.py::test_sos2_per_entity_nan_padding[xpress-*]` and any other suite that parametrizes on solver name.

## Why this matters

Xpress's SOS path was a regression target in #688 — `xpress / direct` and `xpress / lp` are exactly the cells we want covered by default. Today they're hidden behind `--run-gpu`, which CI does not pass, so a real regression in xpress would slip through.

## Suggested fix

Distinguish "GPU-only solver" from "solver with an optional GPU mode":

- `cuPDLPx` (GPU-only) → keep the skip
- `xpress` → don't skip; xpress GPU acceleration is opt-in at solve time via solver options, not at import/parametrize time

A small option here: replace `solver_supports(solver, SolverFeature.GPU_ACCELERATION)` in the conftest hook with a stricter check (e.g. a `GPU_ONLY` feature, or just an explicit set of GPU-only backends). Then xpress runs by default and `cuPDLPx`-style backends still require `--run-gpu`.

## Source

`test/conftest.py`, lines 37–57.

FBumann:
Closes #693.

## Problem

`test/conftest.py::pytest_collection_modifyitems` auto-skipped any parametrized test whose `solver` param carried `SolverFeature.GPU_ACCELERATION`. Xpress sets that flag from version 9.8 for its optional GPU pathway — but xpress's default and primary backend is CPU. The blanket skip hid every `parametrize("solver", …)` xpress cell from default CI:

```bash
pytest test/test_sos_masked.py -k xpress -rs
# all xpress cells → SKIPPED [need --run-gpu option to run GPU tests]
```

Real regressions in xpress (e.g. the SOS path in #688) could slip through default CI as a result.

## Fix

Introduce `SolverFeature.GPU_ONLY` — the property the conftest actually wants to gate on:

|              | `GPU_ACCELERATION` | `GPU_ONLY` |
|--------------|:------------------:|:----------:|
| cuPDLPx      |         ✅          |     ✅      |
| Xpress ≥ 9.8 |         ✅          |     ❌      |
| Gurobi       |         ❌          |     ❌      |

The conftest now checks `GPU_ONLY` instead of `GPU_ACCELERATION`. Xpress keeps `GPU_ACCELERATION` (so the looser GPU tolerance in `test_optimization.py` is still available when xpress is run in its GPU mode) but no longer skips by default.

## Verification

- `test_piecewise_constraints.py::TestSolverSOS2::test_equality_maximize_efficiency[xpress]` runs by default (previously skipped).
- New `test_solver_class_supports_feature` rows lock in the classification: `cuPDLPx → True`, `Xpress → False`, `Gurobi → False`.

### DIFF
--- a/linopy/solvers.py
+++ b/linopy/solvers.py
@@ -116,6 +116,7 @@ class SolverFeature(Enum):
     READ_MODEL_FROM_FILE = auto()
     SOLUTION_FILE_NOT_NEEDED = auto()
     GPU_ACCELERATION = auto()
+    GPU_ONLY = auto()
     IIS_COMPUTATION = auto()
     SOS_CONSTRAINTS = auto()
     SEMI_CONTINUOUS_VARIABLES = auto()
@@ -3295,6 +3296,7 @@ class cuPDLPx(Solver[None]):
         {
             SolverFeature.DIRECT_API,
             SolverFeature.GPU_ACCELERATION,
+            SolverFeature.GPU_ONLY,
             SolverFeature.SOLUTION_FILE_NOT_NEEDED,
         }
     )
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -37,13 +37,19 @@ def pytest_configure(config: pytest.Config) -> None:
 def pytest_collection_modifyitems(
     config: pytest.Config, items: list[pytest.Item]
 ) -> None:
-    """Automatically skip GPU tests unless --run-gpu is passed."""
+    """
+    Auto-skip GPU-only solvers (e.g. cuPDLPx) unless --run-gpu is passed.
+
+    Solvers that *also* have a CPU mode (e.g. xpress, which carries
+    ``GPU_ACCELERATION`` from version 9.8) are not skipped — their default
+    pathway is CPU and they should run in normal CI.
+    """
     if config.getoption("--run-gpu"):
         return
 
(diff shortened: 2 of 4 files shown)


FabianHofmann (approved):
goooood catch!

PR #695, opened 2026-05-19
fix(license-leaks): release license slot from Mosek, COPT, and MindOpt probe and solver paths

### LINKED ISSUE #679: Mosek license leak

- [x] I have confirmed this bug exists on the current [`master`](https://github.com/pypsa/linopy/tree/master) branch of Linopy.

### Issue Description

Hi,

Mosek licenses are bound to the environment and within several Tasks can be created. solver.py calls mosek.Task() without explicit environment and that creates a "private linopy" env. Even if you're not aiming to run a mosek solve you're blocking a license it seems.
Full disclosure: Claude found this, I can't really vogue for this. It seems to be sure though, but they always are. But it seems to fix the leak.

_PATCH_FROM = (
    "    with contextlib.suppress(mosek.Error):\n"
    "        t = mosek.Task()\n"
    "        t.optimize()\n"
    "\n"
    "        available_solvers.append(\"mosek\")"
)

_PATCH_TO = (
    "    # PATCHED by linopy_mosek_probe_fix: use a context-managed Env+Task\n"
    "    # so the probe does not leak a PTS license token forever.\n"
    "    with contextlib.suppress(mosek.Error):\n"
    "        with mosek.Env() as _probe_env, _probe_env.Task(0, 0) as _probe_task:\n"
    "            _probe_task.optimize()\n"
    "        available_solvers.append(\"mosek\")"
)

Cheers,
Andreas

### Reproducible Example

```python
see above
```

### Expected Behavior

see above

### Installed Versions

<details>
Replace this line.
</details>

FBumann:
## Summary

Several solvers leak a license slot because a license-bearing object is constructed without being explicitly released. Started from #679 (Mosek probe) and audited every `_license_probe` and `_run_file` for the same pattern.

### Fixed in this PR

| Site | Bug | Fix |
|---|---|---|
| `Mosek._license_probe` | `mosek.Task()` with no explicit env creates a private `mosek.Env` that holds a license slot and is never released | `with mosek.Env() as env, env.Task(0, 0) as task: …` |
| `Mosek._build_direct`, `Mosek._run_file` | Same bare `mosek.Task()` pattern (cleanup deferred to `_env_stack.close()` only released the task, not the implicit env) | Env explicitly entered onto `_env_stack` before the task |
| `COPT._license_probe` | `coptpy.Envr()` constructed and dropped without `.close()` | Explicit `env.close()` |
| `MindOpt._license_probe` | `mindoptpy.Env()` constructed and dropped without `.dispose()` | Explicit `env.dispose()` |
| `COPT._run_file` | `env_.close()` only ran on the happy path; any exception in `m.solve()` / writes / status parsing leaked the env | Wrapped in `try/finally` |
| `MindOpt._run_file` | Same pattern with `m.dispose()` / `env_.dispose()` | Wrapped in `try/finally` |

Fixes #679.

### Already clean

- `Gurobi._license_probe` (`with gurobipy.Env(): pass`) and `_resolve_env` (env registered on `_env_stack`).
- `Knitro._license_probe` (`KN_new` / `KN_free`) and `_run_file` (`try/finally` with `KN_free`).

### Out of scope (follow-up)

`CPLEX._run_file` (`cplex.Cplex()` stored as `solver_model` with no `m.end()`) and `Xpress._run_file` / `_build_direct` (`xpress.problem()` similarly) have no explicit cleanup at all — they rely on the wrapper's `__del__`. Whether that releases the license cleanly depends on each library's internals, and I don't have CPLEX or Xpress licensed locally to verify, so I'm leaving these for a separate audit with a contributor who can test. Will file a tracking issue.

## Test plan

- [x] New tests in `test/test_available_solvers.py` patch each solver's module with a mock and assert the probe both constructs and closes/disposes the env. Regressions back to bare constructor would fail.
- [x] `pytest test/test_available_solvers.py` — 14 pass.
- [x] `pytest -k "mosek or copt or mindopt" test/test_solvers.py test/test_io.py test/test_optimization.py` — 538 pass, 7 skipped (MindOpt not installed locally). MindOpt `_run_file` change is covered by mock test only; CI has MindOpt and will exercise the real path.
- [x] `ruff check`, `ruff format --check`, `mypy linopy/solvers.py test/test_available_solvers.py` clean.

### DIFF
--- a/linopy/solvers.py
+++ b/linopy/solvers.py
@@ -3210,7 +3210,7 @@ class COPT(Solver[None]):
                 try:
                     m.write(path_to_string(solution_fn))
                 except coptpy.CoptError as err:
-                    logger.info("No model solution stored. Raised error: %s", err)
+                    logger.warning("No model solution stored. Raised error: %s", err)
 
             # TODO: check if this suffices
             condition = m.MipStatus if m.ismip else m.LpStatus
--- a/linopy/solvers.py
+++ b/linopy/solvers.py
@@ -2612,8 +2612,8 @@ class Mosek(Solver[None]):
 
     @classmethod
     def _license_probe(cls) -> None:
-        t = mosek.Task()
-        t.optimize()
+        with mosek.Env() as env, env.Task(0, 0) as task:
+            task.optimize()
 
     def _run_direct(
         self,
@@ -2644,7 +2644,8 @@ class Mosek(Solver[None]):
         assert model is not None
         self.close()
         self._env_stack = contextlib.ExitStack()
-        task = self._env_stack.enter_context(mosek.Task())
+        env = self._env_stack.enter_context(mosek.Env())
+        task = self._env_stack.enter_context(env.Task(0, 0))
         m = self._build_solver_model(
             model,
             task,
@@ -2759,7 +2760,8 @@ class Mosek(Solver[None]):
         assert problem_fn is not None
         self.close()
         self._env_stack = contextlib.ExitStack()
-        m = self._env_stack.enter_context(mosek.Task())
+        mosek_env = self._env_stack.enter_context(mosek.Env())
+        m = self._env_stack.enter_context(mosek_env.Task(0, 0))
         sense = read_sense_from_problem_file(problem_fn)
         io_api = read_io_api_from_problem_file(problem_fn)
         problem_fn_ = path_to_string(problem_fn)
--- a/linopy/solvers.py
+++ b/linopy/solvers.py
@@ -3216,7 +3216,7 @@ class COPT(Solver[None]):
                 try:
                     m.write(path_to_string(basis_fn))
                 except coptpy.CoptError as err:
-                    logger.info("No model basis stored. Raised error: %s", err)
+                    logger.warning("No model basis stored. Raised error: %s", err)
 
             if solution_fn:
                 try:
--- a/linopy/solvers.py
+++ b/linopy/solvers.py
@@ -3028,7 +3028,8 @@ class COPT(Solver[None]):
 
     @classmethod
     def _license_probe(cls) -> None:
-        coptpy.Envr()
+        env = coptpy.Envr()
+        env.close()
 
     def _run_file(
         self,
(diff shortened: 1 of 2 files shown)


FBumann:
@brunoficoxpress if you want to you could give us some advice on how to handle this for xpress :)

brunoficoxpress:
> @brunoficoxpress if you want to you could give us some advice on how to handle this for xpress :)

Hi Fabian, looked into the Xpress side for this PR.

A note on Xpress internals: unlike Mosek/COPT/Gurobi, Xpress uses a global reference-counted license per process (`XPRSinit`/`XPRSfree`). Individual `xpress.problem()` objects do not hold license slots - `problem.__exit__()` calls `XPRSdestroyprob()` (frees memory only) but has no effect on the license. So the fix here is purely about memory management, not license release.

`_run_file` needs the same `ExitStack` pattern as `_build_direct` (rather than a plain `with` block) because `_solve` stores `m` as `self.solver_model` for post-solve use (e.g. `compute_infeasibilities`). The fix mirrors what Gurobi (`_build_direct`) and Mosek (`_build_direct` + `_run_file`) already do - 5 lines across both methods. Happy to open a PR targeting your branch if useful.

```diff
@@ -2037,11 +2037,14 @@ class Xpress(Solver[None]):
     ) -> None:
         model = self.model
         assert model is not None
+        self.close()
+        self._env_stack = contextlib.ExitStack()
         problem = self._build_solver_model(...)
+        self._env_stack.enter_context(problem)
         self.solver_model = problem

@@ -2224,7 +2227,9 @@ class Xpress(Solver[None]):
-        m = xpress.problem()
+        self.close()
+        self._env_stack = contextlib.ExitStack()
+        m = self._env_stack.enter_context(xpress.problem())
         m.read(path_to_string(problem_fn))

**Comment by FBumann** [comment]
@brunoficoxpress Thanks for the quick response. A PR would be highly appreciated. As its more of a memory leak and not a license leak, I would suggest opening it agains master directly.

File: linopy/solvers.py [hunk 1]
```diff
@@ -3193,70 +3196,71 @@ def _run_file(
...
-        # TODO: check if this suffices
-        condition = m.MipStatus if m.ismip else m.LpStatus
-        termination_condition = CONDITION_MAP.get(condition, str(condition))
-        status = Status.from_termination_condition(termination_condition)
-        status.legacy_status = str(condition)
+            if solution_fn:
+                try:
+                    m.write(path_to_string(solution_fn))
+                except coptpy.CoptError as err:
+                    logger.info("No model solution stored. Raised error: %s", err)
```
FabianHofmann:
```suggestion
                    logger.warning("No model solution stored. Raised error: %s", err)
```

linopy/solvers.py
```diff
@@ -3193,70 +3196,71 @@ def _run_file(
...
-        if solution_fn:
-            try:
-                m.write(path_to_string(solution_fn))
-            except coptpy.CoptError as err:
-                logger.info("No model solution stored. Raised error: %s", err)
+            if basis_fn and m.HasBasis:
+                try:
+                    m.write(path_to_string(basis_fn))
+                except coptpy.CoptError as err:
+                    logger.info("No model basis stored. Raised error: %s", err)
```
FabianHofmann:
```suggestion
                    logger.warning("No model basis stored. Raised error: %s", err)
```

FabianHofmann (approved):

FBumann:
@FabianHofmann For some reason i only commited one of your 2 suggestion. Did the other now!

PR #697, opened 2026-05-19
refactor(remote): OETC/SSH as standalone transports (closes #683)

### LINKED ISSUE #683: Refactor OETC as a `Solver` subclass

FBumann:
> Stacked on top of #690 (rebase target after that lands on master).

Closes #683 — with a different design than the issue proposed.

## Why standalone, not `Solver` subclass

The issue framed OETC as a `Solver` subclass so `Model.solve` could fold the `remote=` branch into the unified Solver pipeline. The fit was wrong:

- Remote handlers aren't solvers — they ship a netcdf elsewhere and let *someone else* solve.
- Forcing them through `Solver` required workarounds: a non-colliding field for the worker's solver, property-vs-field collisions on `solver_name`, `SolverName` enum entries for things that aren't algorithms.
- `Solver`'s feature-flag plumbing is about local solver capabilities; for remotes we want to validate the *worker's* solver flags.

So **`Oetc` and `SSH` are standalone classes in `linopy/remote/`, parallel to `Solver`, not subclasses of it.**

## What's new

```python
m.solve("gurobi", remote=OetcSettings(email=..., password=..., name=..., ...), Method=2)
m.solve("highs", remote=SshSettings(hostname=..., setup_commands=["conda activate linopy-env"]))
```

- `solver_name` and `**solver_options` are the same axes as for a local solve; `remote=` selects *where* to run.
- `Model.solve(remote=...)` folds the solution into your model **in place** and returns `(status, termination_condition)` — behaviour-identical to v0.7.0.
- New `model.remote` attribute mirrors `model.solver` for post-solve introspection and connection reuse.
- `SshSettings.setup_commands: list[str]` runs shell commands on the remote before the solver — replaces the old "build a `RemoteHandler` and call `.execute(...)`" pattern.
- Pip extra `remote` → `ssh`; OETC has its own `linopy[oetc]` extra. `linopy[remote]` is dropped — see Breaking Changes in the release notes.

## Remote calls return a `Model`

`Oetc.solve` / `Oetc.collect` / `SSH.solve` return the **solved `Model`**, not a `Result`. The worker (`read_netcdf → m.solve → to_netcdf`) natively produces a fully inflated solved model — forcing it into the flat `Result` shape meant a redundant `Model → flat arrays → Model` round-trip.

- `Model.solve(remote=...)` folds the solved model into the caller's model via the new `Model._assign_from_solved_model` — the v0.7.0 in-place fold logic, extracted into a method. So `Model.solve(remote=...)` is behaviour-identical to v0.7.0: it mutates the original model and returns the status tuple.
- The standalone `Oetc` / `SSH` objects hand back a *new* solved `Model`; the model you pass in is not modified.
- `_scatter_solution_from_solved_model` is deleted. Local solves are unchanged — solvers natively emit flat label-indexed arrays, so they keep the `Result` / `Model.assign_result` path.

Each path is now direct — no flatten/reinflate hop.

## `Oetc` is a connection, not a job

`Oetc(settings)` is a reusable connection: authenticate once, drive any number of jobs.

```python
oetc   = Oetc(settings)
uuid   = oetc.submit(model, "gurobi", Method=2)   # upload + dispatch → job uuid
state  = oetc.status(uuid)                        # one non-blocking status check
solved = oetc.collect(uuid)                       # block until done → solved Model
solved = oetc.solve(model, "gurobi", Method=2)    # one-shot = submit + collect
```

A job is identified solely by its **uuid string**, which makes the lifecycle genuinely async-friendly:

- **Multi-model** — one connection, a list of uuids.
- **Cross-process / resume** — persist the uuid; a fresh `Oetc(OetcSettings.from_env())` in another process (or hours later) collects it. `collect()` needs nothing but the uuid and the settings.
- **Non-blocking** — `status(uuid)` for a user-driven poll loop; the sync primitives compose into any async runtime via `asyncio.to_thread` / a thread pool.

`SSH(settings)` is the synchronous analogue — a reusable connection with `solve(model, solver_name, **options)`; no submit/collect seam (the SSH session is held open and the solve is synchronous).

**Solver config is per call**, not per object: `solver_name` / `**options` are arguments to `submit` / `solve` / `Model.solve`, mirroring the local-solve API.

## Token re-authentication

The OETC auth token has a limited lifetime; `Oetc` now re-authenticates transparently:

- `Oetc._session()` rebuilds the handler when the token has expired — covers a connection reused across the token lifetime (submit, then collect later).
- `OetcHandler.wait_and_get_job_data()` re-signs-in at the top of each poll iteration — so a single `collect()` whose blocking poll outlives the token (the normal case for large solves) keeps working. GCP transfers are unaffected; they use the service-account key, not the user JWT.

## SSH is narrower than `RemoteHandler`

`SSH` covers **solving** a model on a remote machine — not being a general remote-shell wrapper.

| | Old `RemoteHandler` | New `SSH` | Replacement |
|---|---|---|---|
| Solve a model | `solve_on_remote(m)` | `solve(m, solver_name)` | direct rename |
| Env activation before solve | `handler.execute("conda activate …")` | — | `SshSettings.setup_commands=[…]` |
| Reuse connection across solves | implicit (long-lived handler) | implicit (`SSH(settings)` reused) | works the same |
| Arbitrary remote shell commands | `handler.execute(any cmd)` | — | drop to `RemoteHandler` (deprecated) or paramiko directly |

## Deprecations

- `OetcHandler(...)` / `RemoteHandler(...)` construction emits `DeprecationWarning`. Their `solve_on_oetc` / `solve_on_remote` return contracts are unchanged; `solve_on_oetc` delegates to `Oetc` internally.
- `Model.solve(remote=<Handler>)` is deprecated — pass the settings dataclass instead.
- `OetcCredentials` is deprecated — pass `email=` / `password=` directly to `OetcSettings`.
- `OetcSettings.solver` / `OetcSettings.solver_options` are deprecated — pass the solver to `Model.solve(solver_name, ...)` or `Oetc.submit(model, solver_name, ...)`. During deprecation they are still honoured (as a fallback when `Model.solve(remote=OetcSettings(...))` is called without a `solver_name`, and by the deprecated `OetcHandler`), and will be removed with `OetcHandler`.

## Hand testing needed

The entire `test/remote/` suite is **mocked** — no real network. Everything that touches a live OETC account or SSH server needs manual verification before merge:

**OETC** (account + `linopy[oetc]`):
- [ ] `m.solve("gurobi", remote=OetcSettings(...))` end-to-end — auth, GCP upload, job submit, poll, download, solution folded onto `m`.
- [ ] `Oetc(settings)` driven manually — `submit` / `status` / `collect`; `solve()` one-shot.
- [ ] **Cross-process resume** — submit in one process, persist the uuid, `collect()` it from a fresh process.
- [ ] **Token re-auth** — a solve (or a submit→collect gap) longer than the OETC token lifetime; confirm it re-authenticates instead of failing. This is the one path the mocked tests cannot exercise against reality.
- [ ] `OetcSettings.from_env()`.
- [ ] Deprecated paths still work: `OetcHandler(...).solve_on_oetc(...)`, `Model.solve(remote=<OetcHandler>)`, `Model.solve(remote=OetcSettings(..., solver="gurobi"))`.

**SSH** (remote server + `linopy[ssh]`):
- [ ] `m.solve("gurobi", remote=SshSettings(hostname=..., setup_commands=[...]))` end-to-end.
- [ ] `SSH(settings)` reused across several `.solve()` calls — one connection, `setup_commands` run once.
- [ ] Deprecated paths: `RemoteHandler(...).solve_on_remote(...)`, `Model.solve(remote=<RemoteHandler>)`.

**Docs:**
- [ ] `examples/remote-machines.ipynb` against real services (it is `nbsphinx execute=never`, so never CI-checked).

## Follow-ups (not in this PR)

- **Remove `OetcHandler` / `RemoteHandler`.** Their private transport methods still live on the Handler classes; `Oetc` / `SSH` reach into `self._handler._...` to use them. A follow-up will migrate that code into `Oetc` / `SSH` and port the OETC handler-internals tests.
- **Remove `OetcCredentials`, `OetcSettings.solver` / `solver_options`** — after one release cycle of deprecation warnings.
- **In-poll re-auth is proactive** (`is_expired` check); a reactive re-auth on a `401` would also cover a server expiring the token earlier than advertised.

## Test plan

- [x] `pytest test/remote/ test/test_sos_reformulation.py test/test_oetc_settings.py` — 193 pass (all mocked).
- [x] `ruff check`, `ruff format`, `mypy` — clean.
- [ ] Broader suite (`pytest --ignore=test/remote`) — re-run before merge.
- [ ] Hand testing — see the checklist above.


FabianHofmann:
@FBumann I think the concept makes sense. I would like to make sure that it is flexible enough for interfaces like gurobi instant cloud. could you take that into consideration?

FBumann:
@FabianHofmann Yes, I already did think about it. But i can sketch it out a bit more, maybe even create a Remote parent class...
I also didnt add the tests yet, and some cleanupy might also be possible. And the latest context manager from #690 isnt used yet

FBumann:
> @FBumann I think the concept makes sense. I would like to make sure that it is flexible enough for interfaces like gurobi instant cloud. could you take that into consideration?

It fits well for async instant cloud. For the sync instan cloud it doesnt fit as cleanly. But I think it can be wired in without adjusting public api. Probably just refactor some internal helpers for collection

FabianHofmann:
> > @FBumann I think the concept makes sense. I would like to make sure that it is flexible enough for interfaces like gurobi instant cloud. could you take that into consideration?
> 
> It fits well for async instant cloud. For the sync instan cloud it doesnt fit as cleanly. But I think it can be wired in without adjusting public api. Probably just refactor some internal helpers for collection

I guess async is the most important one

FBumann:
> > > @FBumann I think the concept makes sense. I would like to make sure that it is flexible enough for interfaces like gurobi instant cloud. could you take that into consideration?
> > 
> > 
> > It fits well for async instant cloud. For the sync instan cloud it doesnt fit as cleanly. But I think it can be wired in without adjusting public api. Probably just refactor some internal helpers for collection
> 
> I guess async is the most important one

I thin the regular gurobi cloud can be used by just using the regular gurobi solver anyway...? It fully dispatches to gurobipy, and gets a gurobipy solution back, right? SO no extra class needed...

FBumann:
@FabianHofmann 
Here is some advice from CLaude after some back and forth:

```text
## Generality check — Gurobi & other clouds

- Gurobi Instant Cloud / Compute Server sync → already works via m.solve("gurobi", env=cloud_env). Belongs as GurobiSolver config, not a remote class. A GurobiInstantCloud remote class would be a fake — construct an Env and dispatch back into the regular Solver path.
- Gurobi batch solve / detached submit-now-collect-later → genuine fit for the remote handler shape.
- Friction for any future handler: solver_name is mandatory and means inner solver (redundant for solver-specific clouds); _scatter_solution_from_solved_model assumes a netcdf round-trip with a linopy Model on the other side.

## Naming clarity

"Remote" is defensible as user-facing vocabulary but mislabels the technical category, and "async" is the wrong axis entirely:

- The actual invariant shared by Oetc and SSH is a second linopy instance does the solve (local linopy serializes → remote linopy deserializes → .solve() → result back). Gurobi Cloud is also remote geographically, but only one linopy instance is involved — gurobipy talks to a remote engine too.
- "Async" misclassifies SSH (fully synchronous, no job seam) and overpromises on Oetc (.solve() blocks; the seam is optional). True async (blocking=False returning a handle) is orthogonal — it could apply to a local callback-driven Gurobi solve too.
```

SO the new architecture should cover all cases. THe only issue i see is thst "remote" isnt a precise name. "worker" or "offload" might be better. But thats not as important for now i think

FBumann:
@FabianHofmann I argue that the final form of API would be removing `OetcHandler`.
The High level api would be `OetcSettings` passed to `Model.solve()`. The low level API is using `Oetc.from_model()` like a Solver. 

`OetcHandler` delegates for backwards compat, but is deprecated.

This reflects the 2-level state of regular solvers: Pass name + options in `Model.solve()`, or directly invoke with the `Solver` class, which holds the logic/validation for the solver.

Do you agree?

FBumann:
@FabianHofmann Ready for a review

FBumann:
@FabianHofmann ready. Do you wait for feedback of oetc users?

FBumann:
@FabianHofmann Ill look into the async stuff for oetc a bit, but cant test without oetc credentials. SO we will see how far i can go

FBumann:
As im looking into async usage with oetc, the design decision of having Ssh and Oetc return `Result` like regular solvers bites me. With this design, `Remotes` manipulate the original model, which therefore needs to keep existing, instead of simply returning the collected one.

So i think changing this to make `Ssh`/`Oetc` return the the collected `Model` is better in many ways, even if its conceptually different from what `Solver`s do!

FBumann:
**Design change since the description: `Oetc`/`SSH` return a solved `Model`, not a `Result`.**

The "Remote returns `Model` instead of `Result`" alternative — listed as *rejected* under "Other designs considered" — is now the chosen design.

**Key rationale:** `Oetc` / `SSH` get a fully solved model back from the worker, so the simplest thing is to return it as-is. `Model.solve(remote=...)` is the only path that needs to write the solution into the caller's existing model.

- `Oetc.solve()` / `Oetc.collect()` / `SSH.solve()` return the solved `Model` directly — no conversion. `collect()` no longer takes the source model, so it is process-independent (collect a job by uuid from a fresh process).
- `Model.solve(remote=...)` folds the solved model into the caller's model via the new `Model._assign_from_solved_model` — the v0.7.0 in-place fold logic (`model.py` lines 1531-1540) extracted into a method. So `Model.solve(remote=...)` is behaviour-identical to v0.7.0: it mutates the original model and returns `(status, termination_condition)`, never a Model.
- `_scatter_solution_from_solved_model` (the `Model → flat Result` adapter) is deleted. Local solves keep `Result` / `assign_result` — solvers natively emit flat label-indexed arrays.

_To be folded into the PR description._

FBumann:
@FabianHofmann I put some more work in. Its now truly ready for async polling, fully relying on a job uuid.
But i cant test it, see the notes on the PR description

coroa:
I am a bit curious how some of these design decisions match with the Gurobi Batch API, where we have less leeway in changing how they work.

The connection object probably makes sense also in that case (ie. you genuinely do have a uuid str) and you do authenticate.

You share a lot of the send model to solver infra with the actual Gurobi solver (ie. one just needs to set up the Gurobi env differently), i am unsure how one could reuse this in this infrastructure.

What is new is:
1. you need to call `model.optimizeBatch()` instead of `model.optimize()` on the model object
2. you get the status with the uuid str
3. you retrieve the result with `gp.Batch(batchID, env).getJSONSolution()` which needs to be parsed into a Result.

Espeially 3, does not fit with the return Model, not Result, i fear.

FBumann:
> I am a bit curious how some of these design decisions match with the Gurobi Batch API, where we have less leeway in changing how they work.
> 
> The connection object probably makes sense also in that case (ie. you genuinely do have a uuid str) and you do authenticate.
> 
> You share a lot of the send model to solver infra with the actual Gurobi solver (ie. one just needs to set up the Gurobi env differently), i am unsure how one could reuse this in this infrastructure.
> 
> What is new is:
> 
> 1. you need to call `model.optimizeBatch()` instead of `model.optimize()` on the model object
> 2. you get the status with the uuid str
> 3. you retrieve the result with `gp.Batch(batchID, env).getJSONSolution()` which needs to be parsed into a Result.
> 
> Espeially 3, does not fit with the return Model, not Result, i fear.

Im not using gurobi batch, so I am happy that you jump in here.
Could you explain or provide a link to what exactly gurbi batch is? Especially syn or async! As this is a very important distinction, wether it needs its own class.
Thus far I hope we can treat it as a regular Solver class. Maybe even fold it onto the existing GurobiSolver

FBumann:
@coroa Looked a bit into it:

Here is what I found:
- Gurobi Batch is async + uuid-based
- It sends/recieves a `gp.Model`


So its in between `Oetc` and `GurobiSolver`.
Oetc sends a linopy Model, and receives a linopy model - async.
GurobiSolver sends a gurobi.Model and receives a gurobi.Model. (send/recieve on the same machine, no networking)

GurobiBatch sends a gurobi.Model and receives a gurobi.Model - to another machine - async.

So GurobiBatch uses similar primitives (connection/uuid lifecycle), but doesnt use a second linopy instance. So it should naturally reuturn a `Result`, not a new `Model`, and is sth diffferent than OETC or SSH

For me, this is a clean Follow up on this PR, with a `GurobiBatch` class that does async gurobi solving.

| | Materialization | Natural return | Sync/async |
|---|---|---|---|
| `GurobiSolver` (incl. Compute Server sync) | gurobipy model | `Result` | sync |
| **Gurobi Batch** (future) | gurobipy model | `Result` | **async** (BatchID) |
| `SSH` | netcdf → 2nd linopy | `Model` | sync |
| `Oetc` | netcdf → 2nd linopy | `Model` | async (uuid) |

coroa:
Gurobi's RemoteServices offering is honestly confusing.

> Here is what I found:
> 
> * Gurobi Batch is async + uuid-based
> * It sends/recieves a `gp.Model`
> 
> So its in between `Oetc` and `GurobiSolver`. Oetc sends a linopy Model, and receives a linopy model - async. GurobiSolver sends a gurobi.Model and receives a gurobi.Model. (send/recieve on the same machine, no networking)

Note, that you can also do networking with the GurobiSolver class already as it is now, with the example in:

https://docs.gurobi.com/projects/remoteservices/en/13.0/content/programming-with-remote-services/api-compute-server-job.html

ie. you specify a `ComputeServer` and authentication as environment argument and then call `model.optimize` as you do usually. Log and everything is then streamed as if a local gurobi server would be in use, but it is blocking.

> GurobiBatch sends a gurobi.Model and receives a gurobi.Model - to another machine - async.

`optimizeBatch` on the other hand returns a so-called BatchID string (effectively a uuid) and then allows you with this instantiate a Batch representative object that you can use to query the status. Most concise description i know of is in https://docs.gurobi.com/projects/optimizer/en/current/features/batchoptimization.html .

> 
> So GurobiBatch uses similar primitives (connection/uuid lifecycle), but doesnt use a second linopy instance. So it should naturally reuturn a `Result`, not a new `Model`, and is sth diffferent than OETC or SSH
> 

Well, is it really something different than OETC? Note that we could probably also change OETC itself to just return a result internally. I guess that would be more efficient anyway? @KristijanFaust-OET

I suggested originally to put OETC/SSH as a Solver since the remote/local stuff is more an internal detail rather than the Async/Sync distinction.


> For me, this is a clean Follow up on this PR, with a `GurobiBatch` class that does async gurobi solving.

Well, the `remote` argument is not so clear to me, since the Materialization issue is something that could be changed, couldn't it?

FBumann:
> Gurobi's RemoteServices offering is honestly confusing.
> 
> > Here is what I found:
> > 
> > * Gurobi Batch is async + uuid-based
> > * It sends/recieves a `gp.Model`
> > 
> > So its in between `Oetc` and `GurobiSolver`. Oetc sends a linopy Model, and receives a linopy model - async. GurobiSolver sends a gurobi.Model and receives a gurobi.Model. (send/recieve on the same machine, no networking)
> 
> Note, that you can also do networking with the GurobiSolver class already as it is now, with the example in:
> 
> https://docs.gurobi.com/projects/remoteservices/en/13.0/content/programming-with-remote-services/api-compute-server-job.html
> 
> ie. you specify a `ComputeServer` and authentication as environment argument and then call `model.optimize` as you do usually. Log and everything is then streamed as if a local gurobi server would be in use, but it is blocking.
> 
> > GurobiBatch sends a gurobi.Model and receives a gurobi.Model - to another machine - async.
> 
> `optimizeBatch` on the other hand returns a so-called BatchID string (effectively a uuid) and then allows you with this instantiate a Batch representative object that you can use to query the status. Most concise description i know of is in https://docs.gurobi.com/projects/optimizer/en/current/features/batchoptimization.html .
> 
> > So GurobiBatch uses similar primitives (connection/uuid lifecycle), but doesnt use a second linopy instance. So it should naturally reuturn a `Result`, not a new `Model`, and is sth diffferent than OETC or SSH
> 
> Well, is it really something different than OETC? Note that we could probably also change OETC itself to just return a result internally. I guess that would be more efficient anyway? @KristijanFaust-OET
> 
> I suggested originally to put OETC/SSH as a Solver since the remote/local stuff is more an internal detail rather than the Async/Sync distinction.
> 
> > For me, this is a clean Follow up on this PR, with a `GurobiBatch` class that does async gurobi solving.
> 
> Well, the `remote` argument is not so clear to me, since the Materialization issue is something that could be changed, couldn't it?

I like this discussion!

About Gurobi: 
Thats what I meant: all sync stuff can already be done through GurobiSolver, as its essentially **just solver parameters** (and an env)

About Async:

If we adjust `Oetc` to only send and recieve the solver model (or lp file?) we could generalize both oetc and gurobi batch into an async solver. `SSH` is still a bit oddly placed then though...

Im not sure if generalizing now is the right call though. I would rather just add GurobiBatch and generaliez/refactor while we do that. I dont use either Oetc nor SSH nor gurobiBatch, so i cant even verify behaviour...

I see nothing about this blocking fro this PR though. (Except that someone with oetc credentials needs to test it of course)

FBumann:
## In summary:

I think we should focus on the api for passing settings and the async primitives (submit, status, collect).
ANd the fact that SSH, OETC (and GurbiBatch) are a different concept than Solvers.

If we agree on those, adding GurobiBatch later, the refactoring should be purely internal.

@coroa What do you say?

PR #542, opened 2025-12-29
Add Xpress 9.8+ API support with backward compatibility

brunoficoxpress:
## Summary

This PR updates the linopy Xpress solver integration to support the new Python API introduced in Xpress 9.8+, while maintaining full backward compatibility with Xpress 9.6+ through fallback mechanisms.

### 1. API Method Migration (`linopy/solvers.py`)
All method calls updated with try/except blocks for backward compatibility:
- `m.read()` → `m.readProb()`
- `m.setlogfile()` → `m.setLogFile()`
- `m.readbasis()` → `m.readBasis()`
- `m.writebasis()` → `m.writeBasis()`
- `m.writebinsol()` → `m.writeBinSol()`
- `m.postsolve()` → `m.postSolve()`
- `m.getnamelist()` → `m.getNameList()`
- `m.getDual()` → `m.getDuals()`

### 2. Status Handling Improvements (`linopy/solvers.py`)
- Replaced numeric status codes with `xpress.SolStatus` enum constants
- Updated solver status checks to use `xpress.enums.SolveStatus.STOPPED`
- Improved code readability and maintainability

### 3. Exception Handling (`linopy/solvers.py`)
- Replaced broad `Exception` catching with specific `xpress.SolverError` and `xpress.ModelError`
- Prevents masking of unrelated errors while catching only solver-specific issues

### 4. IIS (Irreducible Infeasible Subset) Detection (`linopy/model.py`)
- `solver_model.iisall()` → `solver_model.IISAll()` with fallback
- `solver_model.getiisdata()` → `solver_model.getIISData()` with fallback
- Fixed mypy type checking errors by declaring variables before try/except blocks

### 5. Unicode Handling (`linopy/constraints.py`)
- Added Windows console encoding fix for Unicode characters (≤, ≥) in constraint display

### 6. Documentation Updates
- Updated Xpress documentation URLs in `doc/index.rst` and `doc/prerequisites.rst`

### Test Environment
- **Xpress 9.8.0**: Primary development version
- **Xpress 9.6.0**: Backward compatibility validation

### Test Results

- ✅ All 414 Xpress-specific tests **PASSED**
- ✅ All infeasibility detection tests **PASSED** (IIS functionality verified)
- ✅ Pre-commit hooks (Black, Ruff): **All passed**
- ✅ Backward compatibility confirmed through fallback mechanisms

PR #701, opened 2026-05-19
Xpress 9.9+ backward compatibility and other updates

brunoficoxpress:
Restore Xpress 9.9+ API backward compatibility and fix deprecation warnings

Closes # (if applicable).

Xpress 9.8 renamed several API methods (e.g. `loadproblem` → `loadLP/loadMIP/loadQP/loadMIQP`, `addnames` → `addNames`, `getDual` → `getDuals`, etc.). PR #542 added `try/except AttributeError` wrappers to support both old and new API. These wrappers were inadvertently dropped in the Xpress direct-IO rewrite (787c4ce), reintroducing deprecation warnings when running against Xpress 9.9.

This PR restores the wrappers across all affected call sites in:

- **Problem loading:** selects `loadLP`, `loadMIP`, `loadQP`, or `loadMIQP` based on problem type (integer variables / quadratic objective); falls back to `loadproblem` on Xpress < 9.8
- **Names:** `addNames()` with `addnames()` fallback
- **File I/O:** `readProb`, `readBasis`, `writeBasis`, `writeBinSol` with lowercase fallbacks
- **Solve helpers:** `postSolve`, `setLogFile` with lowercase fallbacks
- **Solution retrieval:** `getDuals()` (was incorrectly `getDual()`)

## Checklist

### DIFF
--- a/linopy/solvers.py
+++ b/linopy/solvers.py
@@ -2180,34 +2180,105 @@ class Xpress(Solver[None]):
             entind = None
             coltype = None
 
-        problem.loadproblem(
-            probname="linopy",
-            rowtype=rowtype,
-            rhs=rhs,
-            rng=None,
-            objcoef=np.asarray(M.c, dtype=float),
-            start=start,
-            collen=None,
-            rowind=rowind,
-            rowcoef=rowcoef,
-            lb=lb,
-            ub=ub,
-            objqcol1=objqcol1,
-            objqcol2=objqcol2,
-            objqcoef=objqcoef,
-            qrowind=None,
-            nrowqcoefs=None,
-            rowqcol1=None,
-            rowqcol2=None,
-            rowqcoef=None,
-            coltype=coltype,
-            entind=entind,
-            limit=None,
-            settype=None,
-            setstart=None,
-            setind=None,
-            refval=None,
-        )
+        objcoef = np.asarray(M.c, dtype=float)
+        has_q = objqcol1 is not None
+        has_int = coltype is not None
+        try:  # Try new API first (Xpress 9.8+)
+            if has_q and has_int:
+                problem.loadMIQP(
+                    probname="linopy",
+                    rowtype=rowtype,
+                    rhs=rhs,
+                    rng=None,
+                    objcoef=objcoef,
+                    start=start,
+                    collen=None,
+                    rowind=rowind,
+                    rowcoef=rowcoef,
+                    lb=lb,
+                    ub=ub,
+                    objqcol1=objqcol1,
+                    objqcol2=objqcol2,
+                    objqcoef=objqcoef,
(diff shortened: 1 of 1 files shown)


linopy/solvers.py
```diff
@@ -2180,34 +2180,105 @@ def _build_solver_model(
...
-            setstart=None,
-            setind=None,
-            refval=None,
-        )
+        objcoef = np.asarray(M.c, dtype=float)
+        has_q = objqcol1 is not None
+        has_int = coltype is not None
+        try:  # Try new API first (Xpress 9.8+)
+            if has_q and has_int:
+                problem.loadMIQP(
```
FabianHofmann, line 2201: in order to boil this down a bit I would suggest to pack common kwargs in a dict passed to each of the functions below.

linopy/solvers.py
```diff
@@ -2285,7 +2366,7 @@ def _run_file(
         sense = read_sense_from_problem_file(problem_fn)
 
         m = xpress.problem()
-        m.read(path_to_string(problem_fn))
+        m.readProb(path_to_string(problem_fn))
```
FabianHofmann: this is old xpress compatible?

linopy/solvers.py
```diff
@@ -2365,7 +2461,7 @@ def get_solver_solution() -> Solution:
                 if m.attributes.rows == 0:
                     dual = np.array([], dtype=float)
                 else:
-                    dual_values = np.asarray(m.getDual(), dtype=float)
+                    dual_values = np.asarray(m.getDuals(), dtype=float)
```
FabianHofmann: this raises with old versions. add fallback logic

FabianHofmann (changes requested):
wonderful! thanks for being so quick! In the refactoring I could not test newer xpress versions and according to the docs the loadproblem should still be valid, but this is better! 

some of our projects still rely on xpress 9.4 which is the lower bound of the compat range. would you mind checking this version as well?

brunoficoxpress:
> wonderful! thanks for being so quick! In the refactoring I could not test newer xpress versions and according to the docs the loadproblem should still be valid, but this is better!
> 
> some of our projects still rely on xpress 9.4 which is the lower bound of the compat range. would you mind checking this version as well?

Good points, thanks for the quick turnaround! Updated the PR with your suggestions.

Tested on both 9.4 and 9.6 - all LP/MIP tests pass on both versions, all pre-9.8 fallbacks exercised correctly (`loadproblem`, `read`, `addnames`, `getDual`, etc.).

If you get any issues, or some other queries related to Xpress, feel free to reach out or tag me. Thank you for the nice work with the library :).

FabianHofmann:
What a responsiveness ;) thanks bruno, merging now

PR #710, opened 2026-05-21
fix: keep coords dimension order for DataArray bounds (#706)

### LINKED ISSUE #706: add_variables: DataArray bounds with missing dimensions yield wrong dimension order
### Issue Description

### Reproducible Example

```python
import xarray as xr
import numpy as np
import linopy

m = linopy.Model()

```

Output:

```
foo: ('x', 'y')
bar: ('y', 'x')
```

### Expected Behavior

### Root Cause

### Installed Versions

FBumann:
## Summary

- **`add_variables`** — when a `DataArray` bound is missing some of the dimensions in `coords`, it is expanded via `DataArray.expand_dims`, which **prepends** the new dimensions and their coordinate variables. Since a `Dataset` derives its dimension order from coordinate insertion order, the resulting variable's dimension order depended on the *type* of the bounds (a scalar bound kept `coords` order; two `DataArray` bounds missing the same dim produced a prepended order). The expanded array is now transposed back to `coords` order and reconstructed with its coordinate variables in that order, so the variable's dimensions always follow `coords`. Closes #706.
- **`_broadcast_points`** (piecewise) — the dimensions missing from the breakpoint array were collected into a Python `set`, so the order `expand_dims` prepended them depended on hash randomization and varied between processes. The expansion map is now built by iterating expressions and their dimensions in order, giving a stable, reproducible dimension order.
- Added regression tests for both.

The fix is a no-op for the common cases: scalar bounds and full-dimensional `DataArray` bounds never reach the new code path. Only `DataArray` bounds with missing dimensions pay the (data-size-independent) reordering cost.

## Test plan

- [x] `test/test_variable.py` — incl. new `test_dataarray_broadcast_missing_dim_order`
- [x] `test/test_piecewise_constraints.py` — incl. new `test_broadcast_points_dim_order_follows_exprs`
- [x] `test/test_piecewise_feasibility.py`, `test/test_model.py`
- [x] `ruff check` / `ruff format` clean
- [x] `mypy` clean

### DIFF
--- a/linopy/model.py
+++ b/linopy/model.py
@@ -135,7 +135,8 @@ def _validate_dataarray_bounds(arr: Any, coords: Any) -> Any:
 
     - Raises ``ValueError`` if the array has dimensions not in coords.
     - Raises ``ValueError`` if shared dimension coordinates don't match.
-    - Expands missing dimensions via ``expand_dims``.
+    - Expands missing dimensions via ``expand_dims`` and transposes the
+      result to the ``coords`` dimension order.
     """
     if not isinstance(arr, DataArray):
         return arr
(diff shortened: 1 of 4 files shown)


MaykThewessen (approved):
LGTM. Focused fix, full CI green, tests cover the three bound-type permutations that triggered #706 (scalar+da, da+scalar, da+da).

Two notes:

1. The `target` reconstruction block in `_validate_dataarray_bounds` guards against `expand_dims` leaving non-dim coords in a different order than the transposed dims. Worth a one-line comment explaining why the second `DataArray(...)` rebuild is needed — purely from reading the diff it looks like dead code after the `transpose(*expected)` call, since `transpose` already reorders dims. The case it catches (auxiliary `arr.coords` keys outside `expected`) is non-obvious.

2. The `piecewise._broadcast_points` fix is the more impactful one for us — silent run-to-run reordering from `set` iteration is exactly the kind of thing that bites multi-process scenario sweeps. Glad it has a regression test.

Useful for downstream PyPSA-Eur consumers that pass `DataArray` bounds via `add_variables` in custom constraint modules. Tested mentally against our `state_of_charge_set` and PST tap-extendable code paths — both bypass this path (scalar bounds), so no breakage expected.

FBumann:
Superseeded by #722

Issue #709, opened 2026-05-21
add_variables: pandas Series/DataFrame bounds with missing dimensions silently drop the dimension


FBumann:
### Issue Description

Follow-up to #450 / #614 / #706, covering **pandas** bounds.

When `lower`/`upper` bounds passed to `add_variables` are missing a dimension that is present in `coords`, the bound should be broadcast to the full dimensionality from `coords` (this is what #614 established for `DataArray` bounds).

For **pandas `Series`/`DataFrame`** bounds this does not happen. If a pandas bound is missing a dimension, the missing dimension is **silently dropped** and the resulting variable has the wrong *shape* — not merely the wrong dimension order (#706), but a missing dimension entirely.

As with #706, the behavior depends on the *type* of the other bound: if at least one bound is a scalar, the scalar carries the full `coords` dimensionality and broadcasts the pandas bound up, so it accidentally works. If **both** bounds are pandas objects missing the same dimension, that dimension is lost.

### Reproducible Example

```python
import pandas as pd
import linopy

x = pd.Index(["a", "b", "c"], name="x")
y = pd.Index(["X", "Y"], name="y")

# pandas Series bound, indexed only by 'x' (missing 'y')
s = pd.Series([1.0, 2.0, 3.0], index=x)

m = linopy.Model()

# scalar lower -> dims follow coords: ('x', 'y')
foo = m.add_variables(lower=0, upper=s, coords=[x, y], name="foo")
print("foo:", foo.dims)   # ('x', 'y')

# both bounds are Series missing 'y' -> 'y' is dropped entirely
bar = m.add_variables(lower=-s, upper=s, coords=[x, y], name="bar")
print("bar:", bar.dims)   # ('x',)  <-- wrong, 'y' is missing
```

Output:

```
foo: ('x', 'y')
bar: ('x',)
```

### Expected Behavior

Both `foo` and `bar` should have dimensions `('x', 'y')` — pandas bounds missing a dimension should be broadcast to the full dimensionality given by `coords`, consistent with `DataArray` bounds.

### Root Cause

Two pieces conspire:

1. `linopy/common.py`, `pandas_to_dataarray` builds the DataArray dimensions **only from the pandas object's own axes** and ignores the `coords` argument entirely (`DataArray(arr, coords=None, dims=dims)`).
[...]

Issue #708, opened 2026-05-21
Operators align coordinates inconsistently (`*` by label, `+` `-` `/` by position)


FBumann:
### Issue Description

linopy's binary operators disagree on how to align coordinates when two operands share a same-size dimension:

| Operator | Alignment |
|---|---|
| `*` | by label |
| `+`, `-`, `/` | by position — coordinate labels discarded (override join) |
| `<=`, `>=`, `==` (constraint creation) | by label |

Because some operators align by label and others by position, algebraically identical expressions produce different results.

### Reproducible Example

```python
import pandas as pd
import linopy

m = linopy.Model()
x = m.add_variables(coords=[["A1", "A5", "A11", "A100"]], name="x")
coef = pd.Series([100, 1, 5, 11], index=["A100", "A1", "A5", "A11"])
# coef["A1"] == 1  (coef's first entry is A100 == 100)

(x + coef).const.sel(dim_0="A1")     # 100   positional — should be 1
(x - coef).const.sel(dim_0="A1")     # -100  positional — should be -1
(x * coef).coeffs.sel(dim_0="A1")    # 1     by label   — correct
(x / coef).coeffs.sel(dim_0="A1")    # 0.01  positional — should be 1
```

### Broken equivalences

Mixing a positional operator with a by-label one makes identical forms disagree:

- `x - a <= 0`  ≠  `x <= a` — `-` positional, `<=` by label (#707).
- `x / a`  ≠  `x * (1 / a)` — `/` positional, `*` by label: `x / coef` pairs `x[A1]` with `1/100`; `x * (1/coef)` pairs it with `1/1`.

### Expected Behavior

Every operator should align coordinates the same way — by label — and a label mismatch should raise rather than silently pair by position.

### Context

Confirmed on linopy 0.7.0. Related: #586, #670, #707. The strict-alignment convention in #591 makes all operators align by label uniformly.

Issue #707, opened 2026-05-21
`x - a <= 0` and `x <= a` build different constraints


FBumann:
### Issue Description

`x - a <= 0` and `x <= a` are the same constraint, but linopy builds them differently in some cases. Subtraction aligns its operands **by position** when they share a same-size dimension — discarding coordinate labels — whereas the `<=` constraint operator aligns **by label**. The two forms disagree whenever the operands' labels are not already in the same order.

### Reproducible Example

```python
import pandas as pd
import linopy

m = linopy.Model()
x = m.add_variables(coords=[["A1", "A5", "A11", "A100"]], name="x")
coef = pd.Series([100, 1, 5, 11], index=["A100", "A1", "A5", "A11"])

print(x - coef <= 0)   # [A1]: x[A1] <= 100   <- positional, wrong
print(x <= coef)       # [A1]: x[A1] <= 1     <- by label, correct
```

`coef["A1"]` is `1`, so the correct constraint is `x[A1] <= 1`. `x <= coef` produces that; `x - coef <= 0` produces `x[A1] <= 100` — it paired `x[A1]` with `coef`'s first entry (`A100`), ignoring labels. Confirmed on linopy 0.7.0.

### Expected Behavior

`x - a <= 0` and `x <= a` are algebraically identical and must build the same constraint — either both align by label, or a label mismatch raises.

### Context

Same positional / override-join behaviour as #586 and #670. #586's own snippet (`add_constraints(x >= series)`) takes the label-aligned path and no longer reproduces, but the bug is alive in arithmetic. The strict-alignment convention in #591 would fix it.

FabianHofmann:
> [!NOTE]
> The following content was generated by AI (Claude Code) and checked against `master` (a731680). Posted as evidence only.

The specific discrepancy reported here no longer reproduces on current `master`: `x - a <= 0` and `x <= a` now build the **same** constraint (identical RHS). The remaining "operands should align by coordinate label rather than by position" concern is the broader v1-semantics work tracked in #714 (implementation in the still-open PR #717). Closing here as the reported inconsistency is resolved; the deeper alignment redesign continues under #714.

FabianHofmann:
Closing per the note above — the reported inconsistency is resolved on master; the broader alignment redesign continues in #714.

PR #715, opened 2026-05-21
test: pin alignment contract for variable/expression/constraint bugs

MaykThewessen:
## Summary

Adds `test/test_alignment_contract.py` with six `xfail(strict=True)` regression
tests that pin the *intended* coordinate-alignment contract reported across the
open alignment-bug cluster (#586, #706, #707, #708, #709).

The bugs all describe the same root issue: linopy's operators and
`add_variables` handle coordinate alignment inconsistently — some by label,
some positionally, some silently dropping dims. The discussion on each issue
keeps re-litigating *what the contract should be*. A failing test pins the
answer: every operator aligns by label, `add_variables` preserves `coords`
order and shape regardless of bound type.

## Issue → test mapping

| Issue | Test | Contract pinned |
|---|---|---|
| #707 | `test_subtract_then_le_matches_direct_le` | `x - a <= 0` and `x <= a` build the same constraint |
| #708 | `test_add_aligns_by_label` | `x + a` aligns by label, not by position |
| #708 | `test_subtract_aligns_by_label` | `x - a` aligns by label |
| #708 | `test_divide_aligns_by_label` | `x / a` aligns by label and `x / a == x * (1 / a)` |
| #706 | `test_add_variables_dataarray_bounds_preserve_coords_order` | `add_variables` dim order follows `coords` for `DataArray` bounds with missing dims |
| #709 | `test_add_variables_pandas_bounds_broadcast_missing_dim` | `add_variables` broadcasts pandas bounds to full `coords` shape |

**#586** is intentionally *not* included as its own test. Its original repro
(`m.add_constraints(x >= series)` with reordered labels) no longer reproduces
on `master` — the `add_constraints` path now label-aligns correctly. #707's
description already notes this: "#586's own snippet … takes the label-aligned
path and no longer reproduces, but the bug is alive in arithmetic." The
surviving arithmetic piece is covered by the #707 / #708 tests above, so
#586 is effectively merged into that coverage.

## Why xfail strict=True

Each test asserts the **expected** behaviour (the bug fixed), not the current
buggy output. As long as the bug reproduces the test xfails; the moment the
fix lands the test starts passing, `strict=True` turns the XPASS into a CI
failure, and the maintainer is forced to remove the `xfail` mark in the same
commit that closes the issue. That is the right state machine: the test
flips from "bug exists" to "regression guard" in one atomic step.

## Confirmed by running locally against `master` (commit `37af4ba`):

```
test/test_alignment_contract.py xxxxxx                                  [100%]
============================== 6 xfailed in 0.45s ===============================
```

## Notes for maintainers

- Please push back on any of the "expected behaviour" assertions before merge
  — pinning the contract is the point of this PR, so disagreement here is
  exactly the conversation we need. In particular: the divide test pins both
  "`/` aligns by label" *and* "`x / a == x * (1 / a)`". If you'd rather not
  commit to the algebraic identity, that line is easy to drop.
- The tests are deliberately small (≤30 lines each), share two minimal
  fixtures, and use only `pd.Series` / `xr.DataArray` — no solver call, no
  network I/O. They run in well under a second.
- If you'd prefer this content to live inside `test/test_constraints.py`'s
  existing `TestConstraintCoordinateAlignment` class or alongside
  `test/test_variables.py`, happy to relocate — kept it in a dedicated file
  to make the contract-pinning intent obvious.

Related issues: #586, #706, #707, #708, #709 (and #591, which proposes the
strict-alignment convention these tests would lock in).

### DIFF
--- a/test/test_alignment_contract.py
+++ b/test/test_alignment_contract.py
@@ -1,0 +1,241 @@
+#!/usr/bin/env python3
+"""
+Alignment-contract tests.
+
+Pin the *intended* contract for the coordinate-alignment bug cluster reported in
+PyPSA/linopy issues #586, #706, #707, #708, and #709.
+
+The contract this file locks in:
+
+1. Binary operators (``+``, ``-``, ``*``, ``/``) between a ``Variable``/expression
+   and a labelled coefficient (``pd.Series`` / ``xr.DataArray``) align operands
+   **by label**, never by position. Algebraically identical expressions
+   (``x - a <= 0`` vs ``x <= a``; ``x / a`` vs ``x * (1 / a)``) must therefore
+   build identical constraints / expressions.
+
+2. ``add_variables`` builds a variable whose dimension order matches ``coords``,
+   regardless of the *type* of the ``lower`` / ``upper`` bounds and regardless
+   of whether the bounds are missing some of the ``coords`` dimensions.
+   pandas bounds with missing dimensions must be broadcast to the full
+   ``coords`` shape, not silently dropped.
+
+Each parameter case is marked ``xfail(strict=True)`` so it fires as long as the
+bug reproduces; when the upstream fix lands the case starts passing, the strict
+xfail turns into ``XPASS`` and CI fails, forcing the maintainer to remove the
+xfail mark in the same commit that closes the issue. That is the right state
+machine for a contract test.
+
+The tests are deliberately parametrized over *labelled* coefficient / bound
+containers (``pd.Series``, ``pd.DataFrame``, ``xr.DataArray``) and over a couple
+of dtypes (``float64``, ``int64``) because each container reaches the alignment
+code via a different conversion path in linopy. Unlabelled inputs (``np.ndarray``,
+plain ``list``, scalar) are intentionally *not* parametrized: they have no
+labels for the contract to align against, so positional behaviour is correct
+for them and they would silently turn into XPASS noise.
+"""
+
+from __future__ import annotations
+
+import numpy as np
+import pandas as pd
+import pytest
+import xarray as xr
+
+import linopy
+from linopy import Model
+
+# ---------------------------------------------------------------------------
+# Coefficient parametrisation (operator-alignment tests, #707, #708)
+# ---------------------------------------------------------------------------
+#
+# Every case represents the *same* mapping ``{"A100": 100, "A1": 1, "A5": 5,
+# "A11": 11}`` in a different container and dtype. The variable below is built
+# in lexical order ``["A1", "A5", "A11", "A100"]``; the order mismatch is what
+# distinguishes label-alignment from positional-alignment. ``coef["A1"] == 1``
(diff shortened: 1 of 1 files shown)


FBumann:
@MaykThewessen thanks for the PR. Ill add a bunch of similar tests soon. Could you try to cover more datatypes etc with pytest.parameterize, while keeping the test themselves and the parameters readable?

MaykThewessen:
Thanks @FBumann — pushed an update (force-with-lease, `ce0db80` → `8ae16d7`) that parametrizes the six tests over labelled coefficient / bound containers and dtypes. 6 cases → 21.

Coverage matrix:

| Test | Issue | Parametrize over |
| --- | --- | --- |
| `test_subtract_then_le_matches_direct_le` | #707 | `pd.Series`, `xr.DataArray` × `float64`, `int64` (4) |
| `test_add_aligns_by_label` | #708 | same (4) |
| `test_subtract_aligns_by_label` | #708 | same (4) |
| `test_divide_aligns_by_label` | #708 | same (4) |
| `test_add_variables_dataarray_bounds_preserve_coords_order` | #706 | `xr.DataArray` × `float64`, `int64` (2) |
| `test_add_variables_pandas_bounds_broadcast_missing_dim` | #709 | `pd.Series` × `float64`, `int64`, `pd.DataFrame` × `float64` (3) |

All 21 parameter cases still `xfail(strict=True)` against master — confirmed locally:

```
test/test_alignment_contract.py ......x...x...x...x...x...x...x...x...x...x...x...x...x...x...x...x...x...x...x...x...x [100%]
============================= 21 xfailed in 1.29s ==============================
```

Readability choices:

- Shared module-level constants (`_COEF_LABELS`, `_VAR_LABELS`, `_X`, `_Y`) so each parameter case is built by a one-line helper (`_series("float64")`, `_dataarray("int64")`, …) — the parameter IDs read as `series-float64` / `dataarray-int64` / `dataframe-float64` rather than the auto-generated repr soup.
- `xfail` mark stays on the test, not on each parameter case — every dtype/container combination reproduces the bug today, so a per-case mark would just be six identical reasons. The moment the upstream fix lands, all cases for a given test will XPASS together and the strict-mark forces removal in the same commit.
- Unlabelled inputs (`np.ndarray`, `list`, scalar) are intentionally *not* parametrized — they have no labels for the contract to align against, so positional behaviour is correct for them and they would silently turn into XPASS noise. I left a note in the module docstring explaining this so it's not re-litigated in review.

Happy to widen further (e.g. `xr.Dataset`, ND `DataArray`, mixed dtypes via `complex128` / `Int64` extension) if you'd like — wasn't sure where the readability budget ran out. Also happy to fold these into your follow-up suite if you'd rather keep this PR minimal and land the broader parametrization in your own commit.

FBumann:
@MaykThewessen Feel free to share your thoughts on #717

FBumann:
THanks for the PR.
Those are now folded in and superseeded in #722 and #717. I validated that those are caught there.

Issue #703, opened 2026-05-20
Extend xarray method coverage on Variable / LinearExpression


FBumann:
## Summary

`Variable`, `LinearExpression`/`QuadraticExpression` and `Constraint` each expose a hand-picked subset of `xarray.Dataset` methods (via `varwrap` / `exprwrap` / `conwrap`). The three lists have drifted apart, several common arithmetic methods are missing entirely, and a few existing ones diverge from xarray — especially around datetime indexes.

This issue collects the analysis and a design sketch for closing the gaps. All of the proposed methods preserve linearity (none multiplies two variables), so they are mostly bookkeeping.

---

## A. Cross-class inconsistencies (existing methods)

The same wrapper is exposed on some classes but not others, with no clear reason:

| method | Variable | Expression | Constraint |
|---|---|---|---|
| `compute` | yes | no | no |
| `chunk` (dask) | no | yes | yes |
| `reindex` / `reindex_like` | no | yes | yes |
| `astype` | no | yes | no |
| `drop` / `drop_vars` | no | yes | no |
| `reset_index` | no | yes | no |
| `rename_dims` | no | yes | yes |
| `shift` fill_value default | yes (`variables.py:1268`) | no (`expressions.py:1474`) | no |

The two that actually bite users:

- **`Variable.reindex` / `reindex_like` missing** — you can reindex an expression or constraint onto a master index but not a variable. Natural operation for datetime work (aligning a variable to a full snapshot index).
- **`Expression.shift` / `Constraint.shift` have no `fill_value`** while `Variable.shift` defaults to the linopy fill value. Shifted slots get `coeffs=NaN`, `vars=-1`; it mostly survives downstream `fillna(0)`, but it is the root of the `diff` issue below.

## B. Existing methods lacking features vs. xarray (datetime focus)

1. **`diff` (`expressions.py:1238`)** — implemented as `self - self.shift({dim: n})`. Two divergences from `xarray.diff`:
   - **Length**: xarray trims the dimension to `N-n`; linopy keeps `N`, so the first `n` rows are spurious (`self[0] - empty`, a leading garbage term still carrying the original coordinate label).
   - **No `label` argument** — xarray has `label="upper"|"lower"`; linopy is hard-wired to `"upper"`.
[...]

PR #627, opened 2026-03-18
Treat NaN as absent in v1 arithmetic

FBumann:
## Summary

- NaN in multiplicative/divisive constants now **masks** the term (makes it absent) instead of raising `ValueError`
- NaN in additive/subtractive constants is treated as 0 (additive identity)
- NaN in constraint RHS still raises `ValueError` — silently skipping constraints is too dangerous
- `Variable.to_linexpr(nan_coeff)` now correctly produces absent slots in v1 (was inconsistently filling with 0)
- New `mixed-coordinate-arithmetic.ipynb` notebook documenting patterns for combining variables with different coordinate subsets (from [PR #591 feedback](https://github.com/PyPSA/linopy/pull/591#issuecomment-4082181622))

## Motivation

Addresses @brynpickering's feedback in #591: the `fillna(0)` ceremony before every multiplication with incomplete cost data was verbose and error-prone. With NaN-as-mask, the pre-align pattern becomes clean:

```python
# Before (NaN raises): fillna required everywhere
cap_a_al, cap_b_al, cost_a_al, cost_b_al = linopy.align(
    cap_a, cap_b, cost_a.fillna(0), cost_b.fillna(0), join="outer"
)
combined = cap_a_al * cost_a_al + cap_b_al * cost_b_al

# After (NaN masks): no fillna needed
cap_a_al, cap_b_al, cost_a_al, cost_b_al = linopy.align(
    cap_a, cap_b, cost_a, cost_b, join="outer"
)
combined = cap_a_al * cost_a_al + cap_b_al * cost_b_al
```

## NaN behavior summary (v1)

| Operation | NaN behavior |
|---|---|
| `expr + nan` | NaN → 0 (additive identity) |
| `expr * nan` | NaN → absent (masks term) |
| `expr / nan` | NaN → absent (masks term) |
| `expr <= nan` | `ValueError` (use `.sel()`/`mask=`) |

## Test plan

- [ ] All 5333 existing tests pass
- [ ] Verify notebooks execute: `arithmetic-convention`, `missing-data`, `mixed-coordinate-arithmetic`
- [ ] Review algebraic property consistency (NaN masking preserves commutativity, distributivity)
- [ ] Check edge cases: scalar NaN, array NaN, Variable vs Expression paths

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #717, opened 2026-05-21
feat: v1 semantic convention

### LINKED ISSUE #256: Improve documentation on overriding coordinates feature
Hi,

I'm currently stuck trying to archive the following rather specific problem in linopy:
```
import linopy
m = linopy.Model()

# 2 days, hourly (48 hours)
t = list(range(1,49))
n = ["a", "b", "c", "d"]

level = m.add_variables(coords=[t, n], name="level", dims=["time", "node"])

change = m.add_variables(coords=[t, n], name="change", dims=["time", "node"])

# previous_level should by cyclic per day, so have time indices
#  24, 1, 2, ..., 23, 48, 25, 26, ..., 47
# |---- 1st day ----|----  2nd day   ----|
# constraint should be valid for each node
level == previous_level + change
```
So problem is in the creation of `previous_level`: I don't know have to do that. I know there is `Variable.roll()`  for creating a shift in specific dimensions, but I basically want this operation but restricted to a subset of the time index. My plan was to use the result of `roll()` and assign the correct variable value to the positions `0` and `24` myself, but assignment does not seem to work. Is there any way to create a self-defined variable vector that has the form of the result of `roll()` (single variables per coordinate tuple), but using an arbitrary assignment?

PS: My fallback solution would be to create single constraints per point in time, but I would like to avoid that if possible ^^

### LINKED ISSUE #257: Inconsistent coordinate override in arithmetic expressions
Based on the discussion in #256, I started using `<variable>.loc[indices]` in arithmetic expressions and observed some inconsistencies (or I still do not understand the behaviour to 100% ^^)

Following setup:
```
import xarray as xr

import linopy

m = linopy.Model()

# 2 days, hourly (48 hours)
t = [0, 1, 2, 3]
n = ["a", "b"]

level = m.add_variables(coords=[t, n], name="level", dims=["time", "node"])

alpha = xr.DataArray([[1, 2], [3, 4], [5, 6], [7, 8]], coords=level.coords)

other_times = [2, 0, 3, 1]
other_time_level = level.loc[dict(time=other_times)]
```

**Some arithmetic expressions**
`level + other_time_level` works as expected, coordinates are overridden based on `level`
```
LinearExpression (time: 4, node: 2):
------------------------------------
[0, a]: +1 level[0, a] + 1 level[2, a]
[0, b]: +1 level[0, b] + 1 level[2, b]
[1, a]: +1 level[1, a] + 1 level[0, a]
[1, b]: +1 level[1, b] + 1 level[0, b]
[2, a]: +1 level[2, a] + 1 level[3, a]
[2, b]: +1 level[2, b] + 1 level[3, b]
[3, a]: +1 level[3, a] + 1 level[1, a]
[3, b]: +1 level[3, b] + 1 level[1, b]
```

When using `level + alpha * other_time_level`, I would have expected to get
```
LinearExpression (time: 4, node: 2):
------------------------------------
[0, a]: +1 level[0, a] + 1 level[2, a]
[0, b]: +1 level[0, b] + 2 level[2, b]
[1, a]: +1 level[1, a] + 3 level[0, a]
[1, b]: +1 level[1, b] + 4 level[0, b]
[2, a]: +1 level[2, a] + 5 level[3, a]
[2, b]: +1 level[2, b] + 6 level[3, b]
[3, a]: +1 level[3, a] + 7 level[1, a]
[3, b]: +1 level[3, b] + 8 level[1, b]
```
since I though that `alpha` would be term to define the coordinates in `alpha * other_time_level`

However, what I actually get is the following (coordinates of `other_time_level` are sorted according to the original time index again)
```
LinearExpression (time: 4, node: 2):
------------------------------------
[0, a]: +1 level[0, a] + 1 level[0, a]
[0, b]: +1 level[0, b] + 2 level[0, b]
[1, a]: +1 level[1, a] + 3 level[1, a]
[1, b]: +1 level[1, b] + 4 level[1, b]
[2, a]: +1 level[2, a] + 5 level[2, a]
[2, b]: +1 level[2, b] + 6 level[2, b]
[3, a]: +1 level[3, a] + 7 level[3, a]
[3, b]: +1 level[3, b] + 8 level[3, b]
```

Interestingly, the result changes when transforming `other_time_level` into a `LinearExpression` first, i.e., call `level + alpha * other_time_level.to_linexpr()`
```
LinearExpression (time: 4, node: 2):
------------------------------------
[0, a]: +1 level[0, a] + 5 level[2, a]
[0, b]: +1 level[0, b] + 6 level[2, b]
[1, a]: +1 level[1, a] + 1 level[0, a]
[1, b]: +1 level[1, b] + 2 level[0, b]
[2, a]: +1 level[2, a] + 7 level[3, a]
[2, b]: +1 level[2, b] + 8 level[3, b]
[3, a]: +1 level[3, a] + 3 level[1, a]
[3, b]: +1 level[3, b] + 4 level[1, b]
```

Finally, I tried to simply add `alpha` instead of multiplying it, i.e., `level + alpha + other_time_level` and this gives the expected result in terms of sorting the single terms into the expected rows:
```
LinearExpression (time: 4, node: 2):
------------------------------------
[0, a]: +1 level[0, a] + 1 level[2, a] + 1
[0, b]: +1 level[0, b] + 1 level[2, b] + 2
[1, a]: +1 level[1, a] + 1 level[0, a] + 3
[1, b]: +1 level[1, b] + 1 level[0, b] + 4
[2, a]: +1 level[2, a] + 1 level[3, a] + 5
[2, b]: +1 level[2, b] + 1 level[3, b] + 6
[3, a]: +1 level[3, a] + 1 level[1, a] + 7
[3, b]: +1 level[3, b] + 1 level[1, b] + 8
```

So my question would be: How to achieve the intended result and is all of the behaviour as it should be?

### LINKED ISSUE #295: Auxiliary non-dimension coordinates leak into expressions and break alignment
Hi people, I am currently struggling with xarray and linopy. 

I have a dimension A whose coordinates can be grouped. For me it was natural to have a dimension A with coordinates for A but also coordinate B that is built on dimension A and then swap dimensions but I failed to do it. 

So here is an example.
dimension A has coordinates: [1, 2, 3, 4, 5, 6] but coordinates [1, 2, 3] belong to group 311 and rest of them to group 322. I would like to create constraints based on those 2 groups. 

I have a variable with this situation:
```
var.coords
Coordinates:
   * A (A) int64 48B 1 2 3 4 5 6
     B (A) int64 48B 311 311 311 322 322 322
```

Looking at xarray docs this may be the way `var.swap_dims({'A': 'B'})` and then I can do sum on other dimensions and set constraint for groups. But Variable doesn't have that method.

Next I tried with groupby on B, I got correct groups but then sum on groupby expression will throw an error that B is already present...

In the end I made B full dimension (I am breaking normalization from Databases theory :D) but looks like suboptimal solution and I have this kind of case for multiple coordinates so instead of having small dimensional case of 7 dimensions I will end with at least 9 

Could I do something better?

### LINKED ISSUE #570: Inconsistent behavior when multiplying with subset coordinates

### LINKED ISSUE #586: Constraints do not match RHS based on coordinates

### LINKED ISSUE #670: bug when adding two linopy expressions/variables that share dimension names but have disjoint label coordinates of same size

- [ ] I have confirmed this bug exists on the lastest [release](https://github.com/pypsa/linopy/releases) of Linopy.

### Issue Description

when adding two expressions sharing dimension but with different coords, the broadcasting does not work as expected when the number of elements in coords are identical, and works if they are different.

### Reproducible Example

```python
import pandas as pd
import linopy

m = linopy.Model()

time = pd.date_range("2025-01-01", periods=3, freq="h")

var_flow1 = m.add_variables(lower=0, coords=[time, ["pv", "one"]])
var_flow2  = m.add_variables(lower=0, coords=[time, ["wind"]])
var_flow3  = m.add_variables(lower=0, coords=[time, ["wind","two"]])

# create an expression with second coord ["pv", "one", "wind"]
ok = var_flow1 + var_flow2
print(ok)
# BUG: silently map wind to pv and two to one instead of taking the union, 
# giving wrong coords ["pv", "one"] instead of ["pv", "one", "wind", "two"]
buggy = var_flow1 + var_flow3
print(buggy)
```

### Expected Behavior

var_flow1 + var_flow3 should return an expression with dim_1 of size 4 equal to ["pv", "one", "wind", "two"]

### Installed Versions

<details>
highspy==1.14.0
linopy==0.6.7
xarray==2026.4.0
</details>

### LINKED ISSUE #708: Operators align coordinates inconsistently (`*` by label, `+` `-` `/` by position)

### LINKED ISSUE #711: Subset constants break associativity: a + factor + b ≠ a + b + factor
### Issue Description

When a constant has a *different* size on a dimension it shares with a variable/expression, linopy left-joins it — reindexing onto the other operand and dropping whatever does not fit. Because the join drops coordinates, the result depends on the order the terms are written: addition is no longer associative.

### Reproducible Example

```python
import pandas as pd
import xarray as xr
import linopy

m = linopy.Model()
a = m.add_variables(coords=[pd.RangeIndex(3, name="time")])  # time 0..2
b = m.add_variables(coords=[pd.RangeIndex(5, name="time")])  # time 0..4
factor = xr.DataArray([10, 20, 30, 40, 50], dims=["time"], coords={"time": range(5)})

print((a + factor + b).const.sel(time=3).item())  # 0.0  — factor dropped
print((a + b + factor).const.sel(time=3).item())  # 40.0 — factor kept
```

`a + factor + b` left-joins `factor` onto `a` (time 0..2) first, losing time 3–4; `a + b + factor` joins `a + b` (time 0..4) first and keeps it.

### Expected Behavior

Addition is associative — the result must not depend on term order. Either every order agrees, or the size mismatch raises.

Confirmed on linopy 0.7.0. Part of the arithmetic-convention work (#591).

### LINKED ISSUE #712: An absent variable is indistinguishable from a zero variable
### Issue Description

Structural operations — `.shift()`, `.where()`, `mask=` — leave *absent slots*: coordinates where the variable does not exist. linopy does not mark these as absent once the variable enters an expression, so after multiplication an absent slot is indistinguishable from a genuine zero-valued term, and `fillna()` cannot reach it.

### Reproducible Example

```python
import pandas as pd
import linopy

m = linopy.Model()
x = m.add_variables(coords=[pd.RangeIndex(5, name="time")])

xs = x.shift(time=1)  # time=0 is now absent
print((xs * 3).isnull().values[0])  # False — the absent slot looks present
```

After `xs * 3` the slot at `time=0` reports `isnull() == False` — a present, 0-valued term — so it cannot be told apart from a real zero term.

### Expected Behavior

An absent slot should stay absent through multiplication (`isnull() == True`), so absent and zero remain distinguishable.

Confirmed on linopy 0.7.0. Part of the arithmetic-convention work (#591).

### LINKED ISSUE #713: NaN in user-supplied data is silently filled instead of surfaced
### Issue Description

When user-supplied data contains NaN — usually a sign of missing or bad input — linopy silently fills it with an operator-specific neutral element rather than surfacing it. The fill value differs per operator, and in every case a data error that should stop the run is absorbed without a trace.

### Reproducible Example

```python
import numpy as np
import pandas as pd
import xarray as xr
import linopy

m = linopy.Model()
x = m.add_variables(coords=[pd.RangeIndex(3, name="time")])
data = xr.DataArray([1.0, np.nan, 3.0], dims=["time"], coords={"time": range(3)})

print((x + data).const.sel(time=1).item())             # 0.0 — NaN -> 0
print((x * data).coeffs.squeeze().sel(time=1).item())  # 0.0 — NaN -> 0 (variable zeroed)
print((x / data).coeffs.squeeze().sel(time=1).item())  # 1.0 — NaN -> 1 (variable unchanged)
```

The NaN at `time=1` is replaced — `0` for `+`, `0` for `*` (zeroing the variable out), `1` for `/` (leaving it unchanged). No exception, no warning.

### Expected Behavior

NaN in user-supplied data should not be silently substituted. The user should say what it means (`data.fillna(0)`, `factor.fillna(1)`, …) — or linopy should raise.

Confirmed on linopy 0.7.0. Part of the arithmetic-convention work (#591).

### LINKED ISSUE #714: Meta: predictable arithmetic — coordinate alignment and NaN handling
Tracking issue for the known problems with linopy's arithmetic — coordinate alignment and NaN handling — and the v1 convention that fixes them.

linopy aligns coordinates with size-based heuristics. When operands don't line up it guesses, and the guesses are sometimes **silently wrong** — no exception, the solver returns `Optimal`, the objective looks plausible. PyPSA/PyPSA#1677 shipped a workaround for one such case (a capacity-expansion result 47% too cheap).

## Bugs — silent wrong answers

- [ ] Operators align coordinates inconsistently — #708 (root cause), #707, #670
- [ ] Subset constants break associativity — #711
- [ ] An absent variable is indistinguishable from a zero variable — #712
- [ ] Non-dimension coordinates leak into expressions — a conflicting auxiliary coordinate is silently kept, not raised — #295

## Opinionated behaviour — silent, debatable choices

- [ ] NaN in user-supplied data is silently filled — #713
- [ ] Multiplying by a subset constant silently drops variables — #570

## Already fixed in v0.7.0

- #569 — subset-multiplication crash / Variable-vs-Expression inconsistency — closeable.

## Out of scope

Predictability issues, but not arithmetic-alignment bugs:

- `add_variables` mishandles bounds with missing dimensions — #706 (`DataArray` bounds → wrong dimension order), #709 (`Series`/`DataFrame` bounds → dimension dropped). #450 is the original report, only partially fixed by #614.
- #351 — `groupby` by a `DataArray`; #308 — `.loc` endpoint inclusivity.

#703 (xarray method coverage) consumes the convention rather than constraining it.

## The fix

A strict **v1 arithmetic convention** behind `linopy.options["arithmetic_convention"]` — align by label, raise on a mismatch, NaN means "absent term". It ships opt-in, becomes the default, and legacy is removed at linopy 1.0. One PR carries the design goals, the spec, tests and code; docs follow. Implementation: #591. Milestone: [Predictable Arithmetics](https://github.com/PyPSA/linopy/milestone/3).

### LINKED ISSUE #736: v1 convention: pair unlabeled operands with dimensions by size, raise on ambiguity
Part of the v1 convention ([#714], spec in [`arithmetics-design/convention.md`](https://github.com/PyPSA/linopy/blob/feat/arithmetic-convention/arithmetics-design/convention.md), introduced by #717). The rule is **specified but not yet implemented** — this issue tracks the implementation.

## The rule

Unlabeled operands — numpy arrays, lists, polars Series — carry no labels to align by, so they pair with the linopy operand's dimensions **by size**: each axis adopts the dimension (and the coordinates) whose length matches. The pairing must be determined by the sizes alone — when it would be a guess, raise.

## Spec vs. today

With `x(time: 5)`, `rect(a: 4, time: 5)`, `sq(a: 4, b: 4)`:

| Operation | Today (leading-positional) | v1 spec (size-determined) |
|---|---|---|
| `np.arange(5.) + x` | maps to `time` | maps to `time` (unique match) |
| `np.arange(4.) + x` | raises (opaque xarray error) | raises with a linopy message (no size match) |
| `np.arange(4.) * rect` | maps to `a` | maps to `a` (unique match) |
| `np.arange(5.) * rect` | raises | **maps to `time`** (unique match, not the leading dim) |
| `np.arange(4.) * sq` | **silently maps to `a`** | **raises — ambiguous** (`a` or `b`?) |
| `np.arange(16.).reshape(4,4) * sq` | maps to `(a, b)` in order | **raises — ambiguous** (sizes cannot tell `(a, b)` from `(b, a)`) |
| `[1., 2., 3., 4.] + x` | legacy: silent left-join / v1: coord-mismatch raise | raises with a linopy message (no size match) |

The fix for an ambiguous case is to name the dimensions: wrap the array in an `xarray.DataArray` (or use a pandas Series/DataFrame with a named index).

## Implementation notes

- The pairing logic belongs at the constant-operand gateway (the `as_dataarray(other, coords=self.coords, dims=self.coord_dims)` call sites in `expressions.py` / `variables.py`), so it should build on the coords-as-truth / `as_dataarray` seam refactored in #732 — **blocked by #732**.
- Legacy semantics keep leading-positional pairing but must warn wherever v1 would raise (transitioning goal: no silent change on opt-in).
- Test home: Slice H (`TestObjectScope`) in `test/test_legacy_violations.py` — the ambiguity raises are genuine legacy violations to pair with `@pytest.mark.legacy` documentation tests.

[#714]: https://github.com/PyPSA/linopy/issues/714

FBumann:
## The strict **v1 semantic convention** for linopy — predictable coordinate alignment and NaN handling.

This is the master PR for the new semantic convention in linopy. It starts with our [**`Design & transitioning goals`**](https://github.com/PyPSA/linopy/blob/feat/arithmetic-convention/arithmetics-design/goals.md), which is carried out in our [**`New Semantics spec`**](https://github.com/PyPSA/linopy/blob/feat/arithmetic-convention/arithmetics-design/convention.md). Both files are tracked in this branch. WHat you read is the current state.

_Both might change until this PR is merged_

The concrete rollout checklist — the three stages (opt-in → default → 1.0) and their per-stage status — is tracked in [**`arithmetics-design/open-items.md`**](https://github.com/PyPSA/linopy/blob/feat/arithmetic-convention/arithmetics-design/open-items.md).

### Scope

The convention ships behind `linopy.options["semantics"]` — `v1` opt-in, `legacy` the default. This PR carries the design, spec, tests and implementation; documentation notebooks follow separately.

### Testing

All tests in linopy will be executed for both semantics.
Differing behaviour will be tested using `pytest.markers`.

_This will increase ci time temporarily until v1 is released._

### Defered to a follow up PR

Docs: Documentation, Migration guide, Release notes

---

> [!NOTE]
> Issue links below added by AI. These are the milestone issues the v1
> convention fixes; they auto-close when this PR merges. #703 (extend xarray
> method coverage) stays open — it is a deferred follow-up (§13), not fixed here.

Closes #714
Closes #713
Closes #712
Closes #711
Closes #708
Closes #670
Closes #586
Closes #570
Closes #295
Closes #257
Closes #736
Closes #256


FBumann:
**Coming next**, in order:

- **Slice F — §11 auxiliary-coordinate conflicts.** Raise on non-dimension-coord conflict during alignment (covers #295). Small.
- **Slice G — §13 reductions audit.** Most reductions already use `skipna=True`; need a focused audit + tests for `sum` / `mean` / `groupby` / `resample` / `coarsen` and the objective. Likely mostly tests.
- **Slice P — `linopy/piecewise.py` and `linopy/sos_reformulation.py`.** Internal callers build expressions by `.isel(...)`-slicing along a piece dim and comparing the two slices (`delta_hi <= delta_lo`); the slices share the dim with different coords, which v1 §8 rejects. Until this lands, `test_piecewise_constraints.py` + `test_piecewise_feasibility.py` carry a module-level `pytestmark = pytest.mark.legacy` and two SOS2 tests are method-marked legacy. Fixing piecewise removes those marks.

Then a final pass on docs (user-facing migration / rollout — deferred so far).

MaykThewessen:
Thanks @FBumann — read `goals.md` + `convention.md` + the [#714] catalogue. Priority ordering (no-silent-wrong-answers > algebraic-laws > absence-first-class > least-surprise) reads exactly right for an optimisation library. Notes from the PyPSA-Eur side, ordered by leverage:

### Highest leverage: lock the §5 / [#627] question before v1 GAs

§5 ("user NaN raises") is the safe default, but if [#627] later flips to "user NaN means absent", code written defensively for strict-v1 — peppering `fillna(...)` at construction sites — will quietly behave differently than the same code under v1-with-#627. Two semantic flips inside the v1 lifecycle costs trust. Worth deciding §5 before v1 promotes to default, even at the cost of slipping the schedule. My vote: keep raise — in PyPSA-Eur, NaN in user data is almost always an index-mismatch bug, and goal 1 is the whole point.

### Biggest migration cost is §8, not §5

PyPSA-Eur builds expressions against `n.snapshots[mask]` subsets and currently relies on size-coincidence positional behaviour. To make the legacy → v1 step survivable on a 50k-LOC codebase, the warning needs to identify the exact mismatch: `(dim, missing_labels[:5])`, not just "shared dim mismatch". And — bigger ask — a `linopy.diagnose_alignment(model)` helper that walks every stored expression and lists every site v1 would raise, runnable under legacy, would let users patch one site at a time before flipping the switch.

Happy to PR this helper once the spec stabilises.

### §10 needs an operator-syntax escape hatch

`join=` lives on named methods only. A scoped context manager — `with linopy.alignment(join="override"): ...` — would let people refactor incrementally without rewriting every `a + b` into `a.add(b, join="override")`. Especially useful where the positional override is intentional (some legacy PyPSA stats code). Scoped, not global; threadlocal.

Happy to PR this too.

### Spec gaps worth nailing down

- **§13 — `mean` over an all-absent dim is 0/0.** Pick one: result is absent (NaN propagates) or raise. My vote: absent, consistent with §6's "absence absorbs".
- **§5 — does the rule cover bounds and RHS, not just `coeffs` / `const`?** `add_variables(lower=series_with_nan)` today silently broadcasts. Under v1 I assume it raises; spec should say so. Same for constraint RHS.
- **§11 — error message on aux-coord conflict should print both values** (or `.head()` thereof), not just the coord name. We've been bitten by [#295] in PyPSA-Eur post-processing where a carrier aux coord silently disappears after `expr.sum("snapshot")`; the diagnostic matters.

### On my #715

Once v1 is opt-in, the six xfail-strict tests flip to XPASS under `options["semantic_convention"]="v1"`. Happy to parametrize the file over `legacy` / `v1` so it tracks the rollout schedule rather than only gating on the eventual default — turns #715 into a v1 conformance harness.

MaykThewessen:
Lovely commit @FBumann — two upgrades I didn't ask for that I should have:

- Per-op message split (`+`/`-` → 0, `*` → 0, `/` → 1) puts the #713 asymmetric fill in the warning text. Users can grep CI logs by `op_kind`.
- `_legacy_masked_variable_message` in `Variable.to_linexpr` catches the `2 * x + y` (masked `y`, no NaN constant) case where no other warn site fires. That's the most common PyPSA-Eur divergence and would have migrated silently. Excellent.

Three follow-up suggestions, ordered by leverage:

1. **Include `(dim, missing_labels[:5])` in coord-mismatch text.** `_legacy_coord_mismatch_message("merge along dim 'time'")` names the dim but not the diff. In a large codebase the dim alone isn't enough to locate the wrong reindex. The `_shared_dim_mismatch_message` v1-raise formatter already prints both sides — could share it between v1-raise and legacy-warn so the strings stay aligned.
2. **Stacklevel sanity check.** Current mix is 3/4/5 across call sites. Worth a test that pins the user-visible frame to the caller's source, not linopy internals — IDE jump-to-source on the warning depends on it.
3. **`diagnose_alignment(model)` audit helper still distinct from warnings.** Warnings fire only when ops run; the helper walks every stored expression in a built model and lists every v1-raise site, runnable under legacy. Different shape — lets users audit before opting in. Happy to PR if you want it; or if you'd rather own it, glad to leave it to you.

linopy/expressions.py
```diff
@@ -2384,6 +2563,33 @@ def merge(
...
     model = exprs[0].model
 
+    data = [e.data if isinstance(e, linopy_types) else e for e in exprs]
+    data = [fill_missing_coords(ds, fill_helper_dims=True) for ds in data]
+
+    # §8: shared *user* dimension coordinates must match exactly across all
+    # operands. Helper dims (_term, _factor) legitimately differ, so we
+    # validate user dims separately and keep xr.concat on join="outer"
+    # (which doesn't enforce "exact" — that's what this check is for).
+    if join is None:
```
MaykThewessen, line 3127: **§8 / §11 enforcement gated behind `if join is None`** — when the caller passes an explicit `join=` (any of `"exact"`, `"override"`, `"inner"`, `"outer"`, `"left"`, `"right"`) the `merge_shared_user_coord_mismatch` and `conflicting_aux_coord` checks below are skipped entirely.

Per §11 the aux-coord conflict raise should fire **regardless** of how the dim mismatch is resolved — `xr.concat(..., compat="override")` silently drops conflicting aux coords and that's exactly the #295 bug v1 is supposed to close. Suggest pulling the aux-coord check out of the `if join is None:` block so it runs on every path.

linopy/expressions.py
```diff
@@ -556,32 +575,112 @@ def _align_constant(
             Whether the expression's data needs reindexing.
         """
         if join is None:
-            if other.sizes == self.const.sizes:
-                return self.const, other.assign_coords(coords=self.coords), False
+            # §11: silently dropping a conflicting aux coord is what
+            # xarray does by default — v1 raises, legacy warns.
+            aux_conflict = conflicting_aux_coord([self.const, other])
```
MaykThewessen: Same gating problem as `merge` (see comment on line 2573): `conflicting_aux_coord([self.const, other])` runs only inside `if join is None:`. A user calling `expr.add(constant, join="override")` bypasses §11 aux-coord enforcement entirely. Aux-coord conflict is independent of dim alignment — should fire on every v1 binary op.

linopy/expressions.py
```diff
@@ -556,32 +575,112 @@ def _align_constant(
...
+                    _legacy_coord_mismatch_message("this operator's constant operand"),
+                    stacklevel=4,
+                )
+                return (
+                    self.const,
+                    other.reindex_like(self.const, fill_value=fill_value),
+                    False,
+                )
+
+        if join == "override":
```
MaykThewessen, line 1028: **`override` semantics drift.** `_shared_dim_mismatch_message` (and `convention.md` §10) document `override` as "positional alignment, made explicit". The implementation here is `other.assign_coords(coords=self.coords)` — that's *coord-renaming*, not positional alignment, and has no shape/size guarantee.

If `other.sizes != self.const.sizes` on a shared dim, `assign_coords` will silently broadcast (or raise an opaque xarray error). The legacy positional path at least gated on `other.sizes == self.const.sizes` (line ~593) before doing the rename. The v1 `override` branch should either keep that gate or rename itself to `relabel`/`coerce` to be honest about what it does.

FBumann:
@MaykThewessen Im not sure how such a `diagnose_alignment(model)` would work, as it might not be clear from a build constraint if the creation path that lead to it would have produced a different constraint in the new semantic convention. 
However, feel free to PR this.

linopy/expressions.py
```diff
@@ -2417,7 +2620,13 @@ def merge(
     if dim == TERM_DIM:
         ds = xr.concat([d[["coeffs", "vars"]] for d in data], dim, **kwargs)
         subkwargs = {**kwargs, "fill_value": 0}
-        const = xr.concat([d["const"] for d in data], dim, **subkwargs).sum(TERM_DIM)
+        # Under v1, §6 requires that an absent slot in any operand stays
+        # absent in the result; ``sum(skipna=False)`` propagates NaN
+        # rather than collapsing it to 0.
+        skipna = not is_v1()
+        const = xr.concat([d["const"] for d in data], dim, **subkwargs).sum(
+            TERM_DIM, skipna=skipna
```
MaykThewessen, line 3163: **§6 fix here is correct but doesn't cover the FACTOR_DIM branch right below.** Lines 2633–2634 (`prod(FACTOR_DIM)` on `coeffs` and `const`) use xarray default `skipna=True`, so an absent factor in a quadratic (`expr * expr`, `var * var`) silently becomes identity `1` and the product is non-absent.

Same `skipna = not is_v1()` treatment should apply to both `prod` calls — or §6 propagation only holds for `+`/`-` and breaks on `*`.

linopy/semantics.py
```diff
@@ -0,0 +1,310 @@
...
+        " propagates through arithmetic instead (`x + y` becomes absent"
+        " at the slot and the constraint drops)."
+        f"\n  Resolve:   wrap with `{name}.fillna(0)` for the legacy"
+        " behaviour under v1"
+        "\n             (no fix needed if you only use the variable in a"
+        " constraint LHS alone — `y >= 0` drops the same way in both)." + _OPT_IN_HINT
+    )
+
+
+def warn_legacy(message: str, *, stacklevel: int = 3) -> None:
```
MaykThewessen: **`stacklevel=3` default lands inside linopy internals on every call site I traced.** Frame walks from the three call sites:

- `expressions.py:1292` (`to_constraint` legacy) → `warn_legacy(..., stacklevel=3)` lands in `__le__`/`__ge__`/`__eq__`, not user code. Needs 4.
- `expressions.py:2579, 2591` (`merge` legacy) → lands in `merge`'s caller (`add`/`__add__`), still linopy. Needs 4 or 5 depending on path.
- `variables.py:352` (`to_linexpr` legacy) → lands in `__add__`/`__mul__` (linopy). Needs 4 or 5.

Single hardcoded value cannot land on user code for every public-API path. Suggest: either thread `stacklevel` from the public-API entry point (each `__op__` / `add` / `merge` passes its own depth in), or use `warnings.warn(stacklevel=…)` after walking the stack to skip linopy frames. The current scheme makes IDE jump-to-source on the warning point at linopy internals.

Worth a small test that pins user-visible frame for each path: build a tiny model in `test/conftest.py`, capture the warning, assert `filename` ends with the test file's basename.

linopy/semantics.py
```diff
@@ -0,0 +1,310 @@
...
+
+def check_user_nan_scalar(*, op_kind: str = "add") -> None:
+    """
+    Enforce §5 for a scalar: v1 raises, legacy warns with op-specific text.
+
+    ``op_kind`` is one of ``"add"`` (covers +/-), ``"mul"``, ``"div"``.
+    """
+    if is_v1():
+        raise ValueError(_user_nan_message())
+    warn_legacy(_legacy_message_for_op(op_kind), stacklevel=5)
```
MaykThewessen: `stacklevel=5` is also off for the user-NaN path. Call chain: `warn_legacy` → `check_user_nan_*` → `_add_constant_legacy`/`_apply_constant_op_legacy` → `_add_constant`/`_apply_constant_op` dispatcher → `__add__`/`add` → user. That's 6 frames from `warn_legacy.warn(...)` to the user — `stacklevel=5` lands at `__add__`/`add` (linopy), not user code.

Affects both `check_user_nan_scalar` (line 200) and `check_user_nan_array` (line 207).

linopy/expressions.py
```diff
@@ -1113,19 +1286,16 @@ def to_constraint(
                 )
             rhs = rhs.reindex_like(self.const, fill_value=np.nan)
 
-        # Remember where RHS is NaN (meaning "no constraint") before the
-        # subtraction, which may fill NaN with 0 as part of normal
-        # expression arithmetic.
         if isinstance(rhs, DataArray):
             rhs_nan_mask = rhs.isnull()
+            if rhs_nan_mask.any():
+                warn_legacy(_legacy_nan_rhs_constraint_message())
```
MaykThewessen: **This warning collapses two distinct bugs into one message.** Line 1287 does `rhs.reindex_like(self.const, fill_value=np.nan)` *before* the `isnull()` check fires. NaNs end up in `rhs` for two completely different reasons:

1. User passed a NaN in the RHS array (the §5 case the warning text describes — "data error" or intended absence).
2. Coordinate mismatch: `rhs` and `self.const` have different labels, `reindex_like` introduces NaN at the unmatched positions.

`_legacy_nan_rhs_constraint_message()` says "NaN in the constraint RHS was silently kept" and suggests `fillna(value)` — wrong advice if the actual cause is a coord mismatch (where the fix is `.sel(...)` / `.reindex(...)`, not `.fillna`).

Suggest: check `dim_coords_differ(self.const, rhs)` *before* the reindex and emit `_legacy_coord_mismatch_message` in that case, keeping the `_legacy_nan_rhs_constraint_message` only for genuine user NaN.

linopy/expressions.py
```diff
@@ -1142,12 +1312,21 @@ def reset_const(self: GenericExpression) -> GenericExpression:
...
+        this is the universal signal. A term with ``vars == -1`` is a
+        *dead term* (contributes nothing), not a slot-level absence
+        signal; ``fillna(value)`` can revive an absent slot to a present
+        constant while leaving the dead sentinel term in place. The
+        legacy convention has no real absence concept (``const`` is
+        always filled with 0); the historical AND of "all vars
+        sentinel" and "const NaN" is preserved verbatim under legacy.
+        """
+        if is_v1():
+            return self.const.isnull()
```
MaykThewessen, line 1793: **§1/§2 storage invariant is not actively enforced at construction.**

`LinearExpression.__init__` (around line 351–407) coerces dtypes but does **not** call `absorb_absence` under v1. So a user-constructed expression (or one produced by a path bypassing `merge`) can carry a "half-absent" row: `const=NaN` with `vars≠-1`, or `coeffs=NaN` with `const` non-null. `isnull()` then reports `self.const.isnull()` while the row still has live terms.

Two options:
- Run `absorb_absence` inside `__init__` under v1 so the invariant is structurally guaranteed.
- Make `isnull()` itself check the full marker set (`(self.vars == -1).all(helper_dims) & self.coeffs.isnull().all(helper_dims) & self.const.isnull()`) so it's robust against constructor-bypass paths.

Either makes the v1 contract self-enforcing rather than depending on every producer remembering to call `absorb_absence`.

linopy/expressions.py
```diff
@@ -556,32 +575,112 @@ def _align_constant(
...
-        else:
+        try:
             self_const, aligned = xr.align(
                 self.const,
                 other,
                 join=join,
                 fill_value=fill_value,
             )
-            return self_const, aligned, True
+        except ValueError as e:
```
MaykThewessen: **Brittle string-match on xarray's exception text.** The wrap below uses `if "exact" in str(e):` to detect alignment failure from `xr.align(..., join="exact")`. xarray's error wording is not API-stable — a future release that rephrases the message will make this fall through to the bare `raise`, silently dropping the actionable `_shared_dim_mismatch_message` suggestion text.

Two safer options:
- Run the dim check *before* `xr.align` (use `merge_shared_user_coord_mismatch` from `semantics.py`) so the raise is structural, not exception-driven.
- Catch xarray's specific exception class (`xr.MergeError` / `xr.AlignmentError` in newer versions) instead of pattern-matching the message.

The `_align_constant` and `merge` paths would also stop producing two different error texts for the same logical condition (consistency win).

linopy/variables.py
```diff
@@ -327,12 +327,37 @@ def to_linexpr(
...
+            coefficient = coefficient.where(~absent)
+        else:
+            # LEGACY: warn if the variable carries absent slots — those
+            # silently contribute 0 here, but v1 will propagate the
+            # absence and produce a different result downstream. This is
+            # the origin of the most common legacy↔v1 divergence (masked
+            # variables in arithmetic) that no other warn-site catches.
+            has_absence = bool((self.labels == -1).any())
+            if has_absence:
+                warn_legacy(_legacy_masked_variable_message(self.name))
```
MaykThewessen, line 380: **Missing §5 check on the `coefficient` parameter.** `Variable.to_linexpr(coefficient)` accepts an array `coefficient` containing user NaN without any `check_user_nan_array(coefficient)` call. Under v1, NaN coefficients flow into the resulting expression silently (§6 will then propagate, masking the original data error); under legacy they are `fillna(0)`'d a few lines below, also silently.

§5 says user-supplied NaN in numeric constants must raise — `coefficient` is exactly that. Add `check_user_nan_array(op_kind="mul")` (since the coefficient enters multiplicatively) before the v1/legacy branch.

Also worth confirming `Constraint.lhs`-side multiplication routes through the same check; `x * series_with_nan` should reach the §5 raise on every operator path.

linopy/expressions.py
```diff
@@ -603,14 +702,56 @@ def _apply_constant_op(
...
+    def _apply_constant_op_legacy(
+        self: GenericExpression,
+        other: ConstantLike,
+        op: Callable[[DataArray, DataArray], DataArray],
+        fill_value: float,
+        join: JoinOptions | None,
+    ) -> GenericExpression:
+        # NaN values are silently filled with neutral elements before the op:
+        # factor → fill_value (0 for mul, 1 for div), coeffs/const → 0.
+        op_kind = "div" if fill_value == 1 else "mul"
```
MaykThewessen: **Fragile coupling: `op_kind = "div" if fill_value == 1 else "mul"`.**

`op_kind` is derived from the `fill_value` magic constant. If `_divide_by_constant` (or any future call site) ever changes its fill — e.g. to support safe-division semantics, or if the fill ever needs to be `inf` — the op-kind classification silently flips and users get the wrong per-op message.

Prefer threading `op_kind` explicitly as a parameter from each call site (`_multiply_by_constant` passes `"mul"`, `_divide_by_constant` passes `"div"`). Decouples message routing from numeric magic.

linopy/semantics.py
```diff
@@ -0,0 +1,310 @@
...
+    shared = set.intersection(*(set(p.keys()) for p in per_ds)) if per_ds else set()
+    for d_name in shared:
+        ref = per_ds[0][d_name]
+        for p in per_ds[1:]:
+            if not ref.equals(p[d_name]):
+                return str(d_name), ref.values, p[d_name].values
+    return None
+
+
+def conflicting_aux_coord(
```
MaykThewessen, line 566: **§11 asymmetric-presence policy not explicit.** The `if len(present) < 2: continue` short-circuit silently propagates the coord from the one operand that carries it. `convention.md` §11 only describes the conflict case (both operands have it, values disagree → raise). The asymmetric case is intentional behaviour but undocumented — worth a one-line addition to §11 ("when only one operand carries the coord, it propagates unchanged") so users don't get surprised when a one-sided coord survives a binary op.

Two minor secondary points:
- `equal_nan` is only set for floating dtypes (line ~282), so string/object aux coords carrying `None`/`NaN` placeholders will compare unequal even when both operands have identical None placement. Edge case but plausible for category labels.
- Shape mismatch and value disagreement both raise via the same `_aux_conflict_message`. They are different failure modes; the message renders both arrays but doesn't tell the user which one they hit.

linopy/variables.py
```diff
@@ -327,12 +327,37 @@ def to_linexpr(
         linopy.LinearExpression
             Linear expression with the variables and coefficients.
         """
+        from linopy.semantics import (
```
MaykThewessen: Function-local `from linopy.semantics import ...` inside `to_linexpr` — called on every variable→linexpr conversion (hot path: every `__add__`, `__mul__`, every coefficient apply). Module-level import would avoid the repeated lookup. Likely measurable in tight constraint-building loops; cheap to fix.

MaykThewessen (commented):
Full line-level pass on the implementation files (`semantics.py`, `expressions.py`, `variables.py`, `config.py`, `piecewise.py`, `sos_reformulation.py`) against the §1–§13 spec in `convention.md`. The spec ↔ code alignment is strong overall — almost every section has a direct implementation site, and the `is_v1()` switch is consistently placed. Findings below are not blockers to opt-in landing, but most should resolve before v1 promotes to default, because they constitute silent-divergence paths the rollout warnings cannot catch.

### High-impact (silent §-violations under v1)

1. **§8 / §11 enforcement gated behind `if join is None:`** in both `merge` (line 2573) and `_align_constant` (line 580). User-supplied `join=` bypasses aux-coord conflict check — the #295 bug v1 is supposed to close still reproduces with `join="override"`.
2. **§6 absence propagation gap on quadratics** (line 2633–2634): the FACTOR_DIM branch's `.prod(...)` uses xarray default `skipna=True`. The TERM_DIM branch right above correctly does `skipna = not is_v1()`. `expr * expr` with one absent factor silently collapses absence to identity 1.
3. **§5 not enforced on `Variable.to_linexpr(coefficient)`** (variables.py:312): array coefficients carrying user NaN flow through with no `check_user_nan_array` call.
4. **§1/§2 storage invariant not constructor-enforced** (expressions.py:1328 / `__init__`): no `absorb_absence` call inside `LinearExpression.__init__`. A bypass-constructed expression can carry "half-absent" rows; `isnull()` then misreports.

### Migration ergonomics

5. **`stacklevel` lands inside linopy internals on every call site I traced** (semantics.py:170 default `=3`, lines 200 / 207 default `=5`). IDE jump-to-source on the legacy warnings points at `__add__`/`merge`/`to_linexpr`, not at the user's call site — defeats the actionable-warning goal #2.
6. **Legacy RHS-NaN warning collapses coord-mismatch and data-error into one message** (expressions.py:1292). User gets `fillna(...)` advice when the underlying cause was a reindex artifact.
7. **`override` join semantic drift** (line 609): docs and `_shared_dim_mismatch_message` advertise "positional alignment, made explicit". Implementation is `assign_coords`, which is coord-renaming with no shape guarantee.

### Robustness

8. **Brittle xarray exception string-match** (line 624): `if "exact" in str(e):` will silently break on any upstream xarray rephrase. Structural check via `merge_shared_user_coord_mismatch` is safer.
9. **`op_kind` derived from `fill_value` magic value** (line 749) — fragile coupling between numeric fill and message routing.

### Spec/doc gaps

10. **§11 asymmetric-presence behaviour** (semantics.py:255) is implemented (propagate from the one operand that has it) but not stated in `convention.md` §11. One sentence closes it.
11. **§13 / `mean` of all-absent dim** still open — I raised this above; no impl decision visible yet.

### Performance (deferrable)

The `_align_constant` / `merge` hot paths run aux-coord checks on every v1 binary op without identity short-circuits. Likely measurable in `PyPSA-Eur`-scale models (thousands of `add_constraints` calls). Cheap to add `if a is b: return False` before the equals-check. Worth a profile pass once functionality is locked.

### What's working well

- The §12 raise plumbing for `to_constraint` is clean — `self.sub(rhs, join=join)` reuses the §8 path correctly.
- `piecewise.py` / `sos_reformulation.py` minimal `drop=True` / `assign_coords` edits are correct under both semantics; no findings.
- Per-op message split + masked-variable warn site (commit `9455683`) catches the most common PyPSA-Eur divergence at its origin. That's the highest-leverage migration affordance in the PR.

Happy to PR fixes for any subset (especially #1, #2, #3, #5, #8) once the design questions above are settled. Treat all line comments as suggestions, not blockers.

MaykThewessen:
You're right, @FBumann — `diagnose_alignment(model)` walking the built artifact is the wrong shape. Once a constraint is in `model.constraints`, the legacy-vs-v1 divergence has already collapsed: a silently-positionally-aligned `x + y` and an explicitly label-aligned one produce identical Datasets. The build path is the only place the difference is visible.

Pivot proposal: capture at op time, not audit time. The warnings from commit `9455683` already fire at every divergence site — turn them into a structured trace rather than `stderr` lines.

```python
with linopy.alignment_trace() as trace:
    m = build_my_model()

trace.report()   # pandas.DataFrame
#   file              line  op_kind  divergence              dim    n_unmatched
#   pypsa/optimize.py  412   add      coord_mismatch          time   17
#   pypsa/optimize.py  588   div      nan_user_fill           -      3
#   my_constraints.py  102   mul      masked_var_arithmetic   bus    2
```

Implementation sketch: context manager installs a `warnings.catch_warnings()` filter for `LinopySemanticsWarning`, parses each into a structured record (each `_legacy_*_message` already has a canonical shape — divergence class, op_kind, dim, fix-hint). On exit, returns a `Trace` object with `.to_df()` / `.summarise()` / `.first_n_per_site(n=5)`.

Two payoffs over raw warnings:
- Deduplicates: 50,000 identical warnings from a tight loop collapse to one row with `count`.
- Survives `filterwarnings('once')` / `'ignore'` — runs under its own catch context, independent of the user's global warning config.

Net: it's the warning machinery you already shipped, with structured capture and reporting on top. No new emission sites, no instrumentation in the hot path. PR-ready once the `legacy-warn` message format stabilises.

I'll open the PR once the spec-drift items from my review are settled — they affect what divergence classes the trace needs to recognise (e.g. if `_align_constant`'s `override` semantics get tightened, that's a new divergence class to surface).

FBumann:
@MaykThewessen Seems like a usefull addition!

linopy/expressions.py
FBumann: This would be somewhat of a hot path i think. And our user facing api does not expose LinearExpressions for users to create directly. So id skip this

FBumann:
@MaykThewessen
Feel free to PR a `linopy.alignment_trace()` context manager. Although if its not to complex i would prefer to have it as a documented helper in the docs instead of inside the codebase. But only if thats possible...

Also feel free to PR a context manager like 'with linopy.alignment(join="override")', although i would argue that you should wait until this PR is reviewed and merged.

I would defer performance polishes like the proposed short-circuits until we merge this, maybe even until we drop the legacy convention. THis will keep the convention more focused and readable as long as its not 100% proven.

linopy/expressions.py
MaykThewessen: Fair on the hot-path point — `merge` already calls `absorb_absence` on exit (line 2642), so every merge-produced expression is invariant-safe; `__init__` overhead on every binary op for a theoretical bypass isn't worth it.

Alternative shape of the same concern: make `isnull()` itself defensive — one extra `(self.vars == -1).all(helper_dims)` AND-mask in the v1 branch:

```python
def isnull(self) -> DataArray:
    if is_v1():
        # Defensive: check full marker set, not just const, in case a
        # future internal producer skips absorb_absence.
        return self.const.isnull() & (self.vars == -1).all(self.helper_dims)
    ...
```

Cheap (one ufunc per call), runs only on `isnull()` invocation (cold path — users call it once at debug time, not in tight loops), no per-construction cost. Catches future regressions if a new internal path ever forgets the `absorb_absence` at its tail. Up to you whether worth adding; happy to drop the line if the merge-only invariant is considered tight enough.

MaykThewessen:
Good call — docs recipe is tighter scope than a public `linopy.alignment_trace()` symbol. Less API surface to maintain, no docstring/signature compat to preserve across releases, users can copy-and-tweak for their own warning-filter needs.

Planned shape: a short notebook under `doc/` (next to existing examples) titled "Migrating from legacy to v1 semantics" with three sections:

1. **Set up a trace context.** ~20 lines: `warnings.catch_warnings()` + `simplefilter("always", LinopySemanticsWarning)` + a record list, wrapped in a `contextlib.contextmanager`.
2. **Parse warnings into a DataFrame.** Regex extracts `divergence_class` from the canonical message prefix (`"Coordinate mismatch"` / `"treated as 0"` / `"divisor.*treated as 1"` / `"'…'.*fillna"`), plus stacklevel-walk for `(file, line)`. ~30 lines.
3. **Report patterns.** `df.groupby(["file", "line", "divergence_class"]).size()` collapsing 50k repeats to a punch list.

End result reads as a recipe a user runs once during opt-in, gets a punch list, fixes top-N sites, re-runs until empty. I'll wait for the spec-drift items from the line review to settle so the regex catalogue matches the final message texts, then PR. Sound good?

linopy/expressions.py
MaykThewessen: Fair on hot-path + no public ctor, but want to flag a v1-internal risk before dropping:

The v1 branch enforces **§6 absence = `const.isnull()`** (pure). Legacy enforces the historical AND (`all vars sentinel AND const NaN`). If `.isnull()` collapses to legacy-AND under both modes, then any v1 internal path that relies on "const NaN ⇒ absent" mis-identifies slots where `fillna(value)` revived an absent slot while leaving the dead sentinel term in place — exactly the case the docstring calls out. A quick `rg '\.isnull\(' linopy/` to confirm no v1 propagation path calls it should settle this either way.

Two options if the audit comes back clean:

1. **Drop the branch**, mark `.isnull()` as legacy-semantics public API, add a private `_is_absent_slot()` helper for any future v1 internal use.
2. **Keep the branch, kill the overhead**: inline `options._semantics == "v1"` instead of the `is_v1()` call — the function-call cost was the hot-path concern, not the boolean check.

Slight preference for (2): preserves the §6 invariant as single source of truth and the inline check is ~2ns. Happy with (1) if the audit shows nothing internal touches it.

MaykThewessen:
Thanks @FBumann — that ordering works for me.

Plan:

Hold alignment(join="override") ctx mgr and the short-circuits until #717 merges (and revisit short-circuits post legacy removal).
For alignment_trace: try a pure-stdlib docs snippet first, leaning on the existing LinopySemanticsWarning channel — no new public API needed. Sketch below. If reviewers think it deserves to live in linopy/semantics.py after all (e.g. for stacklevel correctness across the call sites), happy to promote it then.
Proposed docs snippet (drop into the migration guide, not the package):

```python
import warnings
from collections import Counter
from contextlib import contextmanager

from linopy.semantics import LinopySemanticsWarning


@contextmanager
def alignment_trace(*, summary: bool = True):
    """Capture every LinopySemanticsWarning raised inside the block.

    Use during migration to surface where legacy alignment/NaN behaviour
    is being relied on, without touching the codebase under test.
    """
    with warnings.catch_warnings(record=True) as caught:
        warnings.simplefilter("always", LinopySemanticsWarning)
        yield caught

    if summary:
        hits = Counter(
            (w.filename, w.lineno, str(w.message).splitlines()[0])
            for w in caught
            if issubclass(w.category, LinopySemanticsWarning)
        )
        for (fname, lineno, msg), n in hits.most_common():
            print(f"{n:>4}× {fname}:{lineno}  {msg}")

```
Usage in migration docs:

```python
with alignment_trace() as warns:
    m = build_model()      # legacy semantics still default
    m.solve()
```
# → ranked list of call sites still relying on legacy NaN/align behaviour
Lets users sweep a real PyPSA-Eur / linopy workflow once, get a ranked punch-list of sites to fix, then flip linopy.options["semantics"] = "v1". Zero surface area in the library.

I'll open the docs PR after #717 lands so the import paths and warning class are stable.

FBumann:
Blocked by #732

FBumann:
For reference: [Needed changes in flixopt to pass under the new v1 convention](https://github.com/flixOpt/flixopt/pull/684). With those changes, the code **works under both v1 and legacy**

FabianHofmann:
@FBumann while working on #732 and trying to align the multiindex handling with the new convention, a gap in the convention revealed: The alignment on coords and dims does not only count for two linopy objects but for also for operations with non-linopy objects. IE. when performing a arithmetic operation between a linopy object and a non-linopy array-object, all expectations and conventions should be applied on the non-linopy object too. likely worth defining the object scope.

FBumann:
> @FBumann while working on #732 and trying to align the multiindex handling with the new convention, a gap in the convention revealed: The alignment on coords and dims does not only count for two linopy objects but for also for operations with non-linopy objects. IE. when performing a arithmetic operation between a linopy object and a non-linopy array-object, all expectations and conventions should be applied on the non-linopy object too. likely worth defining the object scope.

@FabianHofmann I think this should currently hold, as linopy owns the arithmetics. Did you find a case where this doesnt?

FabianHofmann:
> > @FBumann while working on #732 and trying to align the multiindex handling with the new convention, a gap in the convention revealed: The alignment on coords and dims does not only count for two linopy objects but for also for operations with non-linopy objects. IE. when performing a arithmetic operation between a linopy object and a non-linopy array-object, all expectations and conventions should be applied on the non-linopy object too. likely worth defining the object scope.
> 
> @FabianHofmann I think this should currently hold, as linopy owns the arithmetics. Did you find a case where this doesnt?

it's true, just explicitly naming that in the conventions would be beneficial. other objects have to first go via `as_dataarray` and then arithmetics are performed. that is `as_dataarray` must be fully aligned with the convention in order to get object-agnostic behavior, ie. operations between `expr` and ` arr` must behave exactly the same as operations between `expr` and `arr_expr` where `arr_expr` is a linear expression which only holds constants of the same shape as `arr`. this needs to hold for commutativity, associativity, distributivity.

FBumann:
> [!NOTE]
> Status update (Claude Code, prompted by @FBumann) — the branch is now current with master and the alignment layer is integrated with the semantics system.

Pushed in two commits:

**`04c7cb4` — Merge master** (brings in #732, #742, #743, #716)

The conflicts were small because this branch never touched `common.py` — the v1 logic lives in its own files. Resolution rule: keep this branch's v1/legacy dispatch structure, use master's `broadcast_to_coords(..., strict=False)` calls inside it. Test adaptations: the unnamed-MultiIndex coords entry in `test_legacy_violations` (required to be named since #732), and the masked-addend tails of `test_nterm` / `test_variable_names` split into `legacy`/`v1` pairs (§6 divergence).

**`f036c24` — Alignment ↔ semantics integration**

| Integration point | Resolution |
|---|---|
| #738 (warning channels) | the MI-projection deprecation now goes through `warn_legacy()` / `LinopySemanticsWarning`; `EvolvingAPIWarning` stays piecewise-only |
| Scenario B, v1 side | under `options["semantics"] = "v1"`, implicit MI-level projection **raises** (§8/§11) with guidance to project explicitly |
| Tests | MI deprecation tests marked `@pytest.mark.legacy`; each has a `@pytest.mark.v1` counterpart asserting the raise |
| Bonus fix | `as_expression` no longer swallows conversion errors ("Cannot convert to LinearExpression: *<original message>*") |

**Verification**: full suite under both semantics — 6446 passed, 532 skipped, mypy + pre-commit clean.

Remaining for this PR (as far as we can tell): the docs/migration-guide follow-up you listed as deferred, and your review of the merge resolutions in `expressions.py` / `variables.py` (the v1/legacy operator dispatch — worth a close look since both sides changed there).

FBumann:
_Claude Code:_

One scope question surfaced while writing the §11 stacked-MultiIndex spec: whether v1 should carry first-class MultiIndex support at all, or store MI input as flat dims + auxiliary level coords (which §11 already governs, with none of the projection machinery).

Filed as #744 with the complexity inventory, a verified flat+aux demo against this branch, and three options. It's your call — but the decision window is before this PR merges, so flagging it now rather than after.

FBumann:
> [!NOTE] 
> Generated with [Claude Code](https://claude.com/claude-code).

**v1 convention — what's left.** The arithmetic rules are all specified and implemented (#736 was the last). Remaining: one decision, the rollout, the 1.0 cleanup.

- [x] **Decide #744** (MultiIndex storage) — blocks the v1-default flip; underpins §11.
- [ ] **Pick when v1 becomes the default** (gated on #744 + the migration guide).
- [ ] **Flip default to v1**, legacy opt-out.
- [ ] **Audit warnings** — every v1 raise needs a matching `warn_legacy` (the no-silent-change guarantee).
- [ ] **1.0 legacy removal** — checklist in `legacy-removal.md`.
- [ ] **Write the migration guide** (`docs-plan.md` is only an outline).
- [ ] **Close #714** once v1 ships as default.

*§13 reductions: `sum` / `groupby.sum` are done & v1-pinned. `mean` / `resample` / `coarsen` named in §13 aren't in linopy at all yet — purely additive (→ #703), not a v1 gap; they'll follow §13 when added.*

**Test cleanup (follow-up):**

- [ ] Use the public API in assertions (`.indexes` / `.coords` / `.sizes` / `.coord_dims`) instead of internal `.data` / `.coeffs.coords`; keep `.coeffs`/`.const` only for genuine `_term`/`_factor` storage checks.
- [ ] Assert the **full** error/warning text in v1-raise and legacy-warn tests, to document the user-facing message.
- [x] Pin the **legacy** side of every v1/legacy divergence, not just raises/warns but different *results* — e.g. §13 `sum`/`groupby.sum` over absent: v1 skips (→20), legacy fills 0 (→25); only the v1 side is currently tested (`TestReductionsSkipAbsent` is all `@pytest.mark.v1`).
- [ ] De-duplicate repeated index/setup fixtures; parametrize v1/legacy where the result matches.

MaykThewessen:
> [!NOTE]
> Generated with [Claude Code](https://claude.com/claude-code).

**Quick triage from the PyPSA-Eur side.**

**CI red (`Check types`).** Four mypy `[return-value]` errors in `test/test_legacy_violations.py:2463-2489`: helpers annotated `-> LinearExpression` but the v1/legacy branches now return `LinearExpression | QuadraticExpression` (or the quad branch on its own). Widening the return annotation (or splitting into linear/quadratic helpers) clears it. Trivial unblock for the otherwise-green matrix.

**On the remaining checklist:**

1. **#744 first.** Agree it's the real gate. The "flat dims + auxiliary labels" framing is the lower-surprise option for downstream users: PyPSA-Eur builds nearly all MI-shaped variables by stacking after construction, so first-class MI in `linopy` would mostly serve a path we don't exercise. Happy to stress-test whichever direction you pick against the network-scale models on our end before the default flip.
2. **`warn_legacy` audit.** Worth scripting: walk every `raise` introduced in this branch, assert a sibling `warn_legacy` exists on the legacy code path. A `pytest` parametrized over the v1-raise sites (importing the legacy callable, asserting `DeprecationWarning`) would lock the no-silent-change guarantee into CI rather than reviewer attention.
3. **Migration guide.** Two things readers will actually need: (a) a table of "code that worked, what happens under v1, minimal rewrite" keyed by §-section, and (b) a `LINOPY_CONVENTION=legacy` escape hatch documented in one place. The §-catalogue from #714 is most of the table already.
4. **Default-flip timing.** Suggest cutting a `1.0.0rc` once #744 lands and the warn-audit is in CI, then sitting on rc for one PyPSA-Eur release cycle before the default flip. Surfaces legacy-warn noise in real downstream solves without forcing pins.

**Test cleanup.** The legacy-side pinning point matters most: divergent *results* (not just raises/warns) are the silent-failure class. Reductions over absent is the obvious one; worth grepping for other §13 / §8 sites where both code paths run clean but disagree.

Will rebase the PyPSA-Eur warmstart branch on top once #744 is resolved so we can give the convention real-model exposure before rc.

FBumann:
Performance regressions are expected. 

Im working on making them visible so we can adress as much as possible before merging!

FBumann:
Added a **Semantics report** workflow (`d1fa07a`) to keep the legacy→v1 memory cost visible during the migration.

- Builds every spec twice (default legacy + `LINOPY_BENCH_SEMANTICS=v1`) under the **same** node ids, then renders a legacy-vs-v1 A/B of build **peak + time** (`benchmem compare`/`plot`) as a PR artifact + step summary.
- **Report-only**, never gates. Legacy stays the unsuffixed default, so CodSpeed keeps its per-id history vs master (and still flags legacy regressions) — v1 lives only in this report until we choose to gate it.
- Transitional: obsolete once legacy is dropped. Pins `pytest-benchmem==0.4.5`.

<sub>🤖 Comment drafted by Claude (AI).</sub>

arithmetics-design/convention.md
```diff
@@ -0,0 +1,257 @@
...
+- `.sel()` / `.isel()` cut operands down to a shared subset — often the
+  clearest fix.
+- The named methods — `.add` `.sub` `.mul` `.div` `.le` `.ge` `.eq` — take a
+  `join=` argument: `exact`, `inner`, `outer`, `left`, `right`, or `override`.
+  `override` is the old positional behavior — still available, but now opt-in
+  and named rather than triggered by a size coincidence.
+- `.reindex()` / `.reindex_like()` conform an operand to a target index
+  (extending past the original creates absent positions — §4).
+- `.assign_coords()` relabels an operand outright (positional alignment, made
+  explicit).
```
MaykThewessen: §10 wording now lags the implementation. `_align_constant` gates `join="override"` on a shared-dim size check and raises *before* the `assign_coords` (expressions.py:708-733, with a comment citing §10), and the operand path routes override through `xr.concat(..., join="override")`, which already requires equal sizes. So the silent-broadcast hazard this paragraph reads as permitting is actually closed in code; only the spec text still describes `override` / `.assign_coords()` as unguarded positional relabels. Small clarification so the spec matches:

```suggestion
- The named methods — `.add` `.sub` `.mul` `.div` `.le` `.ge` `.eq` — take a
  `join=` argument: `exact`, `inner`, `outer`, `left`, `right`, or `override`.
  `override` is the old positional behavior — still available, but now opt-in
  and named rather than triggered by a size coincidence. It still requires the
  shared dimensions to match in size: a genuine size mismatch raises rather
  than relabelling mismatched data, so reach for a label join (`inner` /
  `outer` / `left` / `right`) when the sizes really differ.
- `.reindex()` / `.reindex_like()` conform an operand to a target index
  (extending past the original creates absent positions — §4).
- `.assign_coords()` relabels an operand outright. Unlike `join="override"`,
  which checks the shared dims match in size, this is an *unguarded* relabel:
  it renames positions whether or not they correspond, so the "made explicit"
  here is the caller's responsibility, not a safety check.
```

The line worth keeping crisp: `join="override"` is *guarded* positional alignment (size-checked, raises on mismatch); the bare `.assign_coords()` primitive is an *unguarded* relabel. Follows up the open `_align_constant` override thread.

MaykThewessen (commented):
One small spec-vs-code follow-up on §10 (the `override` / `assign_coords` wording), inline. Implementation already handles the size-mismatch case I'd flagged earlier; this just realigns the spec text. Not a blocker.

linopy/expressions.py
MaykThewessen: Circling back with the current branch (`51841fe`): the implementation side here is handled.

`_align_constant` now gates `join="override"` on a shared-dim size check and raises *before* the `assign_coords`, with a comment citing §10 (expressions.py:708-733):

> join='override' requires matching sizes on shared dimensions, but sizes differ on ...

and the operand-operand `merge` path routes override through `xr.concat(..., join="override")`, which already requires equal sizes. So the silent-broadcast / opaque-downstream hazard I flagged is closed on both paths. 👍 (The brittle `"exact" in str(e)` match I raised in a sibling comment is gone too, replaced by the structural `first_mismatched_dim` check.)

Only the spec text lags now: §10 still calls `override` "the old positional behavior" and `.assign_coords()` "positional alignment, made explicit" without the size-match precondition the code enforces, and it doesn't separate the *guarded* `join="override"` from the *unguarded* bare `.assign_coords()` relabel. I left a one-click suggestion on convention.md §10 with the two-bullet fix. Once that lands (or is declined), this thread can be resolved.

FBumann:
Discussion:
- reindexing should be eliminated from linopy to be more strict and point users to choosing behaviour deliberatels. @FBumann will do a stacked PR here to chow what changes (code, tests, examples)
- join parameter needs to have a value for the current default join (v1)

FBumann:
> [!NOTE]
>
> **Heads-up: v1 reorder handling is currently inconsistent across operators.** On a shared dimension carrying the *same labels in a different order*, `coeff * x` (the variable-coefficient path) already **raises** `ValueError` under v1 — it has its own `first_mismatched_dim` check on the raw coefficient — but `x + array`, `x - array`, `x <= array`, and the expr+expr `merge` path all **silently reindex** (align by label) via `_reindex_reordered_dims` / `conform_merge_dims`' v1 permute.
>
> So v1 today is "reorder-tolerant exact" for `+` / `-` / `<=` / merge, but "strict exact" for `*` — the same reorder, two behaviors.
>
> This is resolved by a stacked exact-alignment change: all paths raise on a pure reorder, with `join="outer"`/`inner`/… and `.sortby()` as the escape hatches, tightening §8 to full `join="exact"`.

FBumann:
@FabianHofmann #838 not only fixed the one remaining regression, it also improved legacy in some cases!

FBumann:
@FabianHofmann Unfortunately I had less time today than i hoped for.

Regarding behaviour, this is ready, but this still needs to be done before we merge into master:

- [x] Polishing the migration guide
- [x] Polishing the release notes
- [x] Polish the tests (inline comments, remove duplictaes, group into classes etc)
- [ ] Adjust the markdown specs for merge into master (placement in docs/droping them, ...?)
- [ ] Check with pypsa, flixopt and other tools

FBumann:
@FabianHofmann I fixed all issues you found.

We need to decide where to place the documentation of the convention.
I would suggest moving it to the docs.
Keep convention, goals, legacy removal.
remove the rest.

FabianHofmann:
> @FabianHofmann I fixed all issues you found.
> 
> We need to decide where to place the documentation of the convention. I would suggest moving it to the docs. Keep convention, goals, legacy removal. remove the rest.

you are a hero. this sounds good, let's do it. 

I added two tracker prs
- https://github.com/PyPSA/PyPSA/pull/1827 for tracking that this branch is fully backwards compatible in legacy mode
- https://github.com/PyPSA/PyPSA/pull/1829 for checking adjusted code works for v0.8 and this pr in both modes. 

I will gather feedback from those here. atm it looks like there is not big issues

FabianHofmann:
@FBumann quick update, making good progress on the pypsa compatibility side. will continue tomorrow, but if it continues like that, I would say we should be able to merge this by the end of the week

FBumann:
I also updated flixOpt here for reference including a reference of transferring patterns 

https://github.com/flixOpt/flixopt/pull/729

FabianHofmann:
@FBumann I am merging this now :rocket: well done and thank you for all the effort. follow-ups can go to master.

FBumann:
@FabianHofmann Wow, what a reign of issues closing!

PR #718, opened 2026-05-21
feat(persistent): in-place solver updates (Solver framework + HiGHS/Gurobi/Xpress/Mosek)

FabianHofmann:

Adds an opt-in persistent-update framework so a built solver can be re-solved against a mutated `Model` without a full rebuild.

**Core**
- `linopy.persistent`: `ModelSnapshot`, `ModelDiff`, `StructuralKey`, `VarKind`, `RebuildReason`. `ModelDiff` stores changes in flat-native arrays (bounds / var-types / RHS / signs / COO coefs / linear objective / sense) plus per-container `VarSlice` / `ConSlice` views.
- `ModelDiff.from_snapshot(snap, model)` and `ModelDiff.from_models(m1, m2)` — snapshot-based and snapshot-free diffs.
- `_coef_dirty` flag on constraints with RHS-setter short-circuit so RHS-only edits skip the coefficient re-walk.

**Solver orchestration**
- `Solver` gains `track_updates`, lazy-build (first `solve(model, …)` builds), `apply_update`, `update`, `disallow_rebuild`, and structured rebuild reasons. Backends without persistent-update support short-circuit to rebuild.
- Per-backend `apply_update`:
  - **HiGHS** — `changeColsBounds` / `changeColsIntegrality` / `changeRowBounds` / `changeCoeff` / `changeColsCost` / `changeObjectiveSense`. Sign change → rebuild.
  - **Gurobi** — `setAttr(LB/UB/VType/RHS/Sense/Obj)`, `chgCoeff`, `ModelSense`. In-place sign change.
  - **Xpress** — `chgbounds` / `chgcoltype` / `chgrhs` / `chgrowtype` / `chgmcoef` / `chgobj` / `chgobjsense`. In-place sign change.
  - **Mosek** — `chgvarbound` / `chgconbound` / `putvartypelist` / `putaijlist` / `putclist` / `putobjsense`. Sign change → rebuild.

**Tests**
- New `test_persistent_snapshot_diff.py` covering all `ModelDiff` semantics.
- New parametrized `test_persistent_apply_update.py` running 9 cases × 4 backends (skipped per backend when license/installation is unavailable).
- Cross-instance, pickle, and threading coverage in `test_persistent_solver_extras.py`.

## Follow-ups (tracked)

Findings from a symmetric rolling-horizon benchmark on a PyPSA-Eur network (426k vars / 1.28M cons, 168h window, 24h step, HiGHS):

- [ ] ~**Basis policy after `apply_update`**: the in-place path reuses the stale basis unconditionally. With HiGHS a valid basis skips presolve, so a 24h-shifted window hot-starts dual simplex on the full unpresolved LP — 126–219s vs 9.4s after `clearSolver()` (13–23x regression). Expose a `reset_solver`-style option on `Solver.solve` and/or auto-clear the basis when the diff touches a large fraction of rows.~ To be done by the user, add to docs 
- [ ] **Sanitization on the `Solver.solve(model=...)` path**: `Model.solve` runs `sanitize_zeros` / `sanitize_infinities` before building (drops ~11% of rows on the benchmark network); the bare `Solver.solve` path never sanitizes, so persistent users submit a larger problem. Can't be bolted on unconditionally — sanitize-induced mask changes between solves can themselves trigger rebuilds (`STRUCTURAL_LABELS`/`SPARSITY`).

## Checklist


FabianHofmann:
@FBumann here we go, if you want to take a first look. docs to come

FBumann:
> @FBumann here we go, if you want to take a first look. docs to come

Ill have a look ltoday

FBumann:
@FabianHofmann Sorry, I wont be able to review this today.

FabianHofmann:
> @FabianHofmann Sorry, I wont be able to review this today.

take your time, there is no hurry. I'll do some integration tests anyway

FBumann:
Are there any conflicts with #717? Are we sure we want to merge and publish this before we go v1?

Just making sure...

MaykThewessen:
Benched PR #718 ModelDiff vs hand-rolled `changeRowsBounds` path on a synthetic 69K-var / 160K-row DC-OPF, 8 weekly-style chunks (RHS + objective mutations only, identical structure across chunks). Per-chunk averages, chunk 0 excluded:

| route | t_build | t_diff | t_apply | t_solve | t_total | in_place | rebuilds |
|---|---|---|---|---|---|---|---|
| blueprint + `changeRowsBounds` (hand-rolled) | 0.000 | 0.001 | 0.006 | 0.759 | **0.766** | 0 | 0 |
| ModelDiff sweep (fresh `Model` per chunk) | 0.222 | 0.012 | 0.039 | 12.365 | **12.652** | 7 | 0 |
| ModelDiff in-place (mutate `model.constraints[name].rhs.values[...]`) | 0.003 | 0.010 | 0.041 | 4.919 | **4.982** | 7 | 0 |
| ModelDiff in-place **+ manual `solver.solver_model.clearSolver()`** | 0.003 | 0.010 | 0.041 | 0.572 | **0.637** | 7 | 0 |

Diff compute + `apply_update` sum to ~50 ms — those are not the bottleneck. The regression is entirely in `t_solve`: after in-place RHS mutation the persistent path keeps HiGHS's prior basis, and HiGHS then skips presolve (`"Solving LP with useful basis so presolve not used"`). For LPs with strong presolve reduction (89 % row reduction on our production model, similar on this synthetic) the skipped presolve costs more than the warm basis saves.

Manually calling `solver.solver_model.clearSolver()` between `solver.update()` and `solver.solve()` recovers presolve and brings the persistent path slightly under the hand-rolled baseline. But that defeats the encapsulation — the user reaches past the `Solver` API to touch the native HiGHS handle.

Suggestion: expose an opt-in on the persistent API so users with presolve-heavy LPs can drop the warm basis without bypassing the encapsulation. Either:

- a `keep_basis: bool = True` kwarg on `Solver.solve()` / `Solver.update()`, or
- a `Solver.clear_basis()` method on the base class with backend-specific implementations (`Highs.clearSolver()`, Gurobi `reset()`, etc.).

Zero structural rebuilds observed across all 7 re-solves in both ModelDiff routes (`RebuildReason.NONE`). The diff correctly classifies these as RHS + coefficient-only changes — the persistent API is doing its job, the missing piece is just the basis-clear hook.

Caveats:
- Synthetic DC-OPF surface (no SoC handoff, no NTC, no PST). Production LP is ~9× larger; presolve reduction and basis-warmstart behavior may differ at scale. Plan to repeat at production scale once the basis-clear hook lands (or we wire `clearSolver()` through manually).
- Linopy pinned to PR #718 head `089cf2e09`. Bench harness is internal; happy to share details on request.

MaykThewessen:
Production-scale follow-up: PSA-MEE `Validation_w2025` LP measures **1.14 M cols × 1.97 M rows** per weekly chunk (DC-OPF + ramp-penalty aux vars). The persistent-update phase — `changeRowsBounds` + `changeColsCost` on the time-varying rows (~15 % of the matrix), the equivalent of `solver.update(model, apply=True)` in the PR #718 API — completes in **0.16 s** at this scale. Linear scaling from the synthetic bench's ~50 ms, so ModelDiff's diff + apply overhead stays negligible at production size.

Solve-phase comparison (B vs SQ vs B+clear) at production scale couldn't be captured this round — the reduced-pipeline harness wasn't LP-feasible past chunk 1, so HiGHS timings came from an infeasibility path and would mislead. Will retry once the harness uses the full PSA-MEE chunk loop with storage + N-1 enabled. The presolve-skip mechanism that drove the synthetic regression (HiGHS logging `"Solving LP with useful basis so presolve not used"` after warm in-place RHS updates) is shape-independent, so I expect the gap to widen with the production LP (89 % presolve reduction on this model — more lost when skipped) — but not posting that as a result until it's measured.

Bottom line for the PR scope: the `solver.update(...)` / `solver.solve(model)` mechanism is the right primitive and the linear scaling holds. The remaining ask is still the opt-in basis-clear hook — without it, every backend that gates presolve on basis presence (HiGHS does, others may) leaves the persistent path slower than `solve_problem_from_model` on warm-state-hostile LPs.

FBumann:
@MaykThewessen @FabianHofmann 
I would argue that we should (for now) not provide methods like `clear_basis()`(or `clear_mip_state()`, `clear_pdlp_iterate()`, `clear_scaling()` etc.).
Those are properly exposed by the Native solver instances and support and naming differs between solvers. 
Users should instead do `Model.solver.solver_model.*`. This also saves us from differing in default behaviour from the respective solver.

We might be able to add such later. But I'd definitely defer it from this first PR. We should put a note into the docs about users maybe profiting from calling such methods on the solver directly.

**One downside:** We propose suage of `Model.solver.solver_model`, which should therefore stay stable.

FBumann:
I read through the code and overall api. I would propose the following api guide lines:

### 1. Mutation API: `Variable.update()` / `Constraint.update()` — and only that

```python
m.variables["x"].update(lower=new_lb, upper=new_ub)
m.constraints["demand"].update(rhs=new_rhs, sign="<=")
```

- One canonical mutation path. Routes through  validation&alignment; flips `_coef_dirty` in one place.
- Drop the existing `.lower` / `.upper` / `.rhs` / `.sign` / `.lhs` / `.coeffs` / `.vars` setters. The read-side properties stay; mutation via `=` becomes `AttributeError` with a pointer to `.update()`. Easy to add back if users push back — much harder to remove later once two APIs coexist.
- The `Constraint.rhs.setter` magic (rhs=Variable/Expression silently rewriting lhs) becomes an explicit method.
- `.values[...]` stays *possible* (xarray exposes it) but **not supported**: snapshot diff stays array-based as a safety net, (needed to compare different models anyway), but no contract that direct mutation propagates.

### 2. No `clear_basis()` — use `solver.solver_model` directly

```python
solver.apply(model_new)
solver.solver_model.clearSolver()    # backend-specific
solver.solve(assign=True)
```

- Each backend has one obvious method; no shared logic to abstract.
- Adding `clear_basis()` opens a cascade (`clear_mip_state`, `clear_pdlp_iterate`, `clear_scaling`, …) where every method needs per-backend semantics that don't unify cleanly (Mosek IPM has no "basis" in HiGHS's sense).
- Linopy's value is build + diff + apply. Native state management is the user's.
- Document the escape hatch in the persistent-solver guide with HiGHS + Gurobi recipes.

### 3. Split `Solver.update(model, apply=...)` into `diff()` / `apply()`

| call                       | reads model | mutates handle | runs solver |
| -------------------------- | :---------: | :------------: | :---------: |
| `solver.diff(model)`       |      ✓      |                |             |
| `solver.apply(model)`      |      ✓      |        ✓       |             |
| `solver.solve(model=...)`  |      ✓      |        ✓       |       ✓     |

- `update(model, apply=True|False)` does two structurally different things behind one bool; default `apply=True` silently mutates.
- Name collides with `gurobipy.Model.update()` (lazy-flush attribute changes) — different semantic.
- Split makes the basis-clear flow legible (`apply` separately from `solve`).


_Good news: No conflicts with the new v1 semantics convention_

FBumann:
Besides the **v1 semantics**, im also working on making `add_variables()` and the underlying `as_dataarray` more predictable and stictly follow the passed coords. THis should simplify some validation here i think.
See #722 #725 #726

MaykThewessen:
@FBumann Agreed: drop `clear_basis()` (and siblings) from this PR.

Native handle via `Model.solver.solver_model.*` already covers the power-user case, and any wrapper would have to flatten genuinely divergent semantics:

| solver  | basis/state reset                              |
|---------|------------------------------------------------|
| HiGHS   | `h.clearSolver()` (drops basis + dual ray)     |
| Gurobi  | `m.reset(0)` solution only, `m.reset(1)` incl. basis |
| Xpress  | `p.postsolve()` / `p.clearmipfilter()` (MIP)   |
| Mosek   | `task.putintparam(iparam.bi_clean_optimizer, ...)` |
| CPLEX   | `c.cleanup(epsilon)`                            |
| PDLP    | iterate is not user-exposed at all              |

A uniform `clear_basis()` would either lie about what it cleared (HiGHS clears more than basis; Gurobi `reset(1)` clears less than HiGHS `clearSolver`) or force per-solver branching back into linopy. Both worse than `solver_model.*`.

My benchmark comment's "B+clear" route was exploratory, not a request for the wrapper. Happy to leave basis-management outside this PR and revisit only if a clear cross-solver use case emerges.

Suggested docs note for `doc/release_notes.rst` (or wherever the persistent-solver section lands):

> **Resetting solver state.** `Solver.update()` preserves the warm-started basis / MIP state by default. Solvers expose basis-, MIP-, scaling-, and iterate-reset under different names; linopy does not unify them. If a re-solve benefits from discarding internal state, call the native solver directly via `Model.solver.solver_model`. Examples: HiGHS `m.solver.solver_model.clearSolver()`; Gurobi `m.solver.solver_model.reset(1)`; Xpress `m.solver.solver_model.postsolve()`. See each solver's API docs for the full menu.

LGTM to defer the unified API: additive later, non-breaking.

FBumann:
@FabianHofmann first round of my Review:

### Architecture

**A1. `_coef_dirty` is a correctness footgun glued to a perf shortcut.**
`linopy/persistent/diff.py:401-402` enables `skip_coef_compare = same_model and not coef_dirty`, and `from_snapshot(same_model=True)` is the **default** (`diff.py:346`). So the flag is *trusted*: if it says nothing changed, the CSR walk is skipped.

Through any supported mutation path — the existing setters at `constraints.py:1122,1136,1180`, and the new `Constraint.update(...)` in #727 (which sets the flag from one place; setters become shims) — this is precise: True iff coefs were touched. Through `c.coeffs.values[...] = ...` (or any in-place numpy mutation) the flag is stale, and `same_model=True` means the diff silently misses it. Backend keeps stale coefs. Silent wrong result.

Two coherent positions:
- **A.** Keep `same_model=True` default. `.values[...]` is documented as unsupported; users who hold up their end get correctness *and* the optimization. Risk: silent wrong diff on any unsupported mutation.
- **B.** Default `same_model=False`. The flag becomes an opt-in perf hint. Cost: extra CSR walk per same-model warm update. Benefit: footgun closed regardless of user discipline.

Recommendation: **B** is the safer default. Cross-model paths already use False; in the same-model path the perf cost is small. Solver-aware callers who promise to use `.update()` exclusively can opt back into the optimization with `same_model=True`.

**A2. Read-only `ModelSnapshot.capture` mutates the model.**
`snapshot.py:159-161` walks every `Constraint` and resets `_coef_dirty=False`. `capture` reads as a pure observation, but it isn't — anyone calling it for inspection (debug, tests) silently changes the model. The dirty-clear belongs in the solver-side bookkeeping (`Solver._update_locked` after a successful apply), not in the snapshot constructor.

**A3. `from_snapshot` vs `from_models` is two parallel walkers.**
`diff.py:341-525` — the two classmethods share ~80% of their body (structural checks, per-container var/con/objective walks, builder finalization). They differ only in whether the baseline buffers come from a stored `ModelSnapshot` or are freshly `_extract_*`'d from `model_a`. Factor out a single `_diff(get_var_base, get_con_base, get_obj_base, same_model)` core; the two public entry points become 10-line adapters.

**A4. Per-backend `apply_update` reinvents the dispatch frame 4×.**
`solvers.py:1404, 1891, 2455, 3183` — each backend has the same seven blocks in the same order (`var_bounds`, `var_type` (+binary bound re-clamp), `con_rhs`, `con_sign`, `con_coef`, `obj_c`, `obj_sense`), each guarded by `if diff.<...>.size:`. The only thing that varies is the API call. A template method in `Solver` iterating sections plus per-backend hooks (`_apply_var_bounds`, `_apply_coefs`, …) collapses ~600 lines and removes drift risk.

**A5. Lock is too coarse.**
`solvers.py:704` holds `self._lock` for the entire `solve(...)` body, including the actual native solve. This serializes solves on the same `Solver` instance for the full solver runtime. Protecting `solver_model`/`snapshot` mutation is correct; holding through `_run_direct` for minutes is not. Narrow to the build/diff/apply window, then drop the lock around the run.

**A6. Snapshot is the wrong unit of caching.**
`_update_locked` re-captures the entire snapshot after every successful in-place update (`solvers.py:811`). If you're tracking updates so each iteration is fast, paying an O(nnz_total) re-capture per iteration defeats the purpose. Either advance the snapshot incrementally from the applied diff (cheap — the diff already contains the new values + indices), or recompute lazily only when a diff is actually requested.

---

### DRY

**D1. `from_snapshot` ↔ `from_models` (see A3).** The biggest DRY hit in the new code.

**D2. `_HIGHS_VTYPE_MAP` lazy-init pattern (`solvers.py:1402, 1469-1477`).** `cls._HIGHS_VTYPE_MAP or self._init_highs_vtype_map()` runs on every `apply_update`. A `@functools.cache` classmethod returning the map is one method, no class-state mutation, no per-call truthiness check.

**D3. Xpress `.astype(np.int64, copy=False).tolist()` repeated 7×** in `apply_update` (lines 2496-2517). A `_to_int_list(arr)` helper or pre-converting once would clean it up.

**D4. "Set binary integrality and then re-clamp bounds to [0,1]"** is hand-rolled in HiGHS (1433-1442), Xpress (2485-2492), MOSEK (3221-3225). Lift to a Solver utility — the rule is solver-independent (it's linopy semantics, not a backend quirk).

**D5. `_extract_var_buffers` / `_extract_con_buffers` are imported privately from snapshot.py by diff.py** (`diff.py:12-21`). When two modules share private helpers, the helpers belong in a shared internal module, not the one that happened to introduce them.

FBumann:
## Test coverage gaps — `RebuildReason` audit

| Reason | Covered? |
|---|---|
| `NONE` | ✅ |
| `STRUCTURAL_CONTAINERS` | ⚠️ only in `in {LABELS, CONTAINERS}` sets |
| `STRUCTURAL_LABELS` | ⚠️ same; per-container branch (`diff.py:557,606`) unreachable from public API |
| `COORD_REINDEX` | ✅ |
| `SPARSITY` | ✅ |
| `QUAD_OBJ` (`diff.py:702`) | ❌ no tests |
| `BACKEND_REJECTED` | ✅ |

**Gaps worth fixing:**

1. **`QUAD_OBJ` has zero tests** — both directions (linear→quad, quad→linear) are reachable. Without coverage, QP users risk silent miscompare.
2. **Loose `in {…}` assertions** in `test_sparsity_change_triggers_rebuild`, `test_cross_model_sparsity_change_rebuilds`, `test_cross_model_structural_mismatch_rebuilds` — tighten to `is` so future refactors of check ordering get caught.
3. **`disallow_rebuild=True` only tested for one reason** (STRUCTURAL_CONTAINERS). The contract is "every rebuild trigger raises" — add tests for SPARSITY, COORD_REINDEX, QUAD_OBJ, BACKEND_REJECTED.
4. **`from_models` negative paths uncovered** — only the happy path is tested. Mirror the `from_snapshot` cases (structural/sparsity/quad/coord) since the buffer-extraction path is independent.
5. **Dead code or missing test** at `diff.py:557` and `:606` — the per-container `STRUCTURAL_LABELS` returns can't fire after the global label check at `:372-377` short-circuits. Either delete (replace with `assert`) or write the test that justifies them.

MaykThewessen:
@FBumann Strong review. Adding bench-grounded color on A1 / A5 / A6 since my numbers are the ones being cited.

### A1: `_coef_dirty` default

Agree with **Option B** (default `same_model=False`, flag as opt-in perf hint). One framing nit: the "~50-200 ms at Mayk-scale, small relative to solve time" line under-sells the cost. At the 1.14 M × 1.97 M production size, the **apply phase** is 0.16 s; a 50-200 ms CSR walk would be 30-125% of *apply*, not 30-125% of solve. Still small absolute, and still the right call. But if the trade-off is sold as "free", reviewers will be surprised when the apply-phase profile shows the walk dominating once a real solver loop runs hundreds of iterations.

Concrete: I'd take B plus a `same_model=True` opt-in documented as "caller asserts only `.update()` / typed setters were used; `.values[...]` invalidates this guarantee." Solver-aware orchestrators (the use case driving #718) will set it; ad-hoc users get correctness.

### A2: `ModelSnapshot.capture` mutating `_coef_dirty`

Agree. Moving the clear into `Solver._update_locked` after a successful apply is correct on two grounds: (1) snapshot becomes a pure observation; (2) the clear becomes coupled to "the solver actually saw these coefs", which is what the flag is supposed to mean.

### A3 / A4 / D-series

No bench-side input. Agree the DRY hits are worth taking. A4 in particular: four parallel 7-block dispatch bodies will drift the moment one backend grows an 8th section.

### A5: lock around full solve

Confirmed real at production scale. My weekly-chunk run holds the solve phase for minutes; with the current `solvers.py:704` lock width, any concurrent `solve()` on the same `Solver` blocks for the full run. Persistent-update orchestrators (the use case here) are exactly the ones likely to want concurrent solves on sibling models or pipelined iterations. Narrowing the lock to build/diff/apply is the right scope.

### A6: re-capture after every successful update

This one bites at production size. `_update_locked` re-capturing the full snapshot at `solvers.py:811` is O(nnz_total) per iteration. At 1.14 M × 1.97 M with ~6× nnz/row, that's tens of millions of cells re-read per chunk just to re-establish a baseline the diff already knew how to advance. The diff carries `(indices, new_values)` for every mutated buffer; applying that delta to the held snapshot is O(|diff|), which is what the persistent path is *for*. Strong +1 to "advance the snapshot incrementally from the applied diff." Lazy recompute on next `diff()` is a fine fallback if incremental advance is awkward for some backend.

### Test gaps

All five gaps look right. `QUAD_OBJ` is the one I'd treat as blocker: QP users are exactly the audience for #718 (warm-starting a barrier solve from a previous iterate), and silent miscompare on a quad to linear flip would be hard to diagnose downstream.

The `disallow_rebuild=True` coverage gap is structural: the contract is "every rebuild trigger raises", but the test only proves one of six does. A parametrized test over `RebuildReason` enum members closes it in five lines.

Happy to take A6 (incremental snapshot advance) as a follow-up PR after #718 lands if Fabian wants to keep this one focused; the API surface doesn't change, only the body of `_update_locked`.

MaykThewessen:
Correction to my A1 framing: I overshot.

Microbench of the actual compare path (`np.array_equal` on labels/indptr/indices + `new_buf.data != base_buf.data` + the `row_value_changed` build at `diff.py:624-629`), numpy 2.4.6, M1, median of 7:

| scale            | n_rows    | nnz        | extra cost (`same_model=False` vs `True`) |
|------------------|-----------|------------|-------------------------------------------|
| synth-69k        | 69 k      | 0.2 M      | **0.5 ms**                                |
| mid-300k         | 300 k     | 1.5 M      | **3.2 ms**                                |
| prod-1.97M @ 6   | 1.97 M    | 11.8 M     | **34.9 ms**                               |
| prod-1.97M @ 10  | 1.97 M    | 19.7 M     | **54.6 ms**                               |

So at the realistic production size (6 nnz/row), the extra walk is **~35 ms**, not the 50-200 ms my last comment implied as the floor. That's ~22 % of a 160 ms apply phase, or ~0.3 % of a 12 s solve. Felix's original "small relative to solve time" framing was right; my "30-125 % of apply" was the worst-case interpretation, not the typical one.

Doesn't change the recommendation (still Option B: default `same_model=False`, flag as opt-in perf hint). Just wanted the numbers on the record so reviewers don't price the trade-off at 3-5× what it actually costs.

Bench script available on request, or I can drop it in `benchmark/` if useful.

coroa:
Have not reviewed the code yet, just read the comments during some spare time, but my position is always to try to have perf by default:

> A. Keep same_model=True default. .values[...] is documented as unsupported; users who hold up their end get correctness and the optimization. Risk: silent wrong diff on any unsupported mutation.
> B. Default same_model=False. The flag becomes an opt-in perf hint. Cost: extra CSR walk per same-model warm update. Benefit: footgun closed regardless of user discipline.

So, i'd argue for the footgun A, but put the security mechanism on by flagging the numpy arrays as read only:
```python
a = np.array([1, 2, 3])
a.flags.writeable = False
a[0] = 99  # ValueError: assignment destination is read-only
```

FBumann:
@coroa I wrote some comments about the user api here: #727 

In short:

We could **skip the user parameter entirely** by storing a `weakref` to the source model on `ModelSnapshot`.

Together with your suggestion about the writable flag this would be very safe, convenient and uses the performant branch whenever possible.

FBumann:
@coroa I asked *C*laude code** about potential gaps of the writeable flag: 

1. **Dask** — `.flags.writeable` is numpy-only; dask-backed buffers ignore it. Inconsistent enforcement on mixed-backend models.
2. **`.view()` escape** — `v = arr.view(); v.flags.writeable = True; v[...] = 99` re-arms the protection. Stops casual mutation, not deliberate.
3. **Lifecycle** — needs an explicit thaw on every rebuild / close / pickle round-trip path. Easy to miss one.
4. **Internal paths** — every linopy internal that touches `.values[...]` during a frozen window breaks (`sanitize_zeros`, SOS reformulation, …). Needs an audit.
5. **Error UX** — `ValueError: assignment destination is read-only` doesn't explain why or how to fix.
6. **Doesn't cover cross-model** — orthogonal to the `from_snapshot(snap_from_m1, m2)` footgun.

Feels 80% safe, but might produce rough edges in our internals and many places where we have to reassign this flag.

FabianHofmann:
@FBumann great review! I am still wondering what the benefit of an explicit `.update` function would be over the already supported setters. the latter already exist and from an API perspective supporting both would be a functional duplication. what would `.update` do differently than funneling the coef dirty logic?

MaykThewessen:
The three proposals (coroa's writable-flag, @FBumann's weakref auto-decide, and the `.update()` API in #727) compose into one design rather than competing. Sketching:

**Unified shape:**

```python
ModelSnapshot.capture(model):
    snap.model_ref = weakref.ref(model)
    snap.copy buffers (the existing snapshot.py:67-78 path)
    freeze live model arrays: var.lower.values.flags.writeable = False, etc.

ModelDiff.from_snapshot(snap, model=None):
    # weakref derives same_model automatically; no user-facing param
    same_model = (model is None) or (snap.model_ref() is model)

Constraint.update(rhs=..., sign=..., coef=...):
    with snap.thawed():        # context manager, re-entrant
        ... apply changes ...
        flip _coef_dirty
    # __exit__ re-freezes

Solver._update_locked:
    with snap.thawed():
        apply_update(diff)
        re-capture or advance incrementally
```

Each proposal lands on a different layer:

| layer            | proposal                  | closes                                        |
|------------------|---------------------------|-----------------------------------------------|
| pairing          | weakref auto-`same_model` | cross-model footgun (`snap_from_m1, m2`)      |
| mutation channel | writable-flag freeze      | A1 (live `.values[...]` mutation)             |
| mutation API     | `.update()`               | the only path that thaws → mutates → freezes  |
| flag flip        | `.update()` chokepoint    | `_coef_dirty` correctness                     |

That answers @FabianHofmann's "what does `.update()` do that setters don't?" — once writable-flag freezes the live arrays after capture, `.values[...]` is no longer a parallel path; it raises. `.update()` is the only thing that thaws-mutates-re-freezes, which is what makes it not a functional duplicate of setters. Setters either get deprecated or become thin aliases for `.update()`.

### On the gap-list

@FBumann the six gaps you listed are real, but the weights I'd put on them:

- **Point 4 (internal-path audit) is the dominant cost**, not point 1 (dask) or point 2 (view-escape). A quick grep for `.values[` + `=` in `linopy/` core: ~30 candidate sites including `sanitize_zeros`, `sos_reformulation`, `piecewise.py` LP padding, every backend `apply_update`, `Constraint._set_rhs`. Each needs a `with snap.thawed():` wrap or a direct buffer-assignment alternative. That's the actual scope of the change, not a footnote.
- **Point 2 (view-escape)** is mostly cosmetic — requires deliberate `arr.view().flags.writeable = True`, same level as `__slots__` not stopping `object.__setattr__`. Stops accidental, not malicious. Acceptable.
- **Point 1 (dask)** is genuinely orthogonal — dask buffers don't expose mutable contents anyway, so writable-flag is a no-op there; the model just becomes implicitly safe in that backend without enforcement.
- **Point 6 (cross-model)** is what the weakref closes, so it's already handled by the composed design.

Two gaps not in the list:

- **Re-entrancy.** `apply_update` may trigger a path that itself wants to thaw (e.g. coef change → SOS reformulation → buffer rewrite). The thaw context has to be a refcount-stacked counter, not a boolean. Easy to get wrong on first cut.
- **Error UX with location.** `ValueError: assignment destination is read-only` is the bare numpy error. Most lightweight fix: catch `ValueError` at the public-API boundary and re-raise as `LinopyFrozenModelError("Mutate via .update(...) between snapshot capture and apply.")`. Subclassing ndarray is not recommended by numpy.

### Recommended landing order

1. **Weakref auto-`same_model` first** — small, mechanical, removes the user-param confusion immediately. Doesn't depend on freeze.
2. **`.update()` API as the single supported mutation path** (FBumann's #727). Setters keep working.
3. **Writable-flag freeze, behind an opt-in `Solver(track_updates=True, strict_freeze=True)` flag.** Audit internal paths before flipping the default.
4. **Flip default once the internal-path audit is clean.**

Step 3 behind a flag lets you ship 1 + 2 immediately and gather real-world data on which internal paths break under freeze before forcing everyone through them. Mayk-scale benches stay green throughout (the freeze is free at runtime; it's the audit that costs).

I can do the internal-path audit + thaw-context wrap as a follow-up PR if useful — mechanical work that benefits from running real PSA-MEE models against each candidate site.

FBumann:
> @FBumann great review! I am still wondering what the benefit of an explicit `.update` function would be over the already supported setters. the latter already exist and from an API perspective supporting both would be a functional duplication. what would `.update` do differently than funneling the coef dirty logic?

@FabianHofmann my main reasons are:
1. `update()` is  a much cleaner api contract:
   - It's a single source of truth of what is possible to update (and what isnt). 
   - It's discoverable: _Need to modify an existing model? use `*.update()` - for both `Variable`s and `Constraint`s._ Everything else is documented inside the respective `update()`, not scattered over different setters, cross referencing.
   - It exposes the linopy internals much less. Users don't need to know which fields `Variable` and especially `Constraint` have.
   - Setting `con.rhs = 2 * x` needs to leak into `lhs`, which is very surprising for a setter, but not for a method.
2. `update()` is more flexible than scattered setters
    - the api can focus on usability: For example, `Constraint.update()` can mirror the signature of `m.add_constraints()`, allowing you to pass an entirely new constraint instead of modifying lhs/rhs/sign individually . _This usecase would feel really weird with only setters. Which to modify first? cross-mutation? ..._.
    - We can also add  additional parameters that describe how to update if needed, which will not be possible with setters.
4. `update()` has much better ergonomics:
    - `Constraint.update()` can mirror the signature of `m.add_constraints()`, allowing to pass full new constraints. Its the same as creating a constraint - **Users didnt construct a Constraint with sign, coeffs, etc**, but with `m.add_variables()`.
    - passing `rhs=4, sign="<="` is much better than 2 separate setters. It skips the weird intermediate state
    - Validation and conflict resolution: For example, when modifying `Variable.upper` and `Variable.lower`, the new upper might conflict with the old lower, but not with the new one. We cannot validate this properly with setters.
5. `update()` is DRY-er:
    - Setting `_coeff_dirty` and potentially `*.flags.writeable` can live in one method instead of scattered over setters.

FBumann:
@FabianHofmann Updated my comment a bit

FabianHofmann:
@FBumann sounds good. I'll have a look at the pr this afternoon (tomorrow I will have more time to work on this in general)

coroa:
I don't understand the intricacies of model snapshots yet, so can't comment on the weakref proposal. mark's comment is close to what i had in mind with a context manager for making internal modifications.

i'd hope the user does not have to specify whether she is passing the same model.

i like the .update proposal, especially for cases where multiple things are updated. I'd keep the getters and allow the user to make in place updates but only when using the context manager which could set a dirty mark.

MaykThewessen:
@coroa @FBumann @FabianHofmann The context-manager framing reconciles the A-vs-B split cleanly — the context manager *is* the consent signal, so we keep the fast path by default without trusting undocumented discipline.

Sketch:

```python
with c.editing() as arr:        # thaws .values, no dirty flag yet
    arr.coeffs.values[mask] = ...
# __exit__: arr.flags.writeable = False; c._coef_dirty = True
```

- Outside the CM: arrays frozen (`writeable=False`), so an accidental `c.coeffs.values[...] = x` raises immediately — Joris' footgun guard.
- Inside: deliberate in-place edits allowed, and exit arms `_coef_dirty` for free. No user-facing `same_model`.
- `.update(...)` stays the ergonomic path for typed multi-field edits (rhs + sign atomic) and routes through the same dirty funnel.
- Snapshot drops `same_model` from its public signature; uses the weakref-to-source + dirty flag to pick fast-vs-full compare internally.

Open edges to close (from the writeable-flag audit upthread):
1. **Internal paths** that touch `.values[...]` (sanitize_zeros, SOS reformulation, …) must run inside the CM or a private thaw — needs the audit Felix flagged.
2. **`.view()` re-arm** stays possible; the CM stops casual mutation, not adversarial. Acceptable.
3. **Lifecycle**: rebuild / close / pickle must re-thaw or re-freeze consistently — one place, not scattered.
4. **Dask buffers** ignore `writeable`; mixed-backend models get inconsistent enforcement. Document as a numpy-path-only guarantee.

If that shape works I can take the CM + freeze plumbing as a follow-up so #718 lands on the current API and the freeze arrives non-breaking.

FBumann:
@MaykThewessen Im not sure if the complexity and instability of manipulating `.values` directly is worth it. Does anyone have the time to create some sort of benchmkark to check about what kind of performance gain (absolute) we are talking here? The new update api already improves performance by miles, so id rather focus on stability instead of the last percentages of performance. 
Is it actually a meaningful performance gain?

FBumann:
~Blocked by #732~

MaykThewessen:
Benchmarked it. Isolated the mutation step (no solve) — typed constraint setter (`con.rhs = arr` / `con.coeffs = arr`, the same path `.update()` funnels through) vs raw in-place `con.values[...]`. linopy master, M1, median of 7 warm updates, chunk 0 excluded, 15% of rows mutated per chunk, 6 nnz/row.

| n_rows | nnz | RHS typed | RHS raw | coef typed | coef raw |
|---:|---:|---:|---:|---:|---:|
| 200 k | 1.2 M | 7.7 ms | 0.03 ms | 0.97 ms | 0.14 ms |
| 800 k | 4.8 M | 19.9 ms | 0.09 ms | 4.16 ms | 0.55 ms |
| 2.0 M | 12 M | 43.8 ms | 0.23 ms | 10.7 ms | 1.43 ms |

The 2 M row / 12 M nnz line matches our production LP (1.14 M cols × 1.97 M rows). Worst case — RHS + coeffs both updated — is **~52 ms typed vs ~1.7 ms raw, so ~51 ms/chunk saved by the raw path**. Against a ~12 s solve that's **~0.4%**; across a 52-chunk full-year run, ~2.7 s total.

Decomposed the typed cost to confirm it's the setter, not a benchmark artifact: assigning a prebuilt `DataArray` is 6.95 ms @ 200 k (the alignment inside the setter), the array `.copy` is only 0.15 ms. So the number is real setter work, and it's O(rows) — bounded and predictable, not a tail risk.

So I agree with @FBumann: the raw-`.values` escape hatch isn't worth it. ~0.4% of solve doesn't justify the freeze / `editing()` context manager, the internal-path audit, the dask-buffer inconsistency, and the lifecycle thaw/re-freeze plumbing. **Withdrawing my freeze-CM suggestion** — typed `.update()` / setters as the single mutation path is the right scope for this PR. Stability over the last fraction of a percent.

Bench script: `benchmark/bench_linopy_mutation_path_typed_vs_raw_values.py` (isolates mutation only, no solve), happy to upstream into linopy's `benchmark/` if useful.

FabianHofmann:
> Benchmarked it. Isolated the mutation step (no solve) — typed constraint setter (`con.rhs = arr` / `con.coeffs = arr`, the same path `.update()` funnels through) vs raw in-place `con.values[...]`. linopy master, M1, median of 7 warm updates, chunk 0 excluded, 15% of rows mutated per chunk, 6 nnz/row.
> 
> n_rows	nnz	RHS typed	RHS raw	coef typed	coef raw
> 200 k	1.2 M	7.7 ms	0.03 ms	0.97 ms	0.14 ms
> 800 k	4.8 M	19.9 ms	0.09 ms	4.16 ms	0.55 ms
> 2.0 M	12 M	43.8 ms	0.23 ms	10.7 ms	1.43 ms
> The 2 M row / 12 M nnz line matches our production LP (1.14 M cols × 1.97 M rows). Worst case — RHS + coeffs both updated — is **~52 ms typed vs ~1.7 ms raw, so ~51 ms/chunk saved by the raw path**. Against a ~12 s solve that's **~0.4%**; across a 52-chunk full-year run, ~2.7 s total.
> 
> Decomposed the typed cost to confirm it's the setter, not a benchmark artifact: assigning a prebuilt `DataArray` is 6.95 ms @ 200 k (the alignment inside the setter), the array `.copy` is only 0.15 ms. So the number is real setter work, and it's O(rows) — bounded and predictable, not a tail risk.
> 
> So I agree with @FBumann: the raw-`.values` escape hatch isn't worth it. ~0.4% of solve doesn't justify the freeze / `editing()` context manager, the internal-path audit, the dask-buffer inconsistency, and the lifecycle thaw/re-freeze plumbing. **Withdrawing my freeze-CM suggestion** — typed `.update()` / setters as the single mutation path is the right scope for this PR. Stability over the last fraction of a percent.
> 
> Bench script: `benchmark/bench_linopy_mutation_path_typed_vs_raw_values.py` (isolates mutation only, no solve), happy to upstream into linopy's `benchmark/` if useful.

great @MaykThewessen! very helpful. let's follow the easier route then and add non-supported value setting logics in the documentation

MaykThewessen:
Took a first cut at **A6 (incremental snapshot advance)** as a follow-up branch off this PR's head, so it's ready to rebase + open once #718 lands.

Branch: [`MaykThewessen:feat/snapshot-incremental-advance`](https://github.com/MaykThewessen/linopy/tree/feat/snapshot-incremental-advance) (commit `f64b265`).

### What

`ModelSnapshot.advance(diff, model)` replaces the full `capture` call in `Solver._update_locked`. It re-extracts only the containers in `diff.changed_variables` / `diff.changed_constraints`, patches `obj_c` at `diff.obj_c_indices`, updates `obj_sense`, and clears `_coef_dirty` on touched constraints. Unchanged containers' buffers are physically untouched (identity-preserved). First-time builds still go through `capture`.

### Bench (M1, median of 7, isolated snapshot-refresh step)

| n_rows    | n_containers | recapture | advance | speedup |
|----------:|-------------:|----------:|--------:|--------:|
| 4 000     | 4            | 0.36 ms   | 0.07 ms | 5.4x    |
| 100 000   | 10           | 3.43 ms   | 0.24 ms | 14.3x   |
| 1 000 000 | 20           | 30.24 ms  | 0.98 ms | 31x     |

Speedup scales as `(n_containers - 1) / n_containers`. Worst case (all updates in one container) is no regression: advance reduces to re-capturing that one container.

### Scope notes

- **Not true O(|diff|).** Still O(nnz in changed containers). True O(|diff|) needs an inverse map from global positions (used in `ModelDiff`) to local positions (used in `ContainerConBuffers`); fiddly enough that I'd defer it until a workload actually needs the last factor. Practical warm-update orchestrators already see >5x gain without it.
- **A2 overlap.** advance owns the `_coef_dirty` clear for the containers it touched. If A2 lands (move dirty-clear out of `capture`), the matching clear in `advance` stays in the same place under the same condition (post-successful-apply).
- **Reject path.** `advance` raises `ValueError` if called with a `rebuild_required` diff. Solver code already takes the `_rebuild` branch; the error is defense-in-depth.

### Tests

`test/test_persistent_snapshot_advance.py` (8 tests) — equivalence (`capture(post) == capture(pre).advance(diff, post)` for RHS, bounds, objective-linear, combined), identity of unchanged buffers, empty-diff no-op, rebuild-required raises, `_coef_dirty` cleared. Full persistent suite green: 66/66.

Happy to open this as a real PR against `main` the moment #718 merges, or rebase onto a different base if that's easier. Or fold the change into #718 directly if @FabianHofmann prefers — small enough surface.

MaykThewessen:
@FabianHofmann happy to fold the A6 commit (`f64b265` on [`MaykThewessen:feat/snapshot-incremental-advance`](https://github.com/MaykThewessen/linopy/tree/feat/snapshot-incremental-advance)) directly into this PR — keeps the snapshot-refresh path consistent at landing time and avoids a stacked PR waiting on this one.

Two ways:

1. Cherry-pick `f64b265` onto your `feat/solver-update` branch yourself — single commit, no conflicts against the current PR head.
2. Enable "Allow edits from maintainers" on the PR (or grant me push to your branch) and I'll push it.

Either works. Let me know which you prefer.

linopy/persistent/diff.py
```diff
@@ -0,0 +1,724 @@
...
+def _cat_obj(parts: list[np.ndarray]) -> np.ndarray:
+    if not parts:
+        return _EMPTY_KIND
+    return np.concatenate(parts)
+
+
+def _cat_str(parts: list[np.ndarray]) -> np.ndarray:
+    if not parts:
+        return _EMPTY_U1
+    return np.concatenate(parts)
```
coroa: The _cat variants are all the same duplicated code.

linopy/persistent/diff.py
```diff
@@ -0,0 +1,724 @@
...
+    con_coef_rows: np.ndarray = field(default_factory=lambda: _EMPTY_I32)
+    con_coef_cols: np.ndarray = field(default_factory=lambda: _EMPTY_I32)
+    con_coef_vals: np.ndarray = field(default_factory=lambda: _EMPTY_F64)
+
+    con_rhs_indices: np.ndarray = field(default_factory=lambda: _EMPTY_I32)
+    con_rhs_values: np.ndarray = field(default_factory=lambda: _EMPTY_F64)
+    con_rhs_signs: np.ndarray = field(default_factory=lambda: _EMPTY_U1)
+
+    con_sign_indices: np.ndarray = field(default_factory=lambda: _EMPTY_I32)
+    con_sign_values: np.ndarray = field(default_factory=lambda: _EMPTY_U1)
```
coroa: you initialize all of these to the same objects, effectively bypassing the default factory.

coroa: I think none of these should have a default, they are all required. The only reason they aren't is because you want to have a full rebuild as part of them.

Let's make everything where we accept a ModelDiff also accept a RebuildReason and return the RebuildReason directly from the ModelDiff.from_snapshot as a sort of error condition.

linopy/persistent/diff.py
```diff
@@ -0,0 +1,724 @@
...
+
+
+_EMPTY_I32 = np.empty(0, dtype=np.int32)
+_EMPTY_F64 = np.empty(0, dtype=np.float64)
+_EMPTY_U1 = np.empty(0, dtype="U1")
+_EMPTY_KIND: np.ndarray = np.empty(0, dtype=object)
+
+
+class RebuildReason(enum.Enum):
+    NONE = "none"
```
coroa: Don't think we need a separate None. A RebuildReason should already say, we need a rebuild

linopy/persistent/diff.py
```diff
@@ -0,0 +1,724 @@
...
+    return all(np.array_equal(a[k], b[k]) for k in keys)
+
+
+def _active_container_positions(var: Variable, var_l2p: np.ndarray) -> np.ndarray:
+    labels = var.labels.values.ravel()
+    active = labels[labels != -1]
+    return var_l2p[active].astype(np.int32, copy=False)
+
+
+def _diff_var_container(
```
coroa: `_diff_var_container` and `_diff_con_container` should live on the DiffBuilder and probably inline the push_con, push_var stuff.

linopy/persistent/diff.py
```diff
@@ -0,0 +1,724 @@
...
+            var_base_coords = _coord_snapshot(var_a) if check_coords else None
+            reason = _diff_var_container(
+                builder,
+                name,
+                var_b,
+                var_base_buf,
+                var_base_coords,
+                var_l2p,
+                ignored,
+                check_coords,
```
coroa: var_l2p, con_l2p, ignored, check_coords should probably be passed into the _DiffBuilder __init__

linopy/persistent/diff.py
```diff
@@ -0,0 +1,724 @@
...
+        data_diff = None
+    else:
+        data_diff = new_buf.data != base_buf.data
+        if data_diff.any():
+            nnz_per_row = np.diff(new_buf.indptr)
+            row_idx_per_nnz = np.repeat(np.arange(n_rows), nnz_per_row)
+            row_value_changed = np.zeros(n_rows, dtype=bool)
+            row_value_changed[row_idx_per_nnz[data_diff]] = True
+        else:
+            row_value_changed = np.zeros(n_rows, dtype=bool)
```
coroa: None would be sufficient

linopy/persistent/diff.py
```diff
@@ -0,0 +1,724 @@
...
+        return RebuildReason.SPARSITY
+    if not np.array_equal(new_buf.indices, base_buf.indices):
+        return RebuildReason.SPARSITY
+
+    n_rows = new_buf.active_labels.size
+    if n_rows == 0:
+        return None
+
+    if skip_coef_compare:
+        row_value_changed = np.zeros(n_rows, dtype=bool)
```
coroa: None would be sufficient

linopy/persistent/diff.py
```diff
@@ -0,0 +1,724 @@
...
+    sign_changed = new_buf.sign != base_buf.sign
+
+    if not (row_value_changed.any() or rhs_changed.any() or sign_changed.any()):
+        return None
+
+    coef_rows = coef_cols = coef_vals = None
+    if row_value_changed.any():
+        coef_rows, coef_cols, coef_vals = _expand_coefs_coo(
+            new_buf, con_l2p, row_value_changed
+        )
```
coroa: Are we sure that we want to expand to coo, rather than storing a reference to the new CSR together with the data_diff mask or the row mask.

linopy/persistent/diff.py
```diff
@@ -0,0 +1,724 @@
...
+    for i in idx:
+        s, e = int(indptr[i]), int(indptr[i + 1])
+        n = e - s
+        cols[cursor : cursor + n] = new_buf.indices[s:e]
+        vals[cursor : cursor + n] = new_buf.data[s:e]
+        cursor += n
+    return rows, cols, vals
+
+
+def _diff_objective(
```
coroa: should be a method on the _DiffBuilder too

linopy/persistent/snapshot.py
```diff
@@ -0,0 +1,172 @@
...
+) -> ContainerConBuffers:
+    csr, con_labels, b, sense = con.to_matrix_with_rhs(var_label_index)
+    return ContainerConBuffers(
+        indptr=csr.indptr.astype(np.int32, copy=True),
+        indices=csr.indices.astype(np.int32, copy=True),
+        data=csr.data.astype(np.float64, copy=True),
+        rhs=np.asarray(b, dtype=np.float64).copy(),
+        sign=np.asarray(sense, dtype="U1").copy(),
+        active_labels=np.asarray(con_labels, dtype=np.int64).copy(),
+    )
```
coroa, line 91: Why do we copy here? Don't we want to share memory if possible?

linopy/persistent/snapshot.py
```diff
@@ -0,0 +1,172 @@
...
+
+def _extract_var_buffers(var: Variable) -> ContainerVarBuffers:
+    labels_flat = var.labels.values.ravel()
+    mask = labels_flat != -1
+    return ContainerVarBuffers(
+        lower=np.ascontiguousarray(var.lower.values.ravel()[mask], dtype=np.float64),
+        upper=np.ascontiguousarray(var.upper.values.ravel()[mask], dtype=np.float64),
+        type=_variable_type(var),
+        active_labels=np.ascontiguousarray(labels_flat[mask], dtype=np.int64),
+    )
```
coroa, line 67: Do we need to care about contiguity? Shouldn't we store it unmasked to share memory?

linopy/variables.py
```diff
@@ -915,18 +915,100 @@ def lower(self) -> DataArray:
...
+        """
+        updates: dict[str, DataArray] = {}
+        own_dims = self.model.variables[self.name].dims
+        for name, val, ref in (
+            ("lower", lower, self.lower),
+            ("upper", upper, self.upper),
+        ):
+            if val is None:
+                continue
+            if not isinstance(val, CONSTANT_TYPES):
```
coroa: isinstance also works with union types, so 
```suggestion
            if not isinstance(val, ConstantLike):
```
should be fine and we don't need CONSTANT_TYPES/get_args

coroa (commented):
Ok, first set of comments. I still need to go over the solvers implementation.

Everything looks sane in general, i'd still try to make a couple of improvements:
1. `persistent/diff.py` can be cleaned up a bit. Nothing major, mostly moving stuff around. Maybe a preference to not use a COO constraint coeff diff.
2. `persistent/snapshot.py` and `constraints.py`: I'd like to make ModelSnapshots (or more explicitly their con and maybe var buffers) share the explicit underlying numpy objects with the Constraint objects in the case of CSRConstraints to save memory. coef_dirty is then not necessary anymore (for the CSR ones), because you can directly test whether it is still the same object. Arguments are memory and cpu efficiency.

FabianHofmann:
@coroa thanks, addressed in 11b02e6:

- `from_snapshot` / `from_models` now return `ModelDiff | RebuildReason` — `RebuildReason.NONE` and `ModelDiff.rebuild_reason` are gone, all `ModelDiff` fields are required (built only by `_DiffBuilder.finalize()`).
- `_diff_var/con_container` and `_diff_objective` are now `_DiffBuilder` methods with `push_*` inlined; `var_l2p`/`con_l2p`/`ignored` live in `__init__`; the `_cat` variants are one helper; the zero-filled `row_value_changed` arrays are `None` now. Also dropped the always-True `check_coords` knob.
- **Memory sharing**: `_extract_con_buffers` no longer copies. CSRConstraint snapshots share the stored arrays and the diff short-circuits on object identity (`a is b`) before comparing — O(1) for untouched frozen containers, and it replaces the unconditional `_coef_dirty` skip for CSR with a verified one. To make identity sound, `CSRConstraint.sanitize_zeros()` is now copy-on-write (it was the only in-place mutation path; everything else already rebinds).
- `isinstance(val, ConstantLike)` — done.

Two pushbacks:

1. **COO coef diff** (your diff.py:651 comment): kept. All four backends consume per-entry triplets (`changeCoeff` / `chgCoeff` / `chgmcoef` / `putaijlist`), so storing CSR + mask would just move the same expansion into four `apply_update` bodies. The COO arrays only cover changed rows, so memory is bounded by changed-row nnz.
2. **Var buffers stay masked copies** (snapshot.py:64): they're small (2 floats/var) and they're the array-based safety net from the earlier `.values[...]` discussion — several tests mutate `lower.values[...]` raw and rely on the copied baseline to detect it. Sharing would compare an array against itself and silently miss exactly those edits. For constraints the trade-off differs: mutable extraction copies inherently, and CSRConstraint is frozen by contract — hence sharing is safe there.

Solvers part of your review still pending from your side — no rush.

FabianHofmann:
@coroa thanks, I addressed most of the points. The refactoring was really needed (already pointed out by @FBumann). COO still looks like the right choice. 

> [!Note] 
> AI written


* `from_snapshot` / `from_models` now return `ModelDiff | RebuildReason` — `RebuildReason.NONE` and `ModelDiff.rebuild_reason` are gone, all `ModelDiff` fields are required (built only by `_DiffBuilder.finalize()`).
* `_diff_var/con_container` and `_diff_objective` are now `_DiffBuilder` methods with `push_*` inlined; `var_l2p`/`con_l2p`/`ignored` live in `__init__`; the `_cat` variants are one helper; the zero-filled `row_value_changed` arrays are `None` now. Also dropped the always-True `check_coords` knob.
* **Memory sharing**: `_extract_con_buffers` no longer copies. CSRConstraint snapshots share the stored arrays and the diff short-circuits on object identity (`a is b`) before comparing — O(1) for untouched frozen containers, and it replaces the unconditional `_coef_dirty` skip for CSR with a verified one. To make identity sound, `CSRConstraint.sanitize_zeros()` is now copy-on-write (it was the only in-place mutation path; everything else already rebinds).
* `isinstance(val, ConstantLike)` — done.

Two pushbacks:

1. **COO coef diff** (your diff.py:651 comment): kept. All four backends consume per-entry triplets (`changeCoeff` / `chgCoeff` / `chgmcoef` / `putaijlist`), so storing CSR + mask would just move the same expansion into four `apply_update` bodies. The COO arrays only cover changed rows, so memory is bounded by changed-row nnz.
2. **Var buffers stay masked copies** (snapshot.py:64): they're small (2 floats/var) and they're the array-based safety net from the earlier `.values[...]` discussion — several tests mutate `lower.values[...]` raw and rely on the copied baseline to detect it. Sharing would compare an array against itself and silently miss exactly those edits. For constraints the trade-off differs: mutable extraction copies inherently, and CSRConstraint is frozen by contract — hence sharing is safe there.

coroa:
> 1. **COO coef diff** (your diff.py:651 comment): kept. All four backends consume per-entry triplets (`changeCoeff` / `chgCoeff` / `chgmcoef` / `putaijlist`), so storing CSR + mask would just move the same expansion into four `apply_update` bodies. The COO arrays only cover changed rows, so memory is bounded by changed-row nnz.

could move into a property on the model diff

maybe instead of changing the coefficient. rewriting the constraint would be faster. unsure whether this would lead to having to change the constraint mapping.

linopy/solvers.py
```diff
@@ -879,9 +897,32 @@ def update(
...
+        mutates either Model.
+        """
+        snapshot = self.snapshot
+        if snapshot is not None:
+            return ModelDiff.from_snapshot(
+                snapshot, model, same_model=False, ignore_dims=ignore_dims
+            )
+        baseline = self.model
+        assert baseline is not None
+        return ModelDiff.from_models(baseline, model, ignore_dims=ignore_dims)
```
coroa: I would remove `_unlocked` from the name and use it from _update_locked

linopy/solvers.py
```diff
@@ -400,11 +421,22 @@ class Solver(ABC, Generic[EnvType]):
...
+      throw-away solver instances and high-level ``Model.solve(...)``.
+    * ``True`` — long-lived mode. A snapshot is captured at build time and
+      re-captured after each successful in-place update, enabling
+      diff-based ``solve(model=...)`` / ``update(model)`` across iterations.
     """
 
     model: Model | None = None
     io_api: str | None = None
     options: dict[str, Any] = field(default_factory=dict)
+    track_updates: bool = False
```
coroa, line 439: What's the memory impact of of the tracking? Is it small enough that a default of True would make sense?

linopy/solvers.py
```diff
@@ -437,10 +478,116 @@ def __post_init__(self) -> None:
...
+                ctx, diff.var_type_positions, diff.var_type_kinds
+            )
+        if diff.con_rhs_indices.size:
+            self._apply_con_rhs(ctx, diff)
+        if diff.con_sign_indices.size:
+            self._apply_con_signs(ctx, diff.con_sign_indices, diff.con_sign_values)
+        if diff.n_coef_updates:
+            self._apply_con_coefs(
+                ctx, diff.con_coef_rows, diff.con_coef_cols, diff.con_coef_vals
+            )
```
coroa, line 519: i know from discussions with @brynpickering that he did experiments (with Gurobi, i think) where deleting and re-creating constraint rows was often faster than updating coefs. Can we test this?

linopy/solvers.py
```diff
@@ -616,38 +790,196 @@ def _build_file(self, **build_kwargs: Any) -> None:
...
+        return ModelDiff.from_models(baseline, model, ignore_dims=ignore_dims)
+
+    def _update_locked(
+        self,
+        model: Model,
+        apply: bool,
+        ignore_dims: Iterable[str] = (),
+        disallow_rebuild: bool = False,
+    ) -> ModelDiff | RebuildReason:
+        if apply and not type(self).supports_persistent_update:
```
coroa: classvars are also resolved on the instance.
```suggestion
        if apply and not self.supports_persistent_update:
```

linopy/solvers.py
```diff
@@ -616,38 +790,196 @@ def _build_file(self, **build_kwargs: Any) -> None:
...
+        assert baseline is not None
+        return ModelDiff.from_models(baseline, model, ignore_dims=ignore_dims)
+
+    def _update_locked(
+        self,
+        model: Model,
+        apply: bool,
+        ignore_dims: Iterable[str] = (),
+        disallow_rebuild: bool = False,
+    ) -> ModelDiff | RebuildReason:
```
coroa: Why does `_update_locked` even have to be called when apply=False. The code flow in this function looks a bit convoluted.

coroa: Maybe the we need to pull out the diff computation.

linopy/solvers.py
```diff
@@ -1209,12 +1553,61 @@ class Highs(Solver[None]):
...
 
+    @classmethod
+    @functools.cache
+    def _vtype_map(cls) -> dict[VarKind, Any]:
+        return {
+            VarKind.CONTINUOUS: highspy.HighsVarType.kContinuous,
+            VarKind.BINARY: highspy.HighsVarType.kInteger,
+            VarKind.INTEGER: highspy.HighsVarType.kInteger,
+            VarKind.SEMI_CONTINUOUS: highspy.HighsVarType.kSemiContinuous,
+        }
```
coroa, line 1560: Shouldn't this be a ClassVar constant?

linopy/solvers.py
```diff
@@ -437,10 +478,116 @@ def __post_init__(self) -> None:
...
+                ctx,
+                diff.var_bounds_indices,
+                diff.var_bounds_lower,
+                diff.var_bounds_upper,
+            )
+        if diff.var_type_positions.size:
+            self._apply_var_types(ctx, diff.var_type_positions, diff.var_type_kinds)
+            self._reclamp_binary_bounds(
+                ctx, diff.var_type_positions, diff.var_type_kinds
+            )
```
coroa, line 511: Is it possible that some solvers need to know about the vartypes before they accept bounds?

linopy/solvers.py
coroa, line 1664: Should we try to re-use the _apply functions in the build_model or is this over kill?

linopy/solvers.py
```diff
@@ -1628,6 +2023,91 @@ def _build_solver_model(
...
+
+    def _apply_begin(self, var_label_index: Any, con_label_index: Any) -> Any:
+        gm = self.solver_model
+        gurobi_vars = gm.getVars()
+        gurobi_cons = gm.getConstrs()
+        if len(gurobi_vars) != var_label_index.n_active_vars:
+            raise UnsupportedUpdate("gurobi var count mismatch")
+        if len(gurobi_cons) != con_label_index.n_active_cons:
+            raise UnsupportedUpdate("gurobi con count mismatch")
+        return (gm, gurobi_vars, gurobi_cons)
```
coroa: namedtuple as context?

coroa (commented):
Ok, i also went over solvers. Not many comments.

I am only thinking about whether we can easily test whether removing and re-creating changed constraints might be faster for individual solvers? At least after some threshold of what was changed.

linopy/solvers.py
MaykThewessen: Worth measuring, and the code is set up well to test it. The coef path is a scalar loop today: HiGHS `changeCoeff` and Gurobi `chgCoeff` are called per element in `_apply_con_coefs`, so this is exactly the case where the delete-and-re-add-row approach would batch better. Xpress (`chgmcoef`) and Mosek (`putaijlist`) already submit the whole COO list in one call, so there's less to win there.

Concrete shape: the `_coef_dirty` mask already identifies the changed rows and the COO arrays are grouped by row, so we know the exact row set to swap. Add a threshold: if the share of dirty coefs in a row (or the dirty-row count) exceeds a cutoff, `delConstrs`/`deleteRows` those rows and re-add them from the new CSR instead of per-element coefficient changes; below the cutoff keep the current path.

One caveat to fold into the measurement: deleting rows invalidates the saved basis for them. For Gurobi that's mostly free, but for HiGHS row deletion is one of the things that forces presolve, so it interacts with the basis-reuse regression in follow-up #1. We'd want to compare net solve time, not just the modify-call cost.

I'd suggest wiring both paths plus a threshold sweep into the same rolling-horizon PyPSA-Eur benchmark that produced the follow-up findings, and tracking it as a follow-up rather than blocking this PR: the scalar loop is correct, just potentially suboptimal past some change fraction.

<sub>🤖 AI-edited (Claude) reply, reviewed by @MaykThewessen.</sub>

FBumann:
I would say we should get this working and release, and try to work on performance tweaks from there, with the soon to be added benchmark tools

linopy/persistent/diff.py
FabianHofmann: made it dtype aware and dropped duplicates

linopy/solvers.py
FabianHofmann: yes, done

linopy/solvers.py
FabianHofmann: this is valid. though I defer this to a follow up. the deletion of constraints interacts with the state of the solvers. for example for highs this will lead to a hard pre-solve which for some cases we want to avoid. we could introduced some more fine-grained control later

linopy/solvers.py
FabianHofmann: done

linopy/solvers.py
FabianHofmann: I would also say this is slightly an  overkill, there is some shared code, but I would say let's keep the two use cases clearly separate for now

linopy/solvers.py
FabianHofmann: I sent a agent to check for all solvers that we currently support to make sure, and there is no type +bound dependency except for binaries which we pack into the reclamp_ function to update binary bounds (which might still carry old bound after type conversion)

linopy/solvers.py
FabianHofmann: done

linopy/solvers.py
FabianHofmann: had it at true before but it still does a copy of quite some bit of the model and it would do so every time you run linopy. so I decided against it.

FabianHofmann:
@coroa @FBumann , if no objections I will merge this early next week (or earlier in case you confirm). so we can build on this regarding the decomposition stuff

PR #719, opened 2026-05-21
Feature/pandas coords

### LINKED ISSUE #709: add_variables: pandas Series/DataFrame bounds with missing dimensions silently drop the dimension

RobbieKiwi:
Closes #709

Fix the bug with pandas coords not being broadcast

## Checklist

### DIFF
--- a/doc/release_notes.rst
+++ b/doc/release_notes.rst
@@ -52,6 +52,7 @@ Most users should keep calling ``model.solve(...)``. If you want more control, y
 
 **Bug Fixes**
 
+* Setting pandas bounds with missing coords now broadcasts to target coords
 * SOS constraints on masked variables no longer cause solver-specific failures (Gurobi ``IndexError``, Xpress ``?404 Invalid column number``, LP parse errors, silent set corruption). ``Model.solve()`` and ``Model.to_file()`` now raise a clear ``NotImplementedError`` referring users to `#688 <https://github.com/PyPSA/linopy/issues/688>`__; pass ``reformulate_sos=True`` as a workaround.
 * ``Model.solve(..., reformulate_sos=True)`` now actually reformulates SOS constraints even when the solver supports them natively. Previously it was silently ignored with a warning.
 
--- a/linopy/model.py
+++ b/linopy/model.py
@@ -134,7 +134,7 @@ def _sanitize_pandas(arr: pd.Series | pd.DataFrame) -> DataArray | None:
         # A pandas dataframe, possible with multi-level columns and multi-level index
         # Unstack all layers of columns
         while isinstance(arr, pd.DataFrame):
-            arr = arr.unstack()  # type: ignore
+            arr = arr.unstack()
         if not isinstance(arr, pd.Series):
             # This should not happen
             logger.warning("Failed to unstack dataframe")
--- a/linopy/model.py
+++ b/linopy/model.py
@@ -126,13 +126,42 @@ def _coords_to_dict(
     return result
 
 
+def _sanitize_pandas(arr: pd.Series | pd.DataFrame) -> DataArray | None:
+    """
+    Attempt to convert the pandas series or dataframe into a datarray with named coords.
+    """
+    if isinstance(arr, pd.DataFrame):
+        # A pandas dataframe, possible with multi-level columns and multi-level index
+        # Unstack all layers of columns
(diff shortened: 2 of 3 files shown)


FBumann:
@RobbieKiwi many ci fails. We might only fix this in #717

FBumann:
@RobbieKiwi Thanks for the PR.

I would like to rewrite some of it to be more efficient:

```python
def _sanitize_pandas(arr: pd.Series | pd.DataFrame) -> DataArray | None:
    names = list(arr.index.names)
    if isinstance(arr, pd.DataFrame):
        names += list(arr.columns.names)
    if any(n is None for n in names):
        return None

    if isinstance(arr, pd.DataFrame):
        arr = arr.stack(list(range(arr.columns.nlevels)), future_stack=True)

    return arr.to_xarray()
```

Also, we need to take care that the dim order stays deterministic (see #719).
Id like to merge #710 first, then adjust it here so pandas gets transposed properly

FBumann:
@RobbieKiwi Due to the close coupling with #710 i decided thats its easiest to combine both into a single PR.
I've ported your tests and made some small changes to your code. If you want to I'd love to get your review on it.

PR #721, opened 2026-05-21
drop python 3.10 support

lkstrp:


FBumann:
No clue why pre-commit failed. maybe push an empty commit to make sure

PR #722, opened 2026-05-23
fix: broadcast pandas/DataArray bounds in coords and preserve dim order

### LINKED ISSUE #706: add_variables: DataArray bounds with missing dimensions yield wrong dimension order

FBumann:
Closes #706, #709. Partially addresses #723 (helper relocated to ``common.py``; call-site audit deferred to follow-up — see issue comment for scope).

## What changes

**Principle.** 0.7.0 made ``coords`` the source of truth for ``DataArray`` bounds in ``add_variables``. This PR closes the two remaining gaps so the rule holds across every bound type and dimension order:

- Pandas ``Series`` / ``DataFrame`` bounds missing a dimension in ``coords`` are now broadcast to ``coords`` instead of being silently dropped (#709).
- The variable's dimension order always follows ``coords`` regardless of bound type. Previously the order of expanded array bounds depended on the type of the bound (scalar bounds kept ``coords`` order; two array bounds missing the same dimension produced a prepended order, #706).

**Implementation.** ``_validate_dataarray_bounds`` and the downstream ``as_dataarray(arr, coords)`` call in ``add_variables`` collapse into a single helper, ``as_dataarray_in_coords`` (in ``linopy/common.py``, alongside ``as_dataarray``). It converts any input (pandas with named axes via ``to_xarray``, otherwise via ``as_dataarray``), validates the result against ``coords``, expands missing dims, transposes to coords order, and reconstructs the coord variables in that order. ``expand_dims({})`` and identity ``transpose`` are metadata-only no-ops, so scalar bounds and full-dim ``DataArray`` bounds already in ``coords`` order pay no extra cost.

**Pandas conversion.** A new ``_named_pandas_to_dataarray`` helper handles pandas with fully named axes (including ``MultiIndex`` on either axis) by stacking all column levels into the row index in one call: ``arr.stack(level=list(range(arr.columns.nlevels)), future_stack=True)``, followed by ``to_xarray()``. Pandas with any unnamed axis falls through to ``as_dataarray``.

**Piecewise.** Same family of fix: ``linopy.piecewise._broadcast_points`` built its ``expand_dims`` map from a ``set``, giving a hash-randomized dim order across processes. Iterates expressions and dims in declaration order now.

## Supersedes

- #710 — DataArray dim-order fix. Carried over.
- #719 — pandas broadcast fix. Carried over, but the test in #719 used ``set(var.dims) == {…}`` and didn't catch the dim-order regression; rewritten to assert tuple-equal dim order, with additional ``MultiIndex`` coverage and a one-liner for the same-values-different-order reindex branch.

## Test plan

- [x] ``test/test_variable.py::TestAddVariablesBoundsWithCoords`` — 57/57. Includes:
  - ``test_bound_broadcast_missing_dim`` parametrized over ``DataArray``, ``Series``, ``DataFrame``, ``Series-multiindex``, ``DataFrame-multicolumns``, ``DataFrame-multiindex``.
  - ``test_dataarray_broadcast_missing_dim_order`` parametrized over scalar/DA bound combinations.
  - ``test_dataarray_coord_reorder`` for the same-values-different-order reindex.
  - ``test_pandas_bound_with_unnamed_axis_falls_through``, ``test_unnamed_coords_short_circuit``, ``test_dataarray_bound_with_multiindex_coord`` covering edge branches in the new helpers.
- [x] ``test/test_piecewise_constraints.py::test_broadcast_points_dim_order_follows_exprs``.
- [x] Broader sweep: ``test_model.py``, ``test_linear_expression.py``, ``test_constraint.py``, ``test_constraints.py``, ``test_piecewise_constraints.py``, ``test_variable_assignment.py``, ``test_variables.py`` — 784 pass, no regressions.
- [x] Pre-commit (ruff/format/blackdoc/codespell) clean.


linopy/model.py
RobbieKiwi: Maybe it would make sense to give mask the same treatment as the bounds to avoid the same pandas issue. Then mask could have the same type hint as lower/upper bounds and the user could for example pass pd.DataFrame for all 3

linopy/model.py
```diff
@@ -126,27 +126,70 @@ def _coords_to_dict(
     return result
 
 
-def _validate_dataarray_bounds(arr: Any, coords: Any) -> Any:
+def _named_pandas_to_dataarray(arr: pd.Series | pd.DataFrame) -> DataArray | None:
```
RobbieKiwi: The code you suggested here I think is a good improvement
https://github.com/PyPSA/linopy/pull/719
One thing it might be worth checking for is that it seems index labels can be strings but could also be tuples. Maybe we should also check for this and return None if they are not all strings?
https://pandas.pydata.org/docs/reference/api/pandas.Index.names.html

RobbieKiwi (approved):

linopy/model.py
FBumann: It would! But this is somewhat breaking (we had a FutureWarning in place, but still...). I will split this into a separate PR!

linopy/model.py
FBumann: Yes!

linopy/model.py
FBumann: Agreed?

FBumann:
Superseded by a fresh PR that consolidates the full scope of this branch (#722 original + #725 + #726 + #729 + follow-ups) with a single comprehensive description. The branch `fix/bounds-coords-broadcast` is unchanged; only the PR thread is being reset for review clarity.

FBumann:
→ Reopened as #732 with consolidated scope and a single description.

Issue #723, opened 2026-05-23
Unify `as_dataarray` coords semantics with `_as_dataarray_in_coords`


FBumann:
Follow-up to #722 (which closes #706 and #709).

#722 introduces ``linopy.model._as_dataarray_in_coords`` as a localized helper for ``add_variables``: when ``coords`` is provided, the result is *guaranteed* to live in ``coords`` (dims, order, and values). Internally it converts the input to a ``DataArray``, validates dims/coords against ``coords``, expands missing dims, and transposes.

The underlying inconsistency it works around lives in ``linopy.common.as_dataarray``, whose ``coords`` argument means four different things depending on the input type:

| input | what ``coords`` does today |
|---|---|
| scalar / list / int / float | broadcasts to full ``coords`` |
| ``numpy.ndarray`` | labels positional axes |
| ``pandas.Series`` / ``DataFrame`` | **ignored** — dims come only from axis names |
| ``xarray.DataArray`` | **ignored** — input returned unchanged |

The new rule from #722 is uniform: *the result lives in ``coords``, regardless of input type.* This issue tracks unifying ``as_dataarray`` around that rule and deleting ``_as_dataarray_in_coords``.

## Why this wasn't part of #722

``as_dataarray`` is called from ~12 sites across ``expressions.py``, ``variables.py``, and ``model.py`` (mask, sign, rhs). Tightening its semantics globally risks regressions and benefits from an audit pass rather than being bundled into a bug-fix PR.

## Call sites to audit

- ``linopy/model.py`` — ``mask`` in ``add_variables`` / ``add_constraints`` (currently relies on ``broadcast_mask`` after the conversion), ``sign`` / ``rhs`` in ``add_constraints``.
- ``linopy/expressions.py`` (lines 341, 584, 613, 1105, 1668, 2002, 2030, 2154, 2289) — arithmetic ops pass ``coords=self.coords, dims=self.coord_dims`` to align an input against the expression's coords. The new rule would raise on extra dims; could surface latent bugs or break working code.
- ``linopy/variables.py:330, 1369``.

## Open questions

1. **Strict by default vs opt-in.** Make the new rule the default and add a ``strict=False`` escape, or add ``strict=True`` and migrate call sites one by one?
2. **Performance.** ``as_dataarray`` is hot. The validate/expand/transpose path is O(ndim) in the metadata-only case, but the validation loop still runs. Worth benchmarking ``model.solve()`` end-to-end on a representative workload before deciding.
[...]

PR #724, opened 2026-05-23
refactor: move as_dataarray_in_coords to common.py (#723 part 1)

FBumann:
First slice of #723. **Stacked on top of #722** — set base to ``fix/bounds-coords-broadcast``; rebase / change base to ``master`` once #722 merges.

## What changes

Pure relocation, no behavior change.

- ``_as_dataarray_in_coords`` and its two helpers (``_coords_to_dict``, ``_named_pandas_to_dataarray``) move from ``linopy/model.py`` to ``linopy/common.py``, alongside the ``as_dataarray`` they parallel.
- The main helper is renamed ``as_dataarray_in_coords`` (no leading underscore) so other modules can import it when migrating call sites.
- ``add_variables`` imports + uses the relocated function; no other call sites change in this PR.

Diff is 134 / -132 lines, almost all of which is the relocation.

## Audit of remaining ``as_dataarray(arr, coords=...)`` call sites

Not migrated in this PR. Per-site notes:

**model.py**

- ``add_constraints``: ``as_dataarray(sign)`` and ``as_dataarray(rhs)`` pass no ``coords`` — strict semantics don't apply. *No migration.*
- ``add_variables`` / ``add_constraints`` mask: ``as_dataarray(mask, coords=data.coords, dims=data.dims).astype(bool)`` followed by ``broadcast_mask(mask, data.labels)``. The strict helper would handle extra-dim and expand_dims steps, but ``broadcast_mask`` also emits a ``FutureWarning`` for sparse-coord NaNs. Migration needs that warning path threaded through (or moved into the helper). *Deferred — follow-up commit.*

**expressions.py** (lines 584, 613, 1105, 1668, 2154)

- Arithmetic ops pass ``coords=self.coords, dims=self.coord_dims`` to label scalar/array inputs, then call ``_align_constant`` for the actual alignment with join/reindex. The strict helper would raise on inputs carrying an extra dim that ``_align_constant`` would otherwise broadcast across. *Higher risk — defer until we have a test that pins the desired behavior for ``LinearExpression + DataArray_with_extra_dim``.*

**variables.py** (line 330)

- Same shape as the expression arithmetic sites. *Deferred with them.*

## Follow-up scope

A second PR after this lands:

1. Migrate the mask paths (move the ``broadcast_mask`` NaN-warning logic into ``as_dataarray_in_coords`` or compose them explicitly).
2. Decide whether expression arithmetic should validate against ``self.coords``. Likely needs a tightened semantic in the v1 convention work (#717) rather than an unannounced behavior change here.
3. Benchmark ``add_variables`` and ``add_constraints`` on a representative model (PyPSA earth?) and compare before/after.

## Test plan

- [x] ``test/test_variable.py`` — 54/54 pass (incl. the MultiIndex + reindex coverage from #722)
- [x] Broader sweep: ``test/test_model.py``, ``test/test_linear_expression.py``, ``test/test_constraint.py``, ``test/test_constraints.py``, ``test/test_piecewise_constraints.py``, ``test/test_variable_assignment.py``, ``test/test_variables.py``, ``test/test_quadratic_expression.py``, ``test/test_scalar_*.py``, ``test/test_sos_constraints.py`` — 838 pass / 4 skipped, no regressions.
- [x] Pre-commit (ruff/format/blackdoc/codespell) clean.

### DIFF
--- a/linopy/common.py
+++ b/linopy/common.py
@@ -9,7 +9,7 @@ from __future__ import annotations
 
 import operator
 import os
-from collections.abc import Callable, Generator, Hashable, Iterable, Sequence
+from collections.abc import Callable, Generator, Hashable, Iterable, Mapping, Sequence
 from functools import cached_property, partial, reduce, wraps
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Generic, TypeVar, overload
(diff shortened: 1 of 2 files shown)


FBumann:
Folded into #722 — the relocation + the cleaner stack(level=…) form for _named_pandas_to_dataarray are now part of that PR (commits bca89e7 and b28f3df). Closing as superseded.

PR #725, opened 2026-05-24
fix: apply coords-as-truth rule to mask in add_variables/add_constraints

FBumann:
Stacked on #722. Extends the same coords-as-truth rule to ``mask``.

- ``mask`` in ``add_variables`` / ``add_constraints`` now routes through ``as_dataarray_in_coords(mask, data.coords)``. Pandas ``Series`` / ``DataFrame`` masks missing a dim are broadcast like bounds; ``mask`` type hint on ``add_variables`` widens to ``MaskLike``.
- ``broadcast_mask`` had no other callers and is deleted.

**Breaking change** (delivering on the previously-emitted ``FutureWarning``, first released in v0.6.3 #580):

- Sparse-coord masks raise ``ValueError`` instead of silently filling missing entries with ``False``. Workaround: ``mask.reindex({...}, fill_value=False)``.
- Extra-dim masks raise ``ValueError`` instead of ``AssertionError``.

PR #726, opened 2026-05-24
refactor: unify as_dataarray; split broadcasting from coords validation

### LINKED ISSUE #723: Unify `as_dataarray` coords semantics with `_as_dataarray_in_coords`

FBumann:
Closes #723. Stacked on #725 (which is stacked on #722).

## What changes

`as_dataarray_in_coords` is folded into `as_dataarray` and split along the seam between "broadcast `arr` against `coords`" and "enforce the `coords` contract":

- **`as_dataarray(arr, coords)`** — the broadcasting primitive. For every input type, the result is aligned with `coords`: positional inputs (numpy / unnamed pandas / scalar) are labeled by position, shared-dim coords are reindexed when values are equal in a different order, dims present in `coords` but not in `arr` are expanded, and the result is transposed to `coords` order. **Extra dims and disagreeing value sets on shared dims pass through**, so xarray broadcasting in expression arithmetic keeps working.
- **`assert_compatible_with_coords(arr, coords)`** — the validation companion. Raises if `arr` introduces dims not in `coords` (was: `as_dataarray_in_coords`'s extra-dim raise) or if a shared dim has disagreeing coord values (was: its "do not match" raise).

`add_variables` and `add_constraints` now call `as_dataarray` followed by `assert_compatible_with_coords` for `lower` / `upper` / `mask`. The previous `as_dataarray_in_coords` helper is deleted.

`_coords_to_dict` now filters MultiIndex level coords out of `xarray.Coordinates` inputs, so the new strict-by-default path treats e.g. `station` (and not its derived `letter` / `num` levels) as the dim. This was already a latent issue once strict semantics governed `as_dataarray`'s coords arg.

## Audit summary (the 12 call sites listed in #723)

| Call site | Strictness | Behavior |
|---|---|---|
| `model.py:705/706` `lower`/`upper` in `add_variables` | strict | `as_dataarray` + `assert_compatible_with_coords` |
| `model.py:715` `mask` in `add_variables` | strict | same |
| `model.py:979` `mask` in `add_constraints` | strict | same |
| `expressions.py:584/613/1105/1668/2154` arithmetic | broadcast | call signatures unchanged; benefits from positional labeling + reindex for free, with no value-set raise on shared dims |
| `variables.py:330` `to_linexpr(coefficient)` | broadcast | unchanged |
| `expressions.py:341/2002/2030/2289`, `model.py:912/919/972`, `variables.py:1369` | no coords | unaffected |

## Breaking changes for end users

**This PR introduces no new user-visible breaking changes.** It's a structural refactor that keeps the strict contract at the `add_variables` / `add_constraints` call sites via `assert_compatible_with_coords`. The breaking changes a user sees when this PR lands on `master` are the union of what #722 and #725 already deliver:

### From #725 (mask handling)
| Today (master) | After this stack | Migration |
|---|---|---|
| Mask with sparse coords silently fills missing entries with `False` and emits `FutureWarning` | Raises `ValueError` | `mask.reindex({...}, fill_value=False)` |
| Mask with extra dims raises `AssertionError` | Raises `ValueError` | Same fix; only the exception type changed |
| Mask as `pd.Series` / `pd.DataFrame` missing a dim was sometimes silently dropped | Broadcast like bounds | None needed (strictly more cases work) |

### From #722 (bounds handling)
| Today (master) | After this stack |
|---|---|
| Pandas `Series` / `DataFrame` bound missing a dim is silently dropped | Broadcast to coords (#709 fix) |
| `DataArray` bound dim order depends on bound type (#706) | Always follows `coords` order |
| `DataArray` bound with extra dim was silently accepted | Raises `ValueError` |
| `DataArray` bound with mismatched coord values raised varied downstream errors | Raises `ValueError: ...coordinates do not match` |
| `Piecewise._broadcast_points` had hash-randomized dim order across processes | Iterates in declaration order |

### Refactor-only changes (no user-visible effect on the strict sites)

- `as_dataarray` itself no longer raises on extra dims or shared-dim value-set mismatch — every public site that used to depend on those raises now gets them from the explicit `assert_compatible_with_coords` call. Surface behavior at `add_variables` / `add_constraints` is preserved.
- Two existing tests updated to reflect the new "coords is source of truth, extras broadcast in" semantics of `as_dataarray` itself:
  - `test_as_dataarray_with_ndarray_coords_dict_set_dims_not_aligned`: extra coord entries now expand into the result (the test was pinning an undocumented "dims wins, coord entry dropped" corner of the old lax path).
  - `test_dataarray_extra_dims`: rewritten so the subset check fires (rather than the value-mismatch check) — the assertion still raises on extra dims, just via the new `assert_compatible_with_coords` helper.

## Test plan

- [x] `test/test_common.py` — added five tests pinning the new split (extra-dim preservation, disjoint shared-dim values, `assert_compatible_with_coords` extra-dim raise, value-mismatch raise, subset-dims allowed).
- [x] Full suite: **3698 passed, 29 skipped** (parity with base).
- [x] `pre-commit run --all-files` clean (ruff/format/blackdoc/codespell).
- [x] Microbenchmark (`dev-scripts/benchmark_as_dataarray.py`, untracked): flat timings vs the base branch on both add_variables-heavy (≈22 ms / 50–57 ms mean) and arithmetic-heavy (≈80–82 ms) workloads.


FabianHofmann:
@FBumann we have some conflicts here now, should I resolve them? I would still need to figure out where things branch off currently

FBumann:
I can

linopy/common.py
```diff
@@ -455,6 +399,184 @@ def as_dataarray_in_coords(arr: Any, coords: Any, **kwargs: Any) -> DataArray:
     return arr
 
 
+def assert_compatible_with_coords(
```
FabianHofmann: since we now often used the term validate for such kind of function, think about `validate_alignment`/`validate_alignment_with_coords`

FabianHofmann (approved):

linopy/common.py
FBumann: Definitely! Also removes the ambiguity about not raising an AssertionError!

FBumann:
@FabianHofmann I think #722 is kind of blocking here atm. Not hard, but will produce conflicts.
We should probably fold that in

FBumann:
@FabianHofmann Ready for final review.

EDIT: If you dont mind, i would merge this and give the final combined PR a proper review.

FabianHofmann (approved):

FabianHofmann:
> @FabianHofmann Ready for final review.
> 
> EDIT: If you dont mind, i would merge this and give the final combined PR a proper review.

already approved. sounds good to me

PR #727, opened 2026-05-24
Variable.update() / Constraint.update() as canonical mutation API

FBumann:
Stacked on #718. Exploratory branch for the API discussion (#718 review thread).

Adds typed ``.update()`` methods on ``Variable`` and ``Constraint`` so every mutation goes through one validated entry point. The existing 7 setters survive as one-line shims that forward to ``.update()``, so no user-visible breakage — but ``_coef_dirty``, broadcasting, sign normalisation, and the rhs-rearrange-to-lhs logic all live in one method.

## API

```python
Variable.update(*, lower=None, upper=None) -> Variable

Constraint.update(
    constraint=None,   # positional: a complete ConstraintLike, e.g. x + 5 <= 3
    *,
    lhs=None,          # replace whole expression
    rhs=None,          # set rhs (Variable/Expression rearranged onto lhs, like add_constraints)
    sign=None,         # set sign
    coeffs=None,       # partial: coefficient values
    variables=None,    # partial: variable labels
) -> Constraint
```

Two call shapes for Constraint, mirroring ``add_constraints``:

```python
c.update(x + 5 <= 3)               # complete replacement via a comparison
c.update(lhs=x, sign="<=", rhs=-2) # partial / explicit
```

Composition rules:
- ``lhs=`` replaces the whole expression first; ``rhs=`` rearrangement then sees the new lhs.
- ``lhs=`` is mutually exclusive with ``coeffs=`` / ``variables=`` (whole vs partial).
- Positional ``constraint`` is mutually exclusive with all keyword arguments.
- ``sign=`` applied last so it composes cleanly.
- Returns ``self`` for chaining.

## Setter shims

Forwarders, no other behaviour:

| setter | forwards to |
|---|---|
| ``var.lower = x`` | ``var.update(lower=x)`` |
| ``var.upper = x`` | ``var.update(upper=x)`` |
| ``c.lhs = expr`` | ``c.update(lhs=expr)`` |
| ``c.rhs = …``  | ``c.update(rhs=…)`` |
| ``c.sign = s`` | ``c.update(sign=s)`` |
| ``c.coeffs = x`` | ``c.update(coeffs=x)`` |
| ``c.vars = v`` | ``c.update(variables=v)`` |

## Closes the #718 review A1 residual

Last commit (``70dbed4``) flips ``ModelDiff.from_snapshot(same_model=…)`` default from ``True`` to ``False``. The flag-trust path (``skip_coef_compare = same_model and not coef_dirty``) is now precise through ``Constraint.update()`` — it sets the flag in one place; setters forward there. But ``c.coeffs.values[...] = ...`` still bypasses ``_coef_dirty``, and with the old default, that bypass silently produced wrong diffs.

The only production caller (``Solver._update_locked``) passes ``same_model`` explicitly, so it's unaffected. Same-model warm-update paths now value-diff the CSR data instead of trusting the flag — small perf cost (50–200 ms at Mayk-scale per his bench), correct by default. Solver-aware callers who own the mutation contract can opt back into the optimization with ``same_model=True``.

## What's NOT in this branch

- ``Variable.update(type=…)`` for binary/integer/continuous switching (would need new validation logic).
- ``Objective.update(...)`` (separate container; the persistent-solver diff already tracks ``obj_linear`` / ``obj_sense``).
- Setter removal. We chose to keep shims for back-compat; some setters predate #718 by ~4 years and removal is a bigger break worth its own PR.
- Migration of internal setter call sites (``sanitize_zeros`` and friends). Setters are pure shims so the migration would be cosmetic; postpone until / unless setters get actually removed.
- Batching the multiple ``assign_multiindex_safe`` calls a multi-attr update makes.

## Warts (inherited, not introduced)

- ``variables=`` kwarg shadows the ``.vars`` attribute spelling — picked the unambiguous name to dodge Python's ``vars()`` builtin shadow; ``.vars`` (read) stays as the attribute.
- ``rhs=Variable/Expression`` silently rearranges onto lhs (kept for symmetry with ``add_constraints``; documented).
- ``lhs=`` xor ``coeffs=`` / ``variables=`` is a runtime check, not a type-system constraint.
- Mixing positional + keyword is a runtime check too.

## Test plan

- [x] ``test_variable.py``, ``test_constraint.py``, ``test_constraints.py``, ``test_constraint_coef_dirty.py``, ``test_model.py``, ``test_variables.py`` — 280 pass (266 existing + 14 new for ``.update()`` directly, incl. the positional ``ConstraintLike`` form).
- [x] PR #718's persistent-solver suite: ``test_persistent_snapshot_diff.py``, ``test_persistent_snapshot_buffers.py``, ``test_persistent_solver_extras.py``, ``test_persistent_solver_orchestrator.py``, ``test_persistent_apply_update.py`` — 112 tests pass on this branch with the ``same_model=False`` default.
- [x] Pre-commit clean.

### DIFF
--- a/linopy/constraints.py
+++ b/linopy/constraints.py
@@ -55,7 +55,6 @@ from linopy.common import (
     maybe_group_terms_polars,
     maybe_replace_signs,
     replace_by_map,
-    require_constant,
     save_join,
     to_dataframe,
     to_polars,
@@ -1143,10 +1142,9 @@ class Constraint(ConstraintBase):
         return self.data.sign
 
     @sign.setter
-    @require_constant
     def sign(self, value: SignLike) -> None:
-        value = maybe_replace_signs(DataArray(value)).broadcast_like(self.sign)
-        self._data = assign_multiindex_safe(self.data, sign=value)
+        """Shim — forwards to :meth:`Constraint.update`."""
+        self.update(sign=value)
 
     @property
     def rhs(self) -> DataArray:
@@ -1154,15 +1152,86 @@ class Constraint(ConstraintBase):
 
     @rhs.setter
     def rhs(self, value: ExpressionLike | VariableLike | ConstantLike) -> None:
-        value = expressions.as_expression(
+        """
+        Set RHS. Constants go through :meth:`Constraint.update`; non-constant
+        rhs (Variable/Expression) is rearranged to LHS in place.
+        """
+        # Constant path: route through the canonical update API.
+        if not isinstance(
+            value,
+            variables.Variable
+            | variables.ScalarVariable
+            | expressions.LinearExpression
+            | expressions.QuadraticExpression,
+        ):
+            self.update(rhs=value)
+            return
+        # Non-constant rhs: existing rearrange-to-lhs behaviour (kept for
+        # backward compatibility; an explicit method would be cleaner).
+        expr = expressions.as_expression(
             value, self.model, coords=self.coords, dims=self.coord_dims
         )
-        residual = value.reset_const()
+        residual = expr.reset_const()
         if residual.nterm == 0:
-            self._data = assign_multiindex_safe(self.data, rhs=value.const)
+            self._data = assign_multiindex_safe(self.data, rhs=expr.const)
             return
         self.lhs = self.lhs - residual
-        self._data = assign_multiindex_safe(self.data, rhs=value.const)
+        self._data = assign_multiindex_safe(self.data, rhs=expr.const)
+
+    def update(
+        self,
+        *,
+        rhs: ConstantLike | None = None,
+        sign: SignLike | None = None,
+    ) -> Constraint:
+        """
+        Update the constraint's RHS and/or sign in place.
+
+        Canonical mutation API. Single-attribute setters (`c.rhs = …`,
+        `c.sign = …`) forward to this method.
+
+        Parameters
+        ----------
+        rhs : ConstantLike, optional
+            New constant RHS. Variable / Expression RHS is not supported
+            here; use the (legacy) ``c.rhs = expression`` setter, which
+            rearranges the residual into ``c.lhs``.
+        sign : SignLike, optional
+            New sign. One of ``"<=" / "==" / ">="`` (or their ``< > =``
+            aliases).
+
+        Returns
+        -------
+        Constraint
+            ``self`` for chaining.
+        """
+        if rhs is None and sign is None:
+            return self
+
+        if rhs is not None and isinstance(
+            rhs,
+            variables.Variable
+            | variables.ScalarVariable
+            | expressions.LinearExpression
+            | expressions.QuadraticExpression,
+        ):
+            raise TypeError(
+                "Constraint.update(rhs=...) only accepts constants; "
+                f"got {type(rhs).__name__}. Use the legacy `c.rhs = expr` "
+                "setter or rebuild via add_constraints(...) for "
+                "expression / Variable rhs."
+            )
+
+        updates: dict[str, DataArray] = {}
+        if rhs is not None:
+            new_rhs = DataArray(rhs).broadcast_like(self.rhs)
+            updates["rhs"] = new_rhs
+        if sign is not None:
+            new_sign = maybe_replace_signs(DataArray(sign)).broadcast_like(self.sign)
+            updates["sign"] = new_sign
+
+        self._data = assign_multiindex_safe(self.data, **updates)
+        return self
 
     @property
     def lhs(self) -> expressions.LinearExpression:
--- a/linopy/constraints.py
+++ b/linopy/constraints.py
@@ -1282,11 +1282,14 @@ class Constraint(ConstraintBase):
         coeffs : ConstantLike, optional
             Replace coefficient values (same sparsity / term structure).
             Lower-level than ``lhs=``; sets ``_coef_dirty``.
-        variables : Variable / DataArray, optional
+        variables : Variable, optional
             Replace variable label array (same sparsity / term
             structure). Lower-level than ``lhs=``; sets ``_coef_dirty``.
-            Mirrors the ``c.vars`` attribute; spelled out here to avoid
-            shadowing Python's ``vars()`` builtin in the kwarg name.
+
+            A raw ``DataArray`` of integer labels is still accepted
+            for back-compat but emits a ``FutureWarning`` — pass a
+            ``Variable`` instead. The DataArray path will be removed
+            in a future release.
 
         Returns
         -------
@@ -1350,11 +1353,21 @@ class Constraint(ConstraintBase):
         if variables is not None:
             from linopy.variables import Variable as _Variable
 
-            v = variables.labels if isinstance(variables, _Variable) else variables
-            if not isinstance(v, DataArray):
+            if isinstance(variables, _Variable):
+                v = variables.labels
+            elif isinstance(variables, DataArray):
+                warnings.warn(
+                    "Passing a DataArray to Constraint.update(variables=...) "
+                    "is deprecated and will be removed in a future release; "
+                    "pass a Variable instead.",
+                    FutureWarning,
+                    stacklevel=2,
+                )
+                v = variables
+            else:
                 raise TypeError(
-                    "Constraint.update(variables=...) expects a DataArray or "
-                    f"Variable; got {type(variables).__name__}."
+                    "Constraint.update(variables=...) expects a Variable; "
+                    f"got {type(variables).__name__}."
                 )
             new_vars = v.broadcast_like(self.coeffs, exclude=[self.term_dim])
             self._assign_data(vars=new_vars)
@@ -1469,10 +1482,12 @@ class Constraint(ConstraintBase):
     def sanitize_zeros(self) -> Constraint:
         """Remove terms with zero or near-zero coefficients."""
         not_zero = abs(self.coeffs) > 1e-10
-        return self.update(
-            variables=self.vars.where(not_zero, -1),
+        self._assign_data(
+            vars=self.vars.where(not_zero, -1),
             coeffs=self.coeffs.where(not_zero),
         )
+        self._coef_dirty = True
+        return self
 
     def sanitize_missings(self) -> Constraint:
         """Mask out rows where all variables are missing (-1)."""
--- a/linopy/constraints.py
+++ b/linopy/constraints.py
@@ -1152,36 +1152,13 @@ class Constraint(ConstraintBase):
 
     @rhs.setter
     def rhs(self, value: ExpressionLike | VariableLike | ConstantLike) -> None:
-        """
-        Set RHS. Constants go through :meth:`Constraint.update`; non-constant
-        rhs (Variable/Expression) is rearranged to LHS in place.
-        """
-        # Constant path: route through the canonical update API.
-        if not isinstance(
-            value,
-            variables.Variable
-            | variables.ScalarVariable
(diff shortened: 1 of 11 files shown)


odow:
This is much closer to how JuMP does things. We have `set_` functions instead of modify, but same thing.

 * https://jump.dev/JuMP.jl/dev/api/JuMP/#set_normalized_rhs
 * https://jump.dev/JuMP.jl/dev/api/JuMP/#set_normalized_coefficient
 * https://jump.dev/JuMP.jl/dev/api/JuMP/#set_objective_coefficient
 * https://jump.dev/JuMP.jl/dev/api/JuMP/#set_objective_sense
 * https://jump.dev/JuMP.jl/dev/api/JuMP/#set_integer

In MathOptInterface we have a CachingOptimizer which controls the "pass to solver OR rebuild" logic:

 * https://github.com/jump-dev/MathOptInterface.jl/blob/master/src/Utilities/cachingoptimizer.jl
 * Here's the "delete a variable if supported" logic: https://github.com/jump-dev/MathOptInterface.jl/blob/f0e9fef21118bc85b78717e134acc05a1798ec10/src/Utilities/cachingoptimizer.jl#L698-L724

I tried to cover the caching optimizer in my MOI talk, but I don't really like it. It was too brief and I did things in the wrong order: https://youtu.be/M31xoZGyj9w?si=NTQHUsnq7o9Lg770&t=2211

FBumann:
## Follow-up: what `.update()` should grow to cover

Mapping against the per-backend ``apply_update`` calls in #718 — the backends all expose more than this PR exposes today. Concrete TODO:

**Next milestones (additive, don't change existing shape):**

- [ ] ``Variable.update(type=…)`` — switch between ``"continuous"`` / ``"integer"`` / ``"binary"`` / ``"semi_continuous"``. All four backends expose it (``changeColsIntegrality`` / ``setAttr(VType)`` / ``chgcoltype`` / ``putvartypelist``). Validation: ``binary`` → bounds in {0,1}, ``semi_continuous`` → positive lower.
- [ ] ``model.objective.update(expression=, sense=)`` (or ``Model.update_objective(...)``) — backends expose ``changeColsCost`` / ``setAttr(Obj)`` / ``chgobj`` / ``putclist`` for linear coefficients and ``changeObjectiveSense`` / ``ModelSense`` / ``chgobjsense`` / ``putobjsense`` for sense. The persistent-solver snapshot already tracks ``obj_linear`` and ``obj_sense`` — only the user-facing mutation API is missing.

**Later, if a workflow surfaces:**

- [ ] Per-cell coefficient shortcut (``c.update(coefficient={x: 2.0})`` or ``c.update_coefficient(x, 2.0)``). Maps to ``changeCoeff`` / ``chgCoeff`` / ``chgmcoef``. Today ``coeffs=`` replaces the whole array.
- [ ] ``Variable.update(cost=…)`` shortcut for setting objective linear coefficients per variable. Canonical location is the objective; this would just be sugar.

**Intentionally out of scope — solvers don't support, so neither should ``.update()``:**

- Coefficient sparsity change → already a rebuild trigger; ``lhs=`` flips it implicitly.
- Quadratic objective in place → rebuild.
- Variable / constraint add / remove → use ``add_variables`` / ``remove_variables``.
- Mask flip → rebuild per #718.

Each item is additive — won't reshape the current ``.update()`` signature.

FBumann:
## API surface from a user perspective

Tying this back to the `_coef_dirty` / `same_model` discussion on #718: if `Variable.update()` / `Constraint.update()` become the **canonical** mutation path, the dirty flag becomes a reliable invariant and the leaky `same_model=True` parameter on `ModelDiff.from_snapshot` can go away.

Two follow-ups worth committing to here:

1. **Auto-detect identity in the diff** — store a `weakref` to the source model on `ModelSnapshot`. `from_snapshot` consults `_coef_dirty` only when identity matches. Users get one safe signature, no flags.

2. **Split `Solver.update(model, apply=True)` into `diff(model)` and `apply(diff)`** (per earlier discussion). `update` doing both is fine as sugar but the primitives should be separate — easier to reason about, easier to test, and removes the need to expose `ModelDiff.from_snapshot` / `from_models` publicly at all (their only legitimate caller becomes `Solver.diff`).

End state: users call `var.update(...)` to mutate, `solver.diff(model)` to inspect, `solver.apply(diff)` to push. No dirty flags, no `same_model`, no `from_snapshot` in the public surface.

FBumann:
With `update` being the user facing concept for updateing a linopy Model, we should rename som of the internals if they do sth related, but different:
  - Solver.apply_update → Solver.apply_diff
  - Solver._update_locked → Solver._apply_diff_locked

This results in a clear split:
- **Update**: Changes to the linopy Model
- **Diff**: The Difference between the Model a solver holds currently and another`linopy.Model`.

_Users do updates, we push DIffs internally to the solver._

FabianHofmann:
> ## API surface from a user perspective
> Tying this back to the `_coef_dirty` / `same_model` discussion on #718: if `Variable.update()` / `Constraint.update()` become the **canonical** mutation path, the dirty flag becomes a reliable invariant and the leaky `same_model=True` parameter on `ModelDiff.from_snapshot` can go away.
> 
> Two follow-ups worth committing to here:
> 
> 1. **Auto-detect identity in the diff** — store a `weakref` to the source model on `ModelSnapshot`. `from_snapshot` consults `_coef_dirty` only when identity matches. Users get one safe signature, no flags.
> 2. **Split `Solver.update(model, apply=True)` into `diff(model)` and `apply(diff)`** (per earlier discussion). `update` doing both is fine as sugar but the primitives should be separate — easier to reason about, easier to test, and removes the need to expose `ModelDiff.from_snapshot` / `from_models` publicly at all (their only legitimate caller becomes `Solver.diff`).
> 
> End state: users call `var.update(...)` to mutate, `solver.diff(model)` to inspect, `solver.apply(diff)` to push. No dirty flags, no `same_model`, no `from_snapshot` in the public surface.

I like that very much. let's do that. we can also make that another pr or adjust it directly in #718 - does not need to be added here. 

For now, I would like to keep the setters and **not** deprecate them. fine if I adjust that here?

FabianHofmann:
I you know what, let's deprecate them :)

FBumann:
@FabianHofmann As you wish :)
I read about pythons convention of removal:
 First, we deprecate a feature (Deprectation Warning, silent by default in python, users need to actively activate DeprecationWarning), then we change that to a Future Warning (Users see it), then we remove the feature, probably with v1.

SO its softer than a FutureWarning. Mostly a notice in the release notes.
Sounds good?

linopy/constraints.py
```diff
@@ -1171,14 +1181,188 @@ def lhs(self) -> expressions.LinearExpression:
...
+            )
+
+        # 2. rhs (rearranges non-constant part onto lhs).
+        if rhs is not None:
+            expr = expressions.as_expression(
+                rhs, self.model, coords=self.coords, dims=self.coord_dims
+            )
+            residual = expr.reset_const()
+            if residual.nterm != 0:
+                self._assign_lhs_expr(self.lhs - residual, rhs=expr.const)
```
FabianHofmann: reading this I would prefer the function name `self._assign_lhs`

linopy/constraints.py
```diff
@@ -1171,14 +1181,188 @@ def lhs(self) -> expressions.LinearExpression:
...
+            else:
+                self._data = assign_multiindex_safe(self.data, rhs=expr.const)
```
FabianHofmann: let's add a `self._assign_data` function for these cases

linopy/constraints.py
```diff
@@ -1171,14 +1181,188 @@ def lhs(self) -> expressions.LinearExpression:
...
 
+    def update(
+        self,
+        constraint: ConstraintLike | None = None,
+        *,
+        lhs: ExpressionLike | VariableLike | ConstantLike | None = None,
+        rhs: ExpressionLike | VariableLike | ConstantLike | None = None,
+        sign: SignLike | None = None,
+        coeffs: ConstantLike | None = None,
+        variables: variables.Variable | DataArray | None = None,
```
FabianHofmann, line 1227: let's add a deprecation warning if dataarray is passed as variables. we should only support variable type in future.

linopy/constraints.py
```diff
@@ -1171,14 +1181,188 @@ def lhs(self) -> expressions.LinearExpression:
...
+        if coeffs is not None:
+            new_coeffs = DataArray(coeffs).broadcast_like(
+                self.vars, exclude=[self.term_dim]
+            )
+            self._data = assign_multiindex_safe(self.data, coeffs=new_coeffs)
+            self._coef_dirty = True
+        if variables is not None:
+            from linopy.variables import Variable as _Variable
+
+            v = variables.labels if isinstance(variables, _Variable) else variables
```
FabianHofmann: somewhere here would be the deprecation warning I mentioned above

linopy/variables.py
```diff
@@ -915,18 +914,96 @@ def lower(self) -> DataArray:
...
+            coords, or if the resulting ``lower > upper`` anywhere.
         """
-        value = DataArray(value).broadcast_like(self.lower)
-        if not set(value.dims).issubset(self.model.variables[self.name].dims):
-            raise ValueError("Cannot assign new dimensions to existing variable.")
-        self._data = assign_multiindex_safe(self.data, lower=value)
+        if lower is None and upper is None:
+            return self
+
+        from linopy import expressions
```
FabianHofmann: this import should be moved to module imports

linopy/variables.py
```diff
@@ -915,18 +914,96 @@ def lower(self) -> DataArray:
...
+
+        non_constant = (
+            Variable,
+            ScalarVariable,
+            expressions.LinearExpression,
+            expressions.QuadraticExpression,
+        )
```
FabianHofmann: use ConstantLike from types.py to test

FabianHofmann: I would suggest to add a function `self._validate_update(lower=None, upper=None)` which checks for correct type and dimensions in a for loop. after validation they can be broadcasted, added and finally checked (final_lower < final_upper) here

FabianHofmann (changes requested):
wonderful @FBumann, some requests below

FabianHofmann:
> @FabianHofmann As you wish :) I read about pythons convention of removal: First, we deprecate a feature (Deprectation Warning, silent by default in python, users need to actively activate DeprecationWarning), then we change that to a Future Warning (Users see it), then we remove the feature, probably with v1.
> 
> SO its softer than a FutureWarning. Mostly a notice in the release notes. Sounds good?

sounds very good

linopy/constraints.py
FBumann: Or even a Future Warning? Is this currently a use case?

linopy/variables.py
FBumann: So _validate_update would NOT check (final_lower < final_upper) ?

FBumann:
@FabianHofmann 
Still todo, waiting for your take:
- Adding `self._validate_update(lower=None, upper=None)`: SHould it also do the lb<=ub cehck?
- variables: variables.Variable | DataArray | None = None: Deprecationwarning or Future Warning right away?

FabianHofmann:
> @FabianHofmann Still todo, waiting for your take:
> 
> * Adding `self._validate_update(lower=None, upper=None)`: SHould it also do the lb<=ub cehck?
> * variables: variables.Variable | DataArray | None = None: Deprecationwarning or Future Warning right away?

sorry, got swallowed by the other prs. yes and futurewarning

FabianHofmann (approved):

PR #728, opened 2026-05-27
refactor(types): unify SUPPORTED_CONSTANT_TYPES with ConstantLike alias

FBumann:
Drop the hand-rolled `SUPPORTED_CONSTANT_TYPES` tuple in `expressions.py`; use `CONSTANT_TYPES = get_args(ConstantLike)` from `linopy/types.py` instead. Single source of truth for "what counts as a constant".

The old tuple had drifted from the `ConstantLike` alias in two places, both narrowed back to the alias:

- **`np.bool_`**: only uniquely covered scalar `np.True_`/`np.False_` (Python `bool` passes via `int`, bool arrays via `np.ndarray`). Feeding one into an LP coefficient is almost always a bug — caller meant `int(...)`. `LinearExpression(np.True_, m)` crashed downstream on master anyway.
- **`np.number`**: silently subsumed `np.complexfloating`, so complex arrays slipped through and miscomputed.

Both now reject at the gate with a clear error.

### DIFF
--- a/linopy/common.py
+++ b/linopy/common.py
@@ -35,6 +35,7 @@ from linopy.constants import (
     sign_replace_dict,
 )
 from linopy.types import (
+    CONSTANT_TYPES,
     CoordsLike,
     DimsLike,
     SideLike,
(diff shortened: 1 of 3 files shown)


FabianHofmann (approved):

FabianHofmann:
@FBumann let's ignore the pypsa-eur CI for now

PR #732, opened 2026-05-27
refactor: unify coords-as-truth handling in add_variables/add_constraints

### LINKED ISSUE #706: add_variables: DataArray bounds with missing dimensions yield wrong dimension order

FBumann:
Closes #706, #709, #723.

## What changes

This branch makes `coords` the single source of truth for every shape-bearing argument of `add_variables` and `add_constraints` (`lower`, `upper`, `mask`), adds MultiIndex-level projection with a deprecation path, and rebuilds the conversion layer in `linopy.common` around two public functions. The work landed over several stacked sub-PRs (#722, #725, #726, #729, #733, #737); this umbrella PR consolidates them for a single review against `master`.

### 1. Bounds — `add_variables`

0.7.0 (#614) made `coords` the source of truth for `DataArray` bounds. This branch closes the remaining gaps so the rule holds for every bound type and dimension order:

- Pandas `Series` / `DataFrame` bounds missing a dimension in `coords` are now broadcast to `coords` instead of being silently dropped (#709).
- The variable's dim order always follows `coords` regardless of bound type (#706).
- `DataArray` bounds with an extra dim or mismatched coord values raise `ValueError` with a labelled message (`"lower bound: …"`).

`add_variables`' docstring now documents the coords-as-truth contract with doctests pinning the error messages.

### 2. Mask — `add_variables` and `add_constraints`

The same coords-as-truth rule applies to `mask`. The `FutureWarning` that `broadcast_mask` had been emitting since v0.6.3 (#580) is now in effect:

| Before (master) | After this PR | Migration |
|---|---|---|
| Sparse-coord mask silently fills missing entries with `False` + `FutureWarning` | Raises `ValueError` | `mask.reindex({...}, fill_value=False)` |
| Mask with extra dim raises `AssertionError` | Raises `ValueError` | Same fix; exception type changed |
| `pd.Series` / `pd.DataFrame` mask missing a dim was sometimes silently dropped | Broadcast like bounds | None needed |

`broadcast_mask` is removed.

### 3. MultiIndex-level projection (deprecated on arrival)

Inputs indexed by *levels* of a stacked-`MultiIndex` `coords` dimension are projected onto that dimension: a level subset broadcasts across the others, the full set aligns element-wise. This fixes PyPSA multi-investment arithmetic (an expression over a `(period, timestep)` `snapshot` MultiIndex times a `period`-indexed weighting).

Terminology: a stacked MultiIndex dim has **levels** (component index names, e.g. `period`/`timestep`) and **level combinations** (its elements — one tuple per position, e.g. `(2030, 't1')`).

**Decided in the #737 thread (scenario B): implicit projections are deprecated everywhere** — they emit `EvolvingAPIWarning` in arithmetic *and* in `add_variables`/`add_constraints`, and will raise under the v1 convention (#717, §8/§11). Coverage gaps (level combinations without a value) raise `ValueError` in `add_variables`/`add_constraints`, with the missing combinations listed in the error.

### 4. The conversion layer — `linopy.common`

The old `as_dataarray`/`as_dataarray_in_coords` pair is replaced by two public functions (developed and reviewed in #737):

```
as_dataarray(arr, coords, dims)                    convert: type dispatch + positional labeling
broadcast_to_coords(arr, coords, dims, *,          + broadcast: project MI levels, reorder,
                    strict=True, label=...)          expand missing dims, transpose to coords order
```

- **`strict=True`** (default): any mismatch with `coords` raises, naming `label` in the error. `label` is required in this mode (enforced via overloads and at runtime). Used by `add_variables` / `add_constraints`.
- **`strict=False`**: mismatches pass through for downstream xarray alignment (the operator's `join=` owns reconciliation). Used by expression arithmetic and `as_expression`.
- **`as_dataarray`** (convert only) is used by `__matmul__`, where broadcasting would wrongly contract dims: `a=(time,name) @ b=(name,location)` must give `(time,location)`, not `(location)`.

Mechanics and policy are separated: a private `_broadcast_to_coords` reports MultiIndex projections (`_LevelProjection`); the public function applies warn/raise policy per mode. Deprecation warnings live in one helper with a `TODO(#738)` to route through #717's `warn_legacy()` once it lands.

### 5. Coords-entry rules — `_coords_to_dict`

The canonical contract for how each `coords` form / entry is named. Docstring of `_coords_to_dict` carries the table; `TestCoordsToDictRules` mirrors it one-test-per-row.

| Entry | Naming source | Outcome |
|---|---|---|
| `pd.Index` with `.name` | `.name` | accepted |
| unlabeled entry | `dims[i]` | accepted |
| unlabeled entry, no `dims[i]` | — | skipped (xarray assigns `dim_0`) |
| `pd.MultiIndex` with `.name` | `.name` | accepted |
| `pd.MultiIndex` without `.name` | `dims[i]` | accepted (named on a copy) |
| `pd.MultiIndex` without `.name`, no `dims[i]` | — | `TypeError` |
| anything else (e.g. `DataArray`) | — | `TypeError` |

### 6. Piecewise, SOS, and small fixes

- `linopy.piecewise._broadcast_points` iterates expressions and dims in declaration order (was hash-randomized).
- SOS reformulation uses `var.indexes[d]` for reformulated bounds.
- Positional inputs aligned to coords with clear shape errors.

## Breaking changes

Compared to v0.7.0:

- Sparse-coord mask raises `ValueError` (was `FutureWarning` + silent fill, deprecated since v0.6.3).
- Mask with extra dims raises `ValueError` (was `AssertionError`).
- `MultiIndex` passed as a sequence-form coord entry must have `.name` set or be paired with `dims=[i]` (`TypeError` otherwise).
- Sequence-form coords entries can no longer be `DataArray`s (`TypeError`; pass `variable.indexes[dim]` instead).
- **New deprecation**: implicit MultiIndex-level projection warns everywhere (`EvolvingAPIWarning`) and will raise under v1. Affects PyPSA multi-investment.

## Known limitations (follow-ups)

- Inputs with **unnamed** MultiIndex levels bypass strict validation silently (see review comment below) — to be addressed in the MI follow-up.
- The deprecation's migration recipe needs documenting; the obvious form (`reindex` onto a stacked MI) is broken upstream (pydata/xarray#11368).
- Warning-channel unification with #717's `LinopySemanticsWarning` is tracked in #738.

## Sub-PRs merged into this branch

- #722 — bounds broadcast + dim order
- #725 — coords-as-truth for `mask`
- #726 — unify `as_dataarray`; split broadcasting from validation
- #729 — MultiIndex coord helpers
- #733 — coords-entry rules, MultiIndex + `dims=` parity, tuple normalization
- #737 — the conversion-layer architecture: `broadcast_to_coords(strict=…, label=…)`, scenario-B deprecation, MI terminology

## Test plan

- [x] Full suite: 3201 passed, 36 skipped (excluding solver/network tests). mypy + pre-commit clean. Doctests pass.
- [x] `TestCoordsToDictRules` — one test per row of the coords-entry rules table.
- [x] `test_common.py` — strictness-contrast tests, MI projection/warning/gap tests, mandatory-label contract.
- [x] `test_variable.py::TestAddVariablesBoundsWithCoords` / `TestAddVariablesMultiIndexCoords`.
- [x] `test_constraint.py` — constraint `rhs` setter regressions (missing-dim broadcast, MI-level projection).
- [x] `test_linear_expression.py` — matmul dim-contraction tests.


FBumann:
@FabianHofmann Although the breaking changes are quite small, id like to test this against pypsa and flixopt, to make sure they dont use any of the now raising paths.
Agreed?
Could you do pypsa?

FBumann:
@FabianHofmann This is the most relevant PR now i think, to continue working on the others

FabianHofmann:
reviewing now, we should be able to pull this in today

FBumann:
@FabianHofmann I tested this again with flixopt and im getting 0 failed tests

FabianHofmann:
small udpate: I am hitting some failures related to multi-indexing. but I should be able to resolve them quickly

FabianHofmann:
It is more complicated than I hoped. cannot make it today. will continue tomorrow

FBumann:
> It is more complicated than I hoped. cannot make it today. will continue tomorrow

I also struggle with multiindexes...

FabianHofmann:
@FBumann (sorry - busy again with projects) I think I have something. should I push it here and you have a look?

FBumann:
Yes

FBumann:
@FabianHofmann 
Two thoughts: 
1. Could some of the multiindex stuff be moved from as_dataarray() to validate_alignment() ? Only as a question, without an opinion from my side
2. Why use _as_dataarray_lax() in matmul? I wanted to not use it internally.

FabianHofmann:
> @FabianHofmann Two thoughts:
> 
> 1. Could some of the multiindex stuff be moved from as_dataarray() to validate_alignment() ? Only as a question, without an opinion from my side
> 2. Why use _as_dataarray_lax() in matmul? I wanted to not use it internally.

1. good question, let me take another look
2. example: `a` has dims (time, name) and `b` has (name, location). when running `a @ b`, b is going through as_dataarray with coords and dims alignment to `a`. that is, it is broadcasted for additional dimensions it does not cover (here `time`) and we get a b_intermediate with (time, name, location).  the resulting operation collapses all overlapping dimenions -> result only contains (location). I am happy to reconsider the general logic here. or add a flag in `as_dataarray`. (should have added a test too)

FBumann:
@FabianHofmann I think this surfaces a design weakness. I propose splitting the conversion logic once more. Ill do a quick PR for it.

FBumann:
> [!NOTE]
> Claude Code review finding — for the MI follow-up, not blocking this PR (the hole pre-dates the #737 merge).

## Known hole: unnamed-MultiIndex-level inputs bypass strict validation

A bound indexed by a MultiIndex with **unnamed levels** passes `strict=True` validation silently, even when its entries don't match coords:

```python
idx = pd.MultiIndex.from_product([[2020, 2030], ['t1', 't2']], names=('period', 'timestep'))
idx.name = 'snapshot'
coords = xr.Coordinates.from_pandas_multiindex(idx, 'snapshot')

# unnamed levels — note: no names=(...)
sparse_unnamed = pd.Series({(2020,'t1'): 1.0, (2030,'t2'): 2.0})

broadcast_to_coords(sparse_unnamed, coords, dims=['snapshot'], label='lower bound')
# → returns a 2-entry result, NO error — should raise (entries don't match the 4-entry coords)
```

With *named* levels the same input correctly raises:
```
ValueError: lower bound could not be aligned to coords: no value for 2 level combination(s)
of MultiIndex dimension 'snapshot': (2020, 't2'), (2030, 't1'). ...
```

**Cause**: the unnamed-level MI can't be matched to the coords' levels by name, so no projection happens; the input keeps its own 2-entry MultiIndex under the `snapshot` dim name, and `validate_alignment`'s MI branch doesn't catch the size/entry mismatch.

**Suggested fix direction** (for the MI follow-up): in `validate_alignment`, when both sides are MultiIndex-backed but with different lengths or entries, raise regardless of level names — the names matter for *projection*, not for *equality checking*.

Real users (PyPSA) always have named levels, so low practical impact — but "strict" shouldn't have a silent bypass.

FBumann:
@FabianHofmann can you look into the discovered MI gap?

FabianHofmann:
yes, running pypsa CI now and report back

FabianHofmann:
CI on pypsa runs through, great. just pushing a fix on MIs

FabianHofmann (approved):

FabianHofmann:
@FBumann let's merge this. I would suggest we make a follow up PR only doing some refactoring:

- move all broadcasting and alignment code from `common.py` to a new module `alignment.py` 
- move all related tests to `test_alignment.py` make more use of test classes to structure them better

PR #735, opened 2026-05-28
benchmarks: reusable registry, with models, patterns and phases, CI smoke and Codspeed

FBumann:
## Overhaul of the internal benchmark suite

Brings `benchmarks/` from a hand-wired pytest collection into a
typer-driven suite with cross-version sweeps, master-baseline CodSpeed CI,
and plotly views. Two orthogonal axes: **size** (scale a model up) and
**severity** (dial a density pathology at fixed size).

### Core pieces

- **Reusable spec registry.** `from benchmarks import REGISTRY` exposes
  11 self-registering model specs (LP / QP / MILP / SOS / piecewise /
  sparse networks / storage / PyPSA carbon) with `build(size)` + feature
  & phase tags. Adding a model is one file.
- **Density patterns (severity axis).** A parallel `PATTERNS` registry of
  5 specs that hold size fixed and sweep a `severity` 0–100 dial measuring
  **sparsity headroom** — the fill a sparsity-aware kernel could drop from
  linopy's dense `_term` rectangle (0 = the non-pathological shape, no
  fill; 100 = the realistic worst case). Each targets a known hot path:
  - `nodal_balance` — `groupby().sum()` skew (#745): 0 uniform → 100 one
    hub holds the rest
  - `kvl_cycles` — sparse `@` (#748): today's dot densifies to
    `nterm=n_branch` regardless of cycle sparsity, so memory is *flat*
    across severity — a latent sentinel that lights up when a sparse `@`
    lands
  - `merge_balance` — ragged `merge` (#749): one wide block pads every
    block to its width
  - `rolling` — `rolling(window).sum()` window width (1 → full horizon)
  - `cumsum` — `cumsum` triangular term growth
- **Phase coverage.** Five phases per spec: `build`, `matrices`,
  `lp_write`, `netcdf` round-trip, `solver_handoff` (highs/gurobi/
  mosek/xpress). Plus PyPSA end-to-end.
- **One CLI.** `python -m benchmarks --help` covers run / smoke / sweep
  / compare / plot / notebook / memory{save,sweep,compare}.
- **Cross-version sweeps.** `sweep 0.6.7 0.7.0` builds one uv venv per
  version; same flow for memory via `memray.Tracker` (model
  construction excluded from the peak).
- **Plotly views.** `plot <snapshots>` picks `scaling` (1), `scatter`
  (2, default), or `sweep` heatmap (3+). `--facets {phase, model}`
  splits across subplots. Auto-detects timing vs memory JSON.
- **Master-baseline CodSpeed + Dependabot loop.** CodSpeed (cachegrind,
  instruction-count regressions) runs on push to master plus a manual
  trigger (`codspeed.yml`), establishing the baseline — kept off every PR
  since the run is ~10–20× slower; regressions surface as master-to-master
  deltas. `[benchmarks]` pins perf-relevant deps individually (numpy/scipy/
  xarray/pandas/polars/dask/highspy/netcdf4); tooling deps grouped into one
  monthly PR. Each perf-relevant bump → one PR → one attributed CodSpeed
  delta on the next master run.
- **Notebook walkthrough** at `benchmarks/notebooks/registry_usage.ipynb`,
  executed in CI so examples can't rot.

### Two axes: size and severity

A `ModelSpec` sweeps `n` — make the same model bigger. A `PatternSpec`
sweeps `severity` 0–100 — hold the model's size fixed and dial up one
operation's *sparsity headroom* (the removable fill it forces into the
dense `_term` rectangle). Within a pattern, the operation's own axis
carries the cliff; the broadcast dims are sized only to lift peak memory
clear of the measurement noise floor, never to carry the signal. Both
implement the same `BenchSpec` protocol, so the phase drivers, CLI,
sweep, memory, and plotting machinery treat them uniformly; test ids
carry the axis (`[nodal_balance-severity=100]` vs `[basic-n=500]`). The
split keeps the question sharp: a regression on a *model* axis is "we got
slower at scale", on a *pattern* axis it's "a density idiom got worse".

### Sample output: 0.6.7 → 0.7.0

Default scatter view — each test at `(baseline_time, ratio)`, top-right
corner = real regressions (slow tests that got slower):

01-timing-scatter

Same comparison, memory dimension (peak RSS via `memray.Tracker`):

02-memory-scatter

Per-phase delta breakdown (`compare --facets phase`):

03-timing-compare-phase

### Going forward

- **Enable CodSpeed:** install the GitHub app on the org, add the repo,
  drop `CODSPEED_TOKEN` into Actions secrets. Until then the master-only
  `codspeed.yml` workflow is red, but the per-PR smoke job and notebook
  execution still run.
- **Add a model:** drop a file under `benchmarks/models/`, register it.
  Phase coverage and CLI surface follow automatically.
- **Add a pattern:** drop a file under `benchmarks/patterns/`,
  `register_pattern` it with a `severity → data shape` build. It joins
  every phase, sweep, and plot with no other wiring.
- **Add a phase:** one `test_<phase>.py` parametrized off the registry.
- **Bump pins:** `[benchmarks]` is the perf-attribution surface;
  `[project.dependencies]` stays loose so downstream consumers see no
  change.


FBumann:
It went so well that we can consider wiring up https://codspeed.io/. I would like to test it.
@FabianHofmann you would need to accept their app to be installed on the repo. But maybe a quick chat would be good

FBumann:
## Dogfooding: does this suite actually help during dev?

Short version: yes. I pointed it at a real, *currently-unreleased* linopy change — the CSR / `freeze_constraints` work ("30–120× faster matrix generation for direct solver APIs", "~10× faster `to_highspy`") — and it both **refused a false positive** and **isolated the real effect**.

### What happened

1. **No false positive.** A cross-version `sweep 0.6.7 0.7.0 --phase solver_handoff` came back **flat (~1.0×)**. Correct — the CSR work isn't in 0.7.0 (it's unreleased), so the suite reported *no change* instead of inventing one.
2. **Why I used `bench`, not `sweep`** (corrected — an earlier version of this comment blamed a highspy version mismatch; that was wrong). `sweep` pins `highspy==1.13.1` into *every* per-version venv, so the solver is held constant across arms — there's no highspy confound *within* a sweep. The real blocker is that the CSR win is gated on the per-**model** flag `freeze_constraints=True`, and **no registry model sets it** (`to_highspy` doesn't auto-freeze — proven below: mutable 18 ms vs frozen 1 ms on the *same* linopy). So a stock sweep of 0.7.0 → master would run the *mutable* path in both arms and show ~nothing. Demonstrating it via sweep would mean adding a frozen-constraints model variant (plus a git/file spec for the unreleased "after" arm — `sweep` accepts those, and now labels the snapshot by the spec's ref/sha so e.g. `git+…@<sha>` writes a clean `linopy-<sha>.json`). `bench` was simply the quickest way to toggle `freeze` in-process: **same linopy, same highspy, only the constraint representation differs.**
3. **End-to-end killed the caching confound.** Building the model once and timing only the handoff hides the freeze cost (it's paid eagerly at `add_constraints`). So I measured build-only and the *fused* build+handoff too — and the parts add up (mutable 24+18 ≈ 44, frozen 35+1 ≈ 37), a nice consistency check on `bench` itself.

### Results

Representative numbers (Apple M-series, highspy 1.12, the sparse many-term `balance` model at `n_buses=300`). **Absolute ms are machine-dependent; the ratios and their direction reproduce.**

| measurement | mutable | frozen (CSR) | |
| --- | --- | --- | --- |
| `to_highspy` matrix construction | 18 ms | 1 ms | **~16–21× faster** |
| model build (fresh) | 24 ms | 35 ms | ~1.7× *slower* (freeze paid upfront) |
| build + handoff, end-to-end | 44 ms | 37 ms | **~1.2× faster**, grows per extra handoff |

The magnitude tracks **terms-per-row**: dense few-term carbon-management ≈ 2.9×, `basic` ≈ 1.7×, this 300-term model ≈ 20× — exactly matching "for constraints with **many terms**". Net: freezing trades ~11 ms upfront build for ~17 ms saved per handoff → a single build+solve already wins slightly, and anything that touches the matrix again (LP write, re-export, iterative re-solve) compounds the win. "Frozen is slower" only holds if you build and never hand off — not a real workload.

### Reproduce

<details><summary><code>bench_csr.py</code> — self-contained, seeded, no pypsa</summary>

```python
import numpy as np
import pandas as pd
import xarray as xr

import linopy
import linopy.io as lio
from benchmarks import bench


def build_sparse(n_buses: int, *, freeze: bool) -> linopy.Model:
    rng = np.random.default_rng(42)
    n_lines, n_time = n_buses, min(n_buses, 24)
    buses = pd.RangeIndex(n_buses, name="bus")
    lines = pd.RangeIndex(n_lines, name="line")
    time = pd.RangeIndex(n_time, name="time")
    bus_from = np.arange(n_lines)
    bus_to = (bus_from + 1) % n_buses

    m = linopy.Model(freeze_constraints=freeze)
    gen = m.add_variables(lower=0, coords=[buses, time], name="gen")
    flow = m.add_variables(lower=-100, upper=100, coords=[lines, time], name="flow")
    incidence = np.zeros((n_buses, n_lines))
    incidence[bus_to, np.arange(n_lines)] = 1
    incidence[bus_from, np.arange(n_lines)] = -1
    incidence_da = xr.DataArray(incidence, coords=[buses, lines])
    demand = xr.DataArray(rng.uniform(10, 100, size=(n_buses, n_time)), coords=[buses, time])
    m.add_constraints(gen + (flow * incidence_da).sum("line") == demand, name="balance")
    m.add_objective(gen.sum())
    return m


def report(label, rs):
    by = {r.label: r.stats["min"] for r in rs.results}
    print(f"{label}: mutable={by['mutable']*1e3:.0f}ms frozen={by['frozen']*1e3:.0f}ms "
          f"-> frozen {by['mutable']/by['frozen']:.1f}x")


N = 300
mut, frz = build_sparse(N, freeze=False), build_sparse(N, freeze=True)

report("handoff   ", bench.compare({
    "mutable": lambda: lio.to_highspy(mut, set_names=False),
    "frozen":  lambda: lio.to_highspy(frz, set_names=False)}, rounds=20))
report("build     ", bench.compare({
    "mutable": lambda: build_sparse(N, freeze=False),
    "frozen":  lambda: build_sparse(N, freeze=True)}, rounds=20))
report("end-to-end", bench.compare({
    "mutable": lambda: lio.to_highspy(build_sparse(N, freeze=False), set_names=False),
    "frozen":  lambda: lio.to_highspy(build_sparse(N, freeze=True), set_names=False)}, rounds=20))
```

</details>

```bash
PYTHONPATH=. python bench_csr.py   # needs a linopy with Model(freeze_constraints=...)
```

So the suite catches real wins, refuses false ones, and attributes them to the right regime (terms-per-row). The reason this particular one needed `bench` rather than `sweep` is that the win is gated on a per-model flag (`freeze_constraints`) no registry model exercises — `bench` toggled it in-process on one linopy + one solver. Once CSR ships in a release (and/or a frozen model variant is added to the registry), a plain cross-version `sweep` would capture it directly.

---

*This investigation and write-up (including the measurements above) were produced with [Claude Code](https://claude.com/claude-code).*

FabianHofmann:
this is very nice! the pr is vast but I would be lenient here as this is not in the production code. on the software stack, @lkstrp does that align with your visions?

FBumann:
@FabianHofmann I would also say this should be different from prod code. I also thought about putting this into its own linopy-bench package instead, but this removes the possibility of wiring up codspeed...

FBumann:
@FabianHofmann @lkstrp Im very open about changing the framework overall. I think having such tools in hand however should greatly simplify judging some PR's and changes.

FBumann:
@FabianHofmann DO you think adding sth like CodSpeed to the codebase would be a nice addition? Im not sure, but we might just try it out...

FabianHofmann:
> @FabianHofmann DO you think adding sth like CodSpeed to the codebase would be a nice addition? Im not sure, but we might just try it out...

trying out sounds good @lkstrp what do you thing? don't know if we should enable it for the whole organization or only for linopy?

FBumann:
Im looking into sparsity a bit more. I need this benchmarking stuff to reliably improve memory usage while avoiding regression

FBumann:
@FabianHofmann Im preparing this to cover different sparsity patterns to make improvements about sparsity tracable

FBumann:
@FabianHofmann I think i will just merge this on my fork and try out codspeed there for now.

FBumann:
@FabianHofmann THe genuine outcome an be seen on a PR where i try to improve memory issues (not ready at all yet)
https://github.com/fluxopt/linopy/pull/25

FBumann:
Superseeded #771

PR #743, opened 2026-06-03
feat: add BaseExpression.has_terms property

### LINKED ISSUE #741: Add public per-slot emptiness check on expressions (`has_terms`) — PyPSA reaches into `.vars`/`_term` internals
## Motivation

PyPSA's nodal balance constraint needs to know which slots of an expression have no live variable terms, and currently has to reach into linopy internals to find out ([pypsa/optimization/constraints.py#L1166](https://github.com/PyPSA/pypsa/blob/master/pypsa/optimization/constraints.py#L1166)):

```python
lhs = merge(exprs, join="outer").reindex(name=buses)
...
empty_nodal_balance = (lhs.vars == -1).all("_term")

if empty_nodal_balance.any():
    if (empty_nodal_balance & (rhs != 0)).any().item():
        raise ValueError("Empty LHS with non-zero RHS in nodal balance constraint.")
    mask = ~empty_nodal_balance
else:
    mask = None
```

Buses with no attached components end up with all-dead terms (`vars = -1`) after the outer merge + reindex. PyPSA needs them to (a) raise on demand at a disconnected bus, and (b) mask those constraint rows out.

`vars` and `"_term"` are internal representation details. Downstream code depending on them couples PyPSA to linopy's storage layout, and shows a gap in the public API.

## Why the existing API doesn't cover it

- `expr.empty` is a single bool for the whole expression.
- `expr.isnull()` is `(vars == -1).all(helper_dims) & const.isnull()` — *absence* (no terms **and** no constant). In the merge + reindex pattern above it happens to give the same answer (reindexed slots carry a NaN constant), but it tests a different concept: a slot whose constant has been filled — e.g. after `fillna(0)` — is no longer "null", yet still has no terms. The two concepts are also explicitly distinct in the v1 convention (#717): dead terms at a present slot vs an absent slot.

So the check PyPSA actually means — *"this constraint row has no variables"* — has no public spelling.

## Proposed API

```python
@property
def has_terms(self) -> DataArray:
    """
    Boolean array, True at slots where the expression has at least one
    live term (vars != -1), reduced over the helper dims.
    """
```

on `BaseExpression`, so both `LinearExpression` and `QuadraticExpression` get it.

PyPSA's code then becomes internals-free:

```python
has_terms = lhs.has_terms
if not has_terms.all():
    if (~has_terms & (rhs != 0)).any().item():
        raise ValueError("Empty LHS with non-zero RHS in nodal balance constraint.")
    mask = has_terms
else:
    mask = None
```

## Relation to `isnull()`

`has_terms` and `isnull()` answer different questions and both have a place:

| slot state | `isnull()` | `~has_terms` |
|---|---|---|
| live terms | False | False |
| no terms, `const = NaN` (absent, e.g. after reindex) | True | True |
| no terms, `const = 0` (present slot, e.g. after `fillna(0)`) | False | True |

I'll follow up with a PR implementing this.

FBumann:
Closes #741.

Adds a public `BaseExpression.has_terms` property: a boolean array, true at slots with at least one live term (`vars != -1`), regardless of the constant.

## Motivation

PyPSA's nodal balance constraint currently reaches into linopy internals to find constraint rows with no variables ([pypsa/optimization/constraints.py#L1166](https://github.com/PyPSA/pypsa/blob/master/pypsa/optimization/constraints.py#L1166)):

```python
empty_nodal_balance = (lhs.vars == -1).all("_term")
```

With this PR it becomes:

```python
mask = lhs.has_terms
```

## Relation to `isnull()`

`isnull()` tests *absence* (no terms **and** no constant); `has_terms` ignores the constant. The two diverge at slots whose constant was revived by `fillna()` — present, but still without terms. Covered by `test_linear_expression_has_terms_ignores_const`.

## Changes

- `has_terms` property on `BaseExpression` (inherited by `LinearExpression` and `QuadraticExpression`), with doctest
- Tests: masking via `where()`, divergence from `isnull()` after `fillna()`, the outer-merge + reindex pattern, constant-only expressions, and quadratic expressions (linear terms with one factor `== -1` must count as live)
- API reference and release-notes entries

### DIFF
--- a/test/test_linear_expression.py
+++ b/test/test_linear_expression.py
@@ -1156,55 +1156,54 @@ def test_linear_expression_isnull(v: Variable) -> None:
     assert expr.isnull().sum() == 10
 
 
-def test_linear_expression_has_terms(v: Variable) -> None:
-    expr = np.arange(20) * v
-    assert expr.has_terms.all()
+class TestHasTerms:
+    """has_terms: true at slots with at least one live term, regardless of the constant."""
 
-    filter = (expr.coeffs >= 10).any(TERM_DIM)
-    masked = expr.where(filter)
-    assert_equal(masked.has_terms, filter.rename("has_terms"))
+    def test_basic_and_masking(self, v: Variable) -> None:
+        expr = np.arange(20) * v
+        assert expr.has_terms.all()
 
+        filter = (expr.coeffs >= 10).any(TERM_DIM)
+        masked = expr.where(filter)
+        assert_equal(masked.has_terms, filter.rename("has_terms"))
 
-def test_linear_expression_has_terms_ignores_const(v: Variable) -> None:
-    # has_terms differs from isnull() at slots whose constant was revived by
-    # fillna: no longer null, but still without terms
-    expr = np.arange(20) * v
-    filter = (expr.coeffs >= 10).any(TERM_DIM)
-    masked = expr.where(filter)
-    assert_equal(masked.isnull(), ~masked.has_terms)
+    def test_ignores_const(self, v: Variable) -> None:
+        # has_terms differs from isnull() at slots whose constant was revived by
+        # fillna: no longer null, but still without terms
+        expr = np.arange(20) * v
+        filter = (expr.coeffs >= 10).any(TERM_DIM)
+        masked = expr.where(filter)
+        assert_equal(masked.isnull(), ~masked.has_terms)
 
-    filled = masked.fillna(0)
-    assert not filled.isnull().any()
-    assert_equal(filled.has_terms, filter.rename("has_terms"))
+        filled = masked.fillna(0)
+        assert not filled.isnull().any()
+        assert_equal(filled.has_terms, filter.rename("has_terms"))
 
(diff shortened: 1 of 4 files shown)


FBumann:
@lkstrp This is mainly to get pypsa out of reaching into linopy internals, which will is a precaution for #717 (although this should not cause problems).

test/test_linear_expression.py
```diff
@@ -1156,6 +1156,57 @@ def test_linear_expression_isnull(v: Variable) -> None:
     assert expr.isnull().sum() == 10
 
 
+def test_linear_expression_has_terms(v: Variable) -> None:
```
FabianHofmann: again, group these added tests into a class for readability?

FabianHofmann (commented):
this makes very much sense

test/test_linear_expression.py
FBumann: Yes.

FBumann: Done — grouped into `TestHasTerms` (5 methods).

FabianHofmann (approved):

Issue #744, opened 2026-06-03
Decision before v1: first-class MultiIndex support vs. flat dims + auxiliary level coords


FBumann:
> [!NOTE]
> AI-written analysis (Claude Code, prompted by @FBumann). This is a scope question for #717 — the decision window closes when v1 ships, because afterwards any change here is a breaking change with a deprecation cycle.

## The question

Every difficult problem in the #732 → #737 → #742 → #717 chain traces back to one root: **linopy supports stacked `pd.MultiIndex` dimensions as a first-class data model.** Should v1 keep that — or replace it with a representation that needs none of the machinery?

This is the radical version of the "learn from the xarray community, which struggled a lot with MIs" argument that decided scenario B.

## What MultiIndex support costs

| Complexity that exists *only* for MI | |
|---|---|
| `_project_onto_multiindex_levels`, `_LevelProjection`, `_as_multiindex` | ~120 lines |
| MI branches in `_coords_to_dict`, `validate_alignment`, `_broadcast_to_coords` (expand-via-template) | ~80 lines |
| `assign_multiindex_safe` (~20 call sites), `get_dims_with_index_levels`, MI serialization in `coords_to_dataset_vars` | ~100 lines |
| §11's stacked-MI paragraph, the scenario A/B design discussion, pydata/xarray#11368 workarounds | weeks |
| `TestMultiIndexProjection` + MI tests across 6 files | ~400 lines |

Plus a permanent tax: every future feature must answer "and what about MultiIndex?".

## The alternative: flat dim + auxiliary level coords

The same information, no `pd.MultiIndex` anywhere — and **§11 already governs auxiliary coords**:

```python
# instead of:  snapshot = MultiIndex[(2020,t1), (2020,t2), (2030,t1), (2030,t2)]
snapshots = pd.RangeIndex(4, name="snapshot")
period    = xr.DataArray([2020, 2020, 2030, 2030], dims="snapshot", ...)
timestep  = xr.DataArray(["t1", "t2", "t1", "t2"], dims="snapshot", ...)
```

**Verified against the current #717 branch** (all snippets run):

```python
x = m.add_variables(coords=[snapshots], name="x")              # ✅ works
expr = (1 * x).assign_coords(period=period)                    # ✅ aux coords attach

# per-period weighting — same explicit recipe as the MI case:
w = xr.DataArray(weights[period.values].values, dims="snapshot", ...)
expr * w                                                       # ✅ works, no projection machinery involved

# groupby a level:
[...]

Issue #740, opened 2026-06-02
Benchmark comparison vs `polar-high` (Polars-backed LP/MIP eDSL on HiGHS) — thoughts?


MaykThewessen:
A new modelling library, [polar-high](https://github.com/nodal-tools/polar-high), was released recently by [Juha Kiviluoma](https://github.com/jkiviluo) (Nodal-Tools Oy). It targets the same niche as linopy: an indexed LP/MIP eDSL on top of HiGHS, but uses a Polars-backed kernel instead of xarray.

The docs include a [reproducible build / solve / memory comparison](https://nodal-tools.fi/polar-high/latest/compare/benchmark/) between polar-high, linopy, and Pyomo on indexed linear programs, all using HiGHS as the solver. The dense-LP test follows linopy's own benchmark, restricted to the linear case so polar-high can run it directly.

### Dense LP — build-only headline (single-thread, HiGHS time-limited to ~1 µs to isolate modelling-layer cost)

![dense LP build-only](https://nodal-tools.fi/polar-high/latest/assets/benchmark.png)

At N=3000 (~18M variables):

| Tool | Total time | Peak memory |
|---|---|---|
| polar-high (regular) | 87 s | 16.4 GB |
| polar-high (save_memory) | 131 s | 15.1 GB |
| linopy (`io_api="lp"`) | 300 s | 23.7 GB |

### Network LP — irregular topology (N nodes, 5N edges, T=168, build-only, 1 thread)

![network LP build-only](https://nodal-tools.fi/polar-high/latest/assets/benchmark_network.png)

At N=10000 (~8.4M variables):

| Tool | Total time | Peak memory |
|---|---|---|
| polar-high (regular) | 51 s | 11.5 GB |
| polar-high (save_memory) | 94 s | 12.2 GB |
| linopy (`io_api="lp"`) | 165 s | 14.1 GB |

### Replication: linopy's original benchmark format (3-panel, full HiGHS solve included)

![linopy-format replication](https://nodal-tools.fi/polar-high/latest/assets/benchmark_replication.png)

The replication is faithful: same dense LP, same closed-form optimum, HiGHS solves the same sparse matrix. linopy's published numbers reproduce within run-to-run noise.

### Summary

Reported ~2.3-3.5x faster than linopy on the build path with ~30% lower peak memory. The Polars-backed lazy/columnar internals are the architectural source of the gap; linopy is xarray/dense-numpy by default.

### Questions for the maintainers and community

[...]

PR #751, opened 2026-06-04
fix(groupby): group by non-dimension coordinate names; fast multi-key grouping by names (#750, #753)

### LINKED ISSUE #750: groupby() by the name of a non-dimension coordinate raises (xarray parity)
## Bug

Grouping a `LinearExpression` by an existing **non-dimension coordinate** fails, even though the equivalent `xarray.Dataset.groupby` works. Both the string form (`groupby("name")`) and passing the coordinate `DataArray` raise.

## Reproducer

```python
import pandas as pd, xarray as xr, linopy

m = linopy.Model()
snapshots = pd.RangeIndex(4, name="snapshot")
period = xr.DataArray(
    [2020, 2020, 2030, 2030], dims="snapshot",
    coords={"snapshot": snapshots}, name="period",
)
x = m.add_variables(coords=[snapshots], name="x")
expr = (1 * x).assign_coords(period=period)   # attach 'period' as an aux coord

expr.groupby("period").sum()   # ValueError: period already exists as coordinate or variable name.
expr.groupby(period).sum()     # KeyError: 'period'
```

Both should group by the `period` coordinate and return a `period`-dimmed expression, the same way `xr.Dataset.groupby("period").sum()` does.

## Current workaround

You have to detach the coordinate first, and (depending on the form) rename it:

```python
expr.drop_vars("period").groupby(period).sum()                          # works
expr.drop_vars("period").groupby(period.rename("inv_period")).sum()     # also works
```

## Cause

In `LinearExpressionGroupby.sum()` (`linopy/expressions.py`):

1. A **string group** is not in `non_fallback_types`, so `groupby("period")` skips the fast path and hits the xarray fallback, which fails re-attaching the non-dimension coordinate.
2. The fast path builds `names_to_drop` from the *dimension index* only (`coords.name`), never the **aux coordinate being grouped**. It then does `ds.rename({GROUP_DIM: final_group_name})`, which collides with the `period` coordinate that is still attached → the `ValueError`.

## Fix sketch

In `sum()`: (a) resolve a string group that names an existing coordinate to `self.data[name]` so it takes the fast path, and (b) add `final_group_name` to `names_to_drop` when it is still an attached coordinate. PR to follow.

## Environment

linopy master (`a74724f`), reproduced in a clean worktree.

### LINKED ISSUE #753: Make multi-key LinearExpression.groupby([names]) fast and flat (align with xarray's grouper API)
> [!NOTE]
> AI-written analysis (Claude Code, prompted by @FBumann), combining a linopy code trace with a dig through PyPSA's groupby usage. The code-level claims and the prototype output below are verified on the #751 branch; the PyPSA-side file references come from that investigation — please sanity-check before acting.

Multi-key grouping of a `LinearExpression` currently forces a choice: *fast* (pandas `DataFrame` grouper → stacked-MultiIndex output) **or** *separate dims per key* (list-of-names → slow xarray fallback). This issue makes `groupby([names])` **fast** while keeping its separate-dims output, without breaking the existing `DataFrame` contract.

## What each layer accepts as a grouper

| Grouper | plain xarray | linopy | PyPSA uses it? |
|---|---|---|---|
| dimension name — `groupby("snapshot")` | ✓ | ✓ (always worked; #751 moved it onto the fast path) | — |
| non-dim coordinate name — `groupby("carrier")` | ✓ | ✓ **fixed in #751** (was `ValueError`) + fast | — |
| list of names — `groupby(["carrier","bus"])` | ✓ | ✓ but **slow** (xarray fallback) | no |
| `DataArray` — `groupby(da)` | ✓ | ✓ (fast) | ✓ nodal balance (`constraints.py`, via `c._as_xarray`) |
| `Grouper` obj — `groupby(x=UniqueGrouper())` | ✓ | ✗ unsupported | — |
| pandas `DataFrame`/`Series` | ✗ `KeyError` | ✓ (fast, **stacked** output) | ✓ statistics (`expressions.py`) |

Note: the pandas `DataFrame`/`Series` grouper is a **linopy-only extension** — plain xarray rejects it (`KeyError`). PyPSA is also internally inconsistent: its constraint/nodal-balance code uses the idiomatic `DataArray` grouper, only the statistics layer uses the pandas one.

## The gap: fast XOR separate-dims for multi-key

| Spelling | Path | Speed | Output |
|---|---|---|---|
| `groupby(["period","season"])` (names) | xarray fallback | slow | `(period, season)` grid (separate dims) |
| `groupby(df[["period","season"]])` (DataFrame) | reindex fast path | fast | stacked-MultiIndex `group` dim |

Root cause in `LinearExpressionGroupby.sum`: `_resolve_group` resolves a *single* name to its coordinate (fast path), but a multi-name list is returned unchanged, fails the `non_fallback_types = (Series, DataFrame, DataArray)` check, and drops to the fallback. The only multi-key fast path is the `pd.DataFrame` branch, which emits a stacked `pd.MultiIndex`.

## Proposed direction (non-breaking)

Key the output shape on **how the group was supplied** — so the one fast path serves both spellings:

1. **`groupby([names])` → fast, separate dims.** Resolve a list of coord names to a value frame so it takes the reindex fast path:
   ```python
   if isinstance(group, (list, tuple)) and all(isinstance(g, str) and g in data.coords for g in group):
       group = data[list(group)].to_dataframe()[list(group)]   # tag this as "came from names"
   ```
   Then, **for the names spelling only**, `unstack` the stacked result back into separate dims. This reproduces today's fallback output *exactly* — verified on a sparse crossing (combo `(2020,'s')` absent):

   ```
   FALLBACK  list-of-names       : {'period': 2, 'season': 2, '_term': 2}   (slow)
   FAST      DataFrame (stacked) : {'group': 3, ...}  -> unstack('group') -> {'period': 2, 'season': 2, '_term': 2}   (fast)
             missing (2020,'s') cell -> vars [-1, -1], coeffs [nan, nan]    (filled, == fallback)
   ```

2. **`groupby(df)` (user-supplied `DataFrame`) → unchanged.** Keep the stacked-MultiIndex `group` output. The tested contract (`test_linear_expression_groupby_with_dataframe`, asserting `set(group.values) == set(MultiIndex.from_frame(groups).values)`) still holds — **no deprecation, no breaking change**.

Implementation notes:
- `sum` needs to know the frame came from names (a flag from `_resolve_group`, or handle the list branch directly), so the final `unstack` applies only to the names spelling.
- The names path builds a stacked MultiIndex internally and immediately unstacks it — a transient intermediate index, fine.
- The closing `unstack` densifies to the full grid — the *same* shape today's fallback produces, so parity, not a regression.

## Optional follow-up (not required by the above)

With a fast separate-dims list grouping available, PyPSA *could* migrate its statistics sites to the idiomatic, label-safe form for consistency with its constraint code — but it's no longer forced, since the `DataFrame` grouper keeps working:

```python
gx = grouping.to_xarray()                          # name-indexed, label-aligned (no positional footgun)
da = da.assign_coords(carrier=gx.carrier, bus=gx.bus)
da.groupby(["carrier", "bus"]).sum()               # now fast, separate dims
```

## Relationship to #751

This builds directly on #751 (which fixes #750):

- #750/#751 is specifically about grouping by a **non-dimension coordinate name** — `groupby("carrier")` raised `ValueError: carrier already exists as coordinate` before, works after (verified). Grouping by a *dimension* name always worked; #751 only moved it onto the fast path via `_resolve_group` (commit `5a9b2e5`).
- This issue extends that same resolver to the multi-name list case.
- #751's `TestGroupbyByAttachedCoordinate` matrix already pins the separate-dims output for the list spelling, so the target shape is established and tested.

#751 deliberately stops at single-key parity and does not depend on this; nothing here blocks it.

## Future directions (sparse representations)

The grid memory cost is fundamental to *dense* xarray: separate dims are a dense cartesian grid, so a sparse/correlated key crossing materialises mostly-fill cells (measured ~100× vs the DataFrame grouper for a diagonal crossing; see #740). Truly getting *separate dims and compact* needs a sparse representation. Options, roughly by payoff vs cost:

1. **`sparse.COO` duck-arrays under xarray** — separate dims *and* observed-only storage. Requires linopy to be sparse-aware end to end (solver export, `matmul`, `merge`).
2. **Long-format / polars kernel (#756)** — groupby becomes a relational `group_by().agg()`, inherently sparse; no grid, no MultiIndex. Cleanest long-term; biggest change.
3. **Integer-code segment-sum** into a single observed-combo dim with the keys as plain coords — avoids both the grid and the MultiIndex, but is non-idiomatic (no `.sel`).
4. **Lazy / fused grouping** — keep the compact form and only densify if the result is indexed by label.

Until a sparse kernel lands, the two-shape split above (grid via `[names]`, compact MultiIndex via `df`) plus the blow-up warning is the pragmatic dense-xarray answer. This is also evidence for the #756 kernel decision: groupby's grid blow-up, `matmul`'s zero-padding (#748), and `merge`'s padding (#749) all dissolve together under a sparse/long kernel.

## See also

- #730 (open) — `.vars`→`.variables` rename: same xarray-API-unification family.
- #741, #752 (closed) — "PyPSA reaches into linopy internals": same PyPSA/linopy alignment theme.

Surfaced while reviewing #751.

FBumann:
## What

Brings `LinearExpression.groupby` to parity with `xarray.Dataset.groupby` for grouping by coordinate names — on the fast path, with no breaking change. Two related pieces:

**1. Group by an attached non-dimension coordinate (closes #750).**

```python
expr = (1 * x).assign_coords(period=period)   # 'period' rides on the 'snapshot' dim
expr.groupby("period").sum()   # before: ValueError: period already exists as coordinate
expr.groupby(period).sum()     # before: KeyError: 'period'
```

Both now work and take the fast path, mirroring `xarray`. (Grouping by a *dimension* name always worked — only attached non-dim coordinates were broken.) The old workaround was to detach the coord first (`expr.drop_vars("period").groupby(period).sum()`).

**2. Multi-key grouping by names now takes the fast path (closes #753).**

```python
expr.groupby(["period", "season"]).sum()   # before: silently dropped to the slow xarray fallback
```

It now rides the reindex fast path and returns the same output — one dimension per key, **byte-identical** to the fallback, sparse fill cells included. The `pd.DataFrame` grouper is untouched and keeps its compact stacked-`MultiIndex` output, so this is non-breaking.

## How

In `LinearExpressionGroupby`:

- **`_resolve_group`** normalizes a key: unwrap a single-element list (`groupby(["period"])` → scalar), and resolve a string coord name to its coordinate so it takes the fast path.
- **`sum()`** drops every coordinate aligned to the grouped dimension before reshaping, so an attached aux coord (including the one being grouped by) no longer collides on the final `rename`.
- The **`groupby` property** detaches a free (non-indexed) coordinate before handing it to xarray — fixing the `use_fallback=True` path — but never a `MultiIndex` level.
- A **list of coord names** (1-D, sharing one dim) is gathered into a value frame to ride the fast path, then `unstack`ed back into one dimension per key.

## Memory

One dimension per key is a dense cartesian grid, so a sparse key crossing materialises mostly-fill cells (measured ~100× vs the compact `DataFrame` grouper for a diagonal crossing — see #740). A `UserWarning` nudges sparse/high-cardinality users to the `DataFrame` grouper; it reads the collapsed `MultiIndex` levels, so it is O(observed), not O(N), and fires *before* `unstack` allocates the grid. Getting separate dims *and* compact storage would need a sparse/long-format kernel — tracked in #757 (the groupby-densification follow-up) under the sparse-kernel umbrella #756.

## Tests

`TestGroupbyByAttachedCoordinate` asserts grouped `vars`/`coeffs` against **hard-coded** results on a deterministic model: single-key (name/`DataArray` × `use_fallback`), 2-D variable, dimension-coordinate-by-name, single-element list, `MultiIndex` level, and a `pytest.raises` row pinning that a list of `DataArray`s is unsupported. `TestMultiKeyFastPath` covers the multi-key fast path: fast == fallback (list/tuple, including a sparse crossing), separate-dims-not-stacked, sparse-combination-filled, DataFrame-grouper-stays-compact, and the blow-up warning (fires when sparse, silent when dense). Full suite green; ruff + mypy clean.

<details>
<summary>Context: relation to PyPSA usage and #744</summary>

This is a building block for the flat-dimension + auxiliary-level-coord direction discussed in **#744**. If `n.snapshots` becomes a flat dimension carrying `period` / `scenario` level coordinates (instead of a stacked `pd.MultiIndex`), aggregating an expression over a level becomes `expr.groupby("period").sum()` — exactly what this PR makes work, now also for multiple levels at once (`groupby(["period", "scenario"])`).

PyPSA still carries per-period workarounds citing a broken MultiIndex groupby (pydata/xarray#6836); that upstream bug is now fixed (verified on xarray 2025.9.0), but those comments are about MultiIndex grouping and per-period *rolling*, orthogonal to this PR.
</details>

Closes #750, #753. Drafted with an agent (Claude Code).


FabianHofmann:
@fbumann

> [!NOTE]
> The following content was generated by AI (Claude Code), posted on behalf of @FabianHofmann.

A follow-up commit added an `observed` parameter to `LinearExpressionGroupby.sum` (pandas-style, cf. the `observed=True` default flip in pandas 3.0) addressing the memory concern with multi-key grouping:

```python
expr.groupby(["period", "season"]).sum(observed=True)
```

keeps the result stacked over only the **observed** key combinations (a `MultiIndex` `group` dimension) instead of unstacking into one dimension per key, which materialises the dense cartesian grid. The default `observed=False` keeps xarray parity (fast path == fallback, byte-identical); the dense-grid `UserWarning` now points to `observed=True` instead of the `pd.DataFrame` grouper workaround. `observed=True` with `use_fallback=True` raises, since the xarray fallback cannot honor it.

The multikey gather and unstack epilogue were also extracted into helpers (`_multikey_value_frame`, `_unstack_multikey`), and the release notes gained a Features entry.

FBumann:
Amazing!

PR #765, opened 2026-06-05
Add same-machine linopy vs polar-high build+IO benchmark

MaykThewessen:
Follow-up to #740. The cross-tool numbers there were measured on different hardware, so only ratios were comparable. This adds a self-contained `benchmark/io_api_compare/` harness that runs all four columns through one process-per-cell harness, one HiGHS build, one machine:

- `linopy-lp` (build -> `.lp` file -> HiGHS reparse)
- `linopy-direct` (build -> in-memory highspy load)
- `polar` (polar-high regular)
- `polar-sm` (polar-high `save_memory=True`)

Same dense `basic_model` LP, HiGHS time-limited to isolate the modelling + IO path, peak RSS via `/usr/bin/time -l`.

### Results (macOS arm64, single thread, HiGHS 1.14.0, linopy 0.7.0, polar-high 2.4.5)

At N=3000 (~18M variables):

| Tool | build+IO time | peak RSS |
|---|---|---|
| `linopy-direct` | **33.7 s** | 15.6 GB |
| polar-high (regular) | 53.9 s | 12.8 GB |
| polar-high (save_memory) | 82.7 s | **11.0 GB** |
| `linopy-lp` | 94.4 s | 16.0 GB |
tool_compare

Ranking stable across N=500..3000.

### Takeaways

- **`io_api="direct"` reverses the time ranking** vs `lp`: it's the fastest of the four on build+IO, ~1.6x faster than polar-high regular on the same machine.
- **polar-high keeps a real memory edge** (~18% lower peak RSS; save_memory leanest at 11.0 GB).
- The modelling-layer gap #740 surfaced is a **memory** gap, not a speed gap.

Plot (`tool_compare.svg`/`.png`) and raw CSV are included. Happy to adjust scope, wire it into the Snakemake flow, or drop the committed artifacts if you'd rather keep `benchmark/` script-only.


FBumann:
I think polars-high is not mature enough to add it to our benchmark. But tis PR is highly apreciated as an artifact and measurement.

coroa:
Note also that polars-high is more or less a highs only framework. For all other solvers it builds the problem and communicates to highs then dumps it into a MPS and then loads this MPS in the other solver.

FabianHofmann:
thanks @MaykThewessen , given the context and the separate benchmark refactor by @FBumann,  I would close this pr, but let's reuse it later if if the topic rises again

PR #742, opened 2026-06-03
refactor: move alignment code to linopy/alignment.py; restructure tests

FBumann:
Follow-up requested by @FabianHofmann in #732:

> move all broadcasting and alignment code from `common.py` to a new module `alignment.py`; move all related tests to `test_alignment.py`, make more use of test classes to structure them better

## Commit 1 — the module move (pure, no behavior change)

`linopy/alignment.py` owns the seam between user input and linopy's labelled arrays:

| Group | Functions |
|---|---|
| coords parsing | `_coords_to_dict`, `_as_index`, `_as_multiindex` |
| conversion | `get_from_iterable`, `pandas_to_dataarray`, `numpy_to_dataarray`, `_named_pandas_to_dataarray`, `fill_missing_coords`, `as_dataarray` |
| MI projection | `_LevelProjection`, `_project_onto_multiindex_levels`, `_warn_implicit_projections` (with `TODO(#738)`) |
| broadcasting | `_broadcast_to_coords`, `broadcast_to_coords` |
| validation | `validate_alignment` |
| symmetric | `align` |

`common.py` keeps the general utilities (formatting, label indexes, polars helpers, decorators). Hard move, no re-exports (nothing outside linopy imports these from `linopy.common` — checked PyPSA). The split is clean: nothing remaining in `common.py` depends on anything moved.

## Commit 2 — the test restructure

`test/test_alignment.py`, one class per concept:

```
TestAsDataarrayFrom{Pandas,Numpy,Scalar,DataArray}, TestAsDataarrayMultiIndexCoords
TestCoordsToDict            # the coords-entry naming rules
TestBroadcastToCoords       # strict=False mechanics
TestMultiIndexProjection    # projection values, deprecation warnings, gaps
TestStrictMode              # strict=True contract
TestValidateAlignment
TestAlign
```

- **Fixtures**: `mi_index` / `mi_coords` / `by_level1` replace the MultiIndex setup that was repeated inline in ~10 tests.
- **Parametrization**: the pandas dims-naming rules (12 near-identical tests → 2 parametrized tables) and the numpy labeling rules (9 → 1 table + 2 edge cases).
- `test_common.py` keeps the utility tests (`iterate_slices`, polars, formatting, …).

## v1 readiness (#717)

[...]

PR #766, opened 2026-06-05
fix: regression with tuple coords entries as xarray's (dim_name, values)

FBumann:
Mandatory to pull this in before next release.
Fixes a bug which was introduced by the alignment refactor. tuple coords were treated differently. Its now consistent with v0.7.0 and xarray.
And additionally improving and reorganizing test coverage

 > [!NOTE]

 <details>
 <summary>Details</summary>

## What

`_coords_to_dict` treated a tuple coords entry as a bare value sequence (like a list), which broke the standard xarray form `coords=[("origin", origins)]`: a scalar bound against such coords raised an inscrutable numpy *"setting an array element with a sequence … inhomogeneous shape"* error.

```python
m.add_variables(lower=0, coords=[("origin", origins), ("dest", dests)])  # crashed
```

v0.7.0 delegated coords to xarray and supported this form; the alignment refactor in #742 (unreleased) regressed it and pinned the value-sequence reading in tests.

## Fix

Restore xarray's convention in `_coords_to_dict`:

- A **tuple** entry is `(dims, data[, attrs])` — its first element names the dimension.
- A **bare value sequence** uses a `list` (or `range` / `ndarray`).
- Length-1 / scalar-value tuples raise a clear `(dim_name, values)` error, mirroring xarray's own rejection.

Unnamed `pd.Index` entries are now named on a copy so an entry's `.name` matches its dim consistently across forms (no caller mutation).

## Tests

- Dropped the rows that pinned the wrong value-sequence reading; added `(name, values)` coverage (parse-level + end-to-end), the multi-tuple scalar-bound regression, and tuple edge cases.
- Restructured `test_alignment.py`'s coords tests into form-parameterized tables and split the end-to-end `add_variables` checks into their own class.

Full test suite passes (3787 passed, 29 skipped).

## Docs

Corrected the unreleased release note to state the xarray `(dim_name, values)` convention (the prior "bare-tuple entries behave like lists" wording described the regressed behaviour).

 </details>


test/test_alignment.py
```diff
@@ -445,119 +441,172 @@ def test_explicit_dims_win_over_inference(self) -> None:
...
+            ([pd.Index([0, 1, 2], name="x")], None, ("x",)),
+            ([pd.Index([0, 1, 2])], ["x"], ("x",)),
+            ([[0, 1, 2]], ["x"], ("x",)),
+            ([range(3)], ["x"], ("x",)),
+            ([np.array([0, 1, 2])], ["x"], ("x",)),
+            ([[0, 1, 2]], None, ("dim_0",)),
+            ([range(3)], None, ("dim_0",)),
+            ([np.array([0, 1, 2])], None, ("dim_0",)),
+            ([pd.Index([0, 1, 2])], None, ("dim_0",)),
+            ([("origin", ["a", "b"]), ("dest", ["x", "y"])], None, ("origin", "dest")),
```
FabianHofmann, line 588: crazy how many combinations this supports :D

FabianHofmann (approved):
thanks for the follow-up and the tidying of the tests

test/test_alignment.py
FBumann: Yes... 
But it always has. Its now just really obvious!

PR #771, opened 2026-06-07
bench: Add internal performance benchmark suite + CodSpeed CI

FBumann:
I worked on this a bit on my fork. The results can be seen on this PR on my fork, where i started to do some performance work: https://github.com/fluxopt/linopy/pull/25

Its meant to be used during development, as well as in ci.

### local dev:

- provides a cli tool to benchmark both walltime and memory usage.
- Can be used to comapre the whole package, or focuse on a single method.
- Includes tools for quick evaluation

### ci (codspeed):

runs a memory benchmark on a suite of models and modeling patterns and compares it to the master baseline. will catch changes/regression of memory usage.
**On master or by adding the "trigger:benchmark" label:** Runs walltime on a dedicated codspeed runner with great consistency

**To activate CodSpeed upstream:** a maintainer needs to connect the repo to CodSpeed (install the CodSpeed app on the PyPSA org/repo) — that's the one required step; the workflows already authenticate via OIDC, no token secret. The `trigger:benchmark` label is created. The walltime job needs CodSpeed macro runners enabled on the org (or drop `codspeed-macro.yml` to keep just the free per-PR memory job).

Notebook walkthrough see [walkthrough.md](https://github.com/fluxopt/linopy/blob/master/benchmarks/walkthrough.md)

**Rough ci times:**
- memory: 3 minutes
- walltime: 10 minutes
- (instrumentation: 20+ minutes - deliberatly not added, see below)

Some setup quirks documented in https://github.com/fluxopt/linopy/issues/27. Probably irrelevant since now

> [!NOTE]
> This description was drafted by AI (Claude); adjust/trim per AGENTS.md. The code is accumulated fork work.

Adds an internal performance-benchmark suite for linopy (build → matrix gen → LP / netCDF export → solver handoff), wired to CodSpeed CI, plus a few correctness fixes surfaced while building it.

### Benchmark suite (`benchmarks/`)
A registry-driven harness measuring **time and peak memory** across linopy's phases.
- **Models** (`REGISTRY`) — whole models swept over size `n`; **patterns** (`PATTERNS`) — realistic modelling idioms swept over a 0–100 `severity` dial (how a data shape going dense propagates through build/export).
- **One CLI** (`python -m benchmarks`): `run` / `sweep` (across linopy versions, via `uv`) with `--metric {time,memory,both}`; `compare` / `plot` auto-detect the metric from the snapshot; plus `smoke`, `list`/`show`/`filter`, `notebook`.
- **Explicit size tiers** per spec (`SIZES` / `QUICK_SIZES` / `LONG_SIZES`): `--quick` per-PR subset, default = `SIZES − LONG_SIZES`, `--long` = all. A single `skip_reason` gates pytest and the memray engine identically.
- **Plotting** — sweep heatmap on a `log₂(fold)` colour scale (symmetric about 1×) with a `--clip` clamp and round fold-change labels; scatter / compare / scaling views; loaders fail cleanly on malformed snapshots.
- **Executable walkthrough** (`walkthrough.md`, run in CI to prevent doc-rot) + README.

### CI
CodSpeed instrumentation: **memory** (heap allocations) on every PR + master push; **walltime** (macro runner) on master push and `trigger:benchmark`-labelled PRs; OIDC auth (no token secret). The simulation/cachegrind instrument is deliberately left out — instruction-count precision is overkill here and burns CI time; memory + walltime cover what matters. We can add it if we dont mid the ci time (only master?)

### Library fixes
- `linopy/alignment.py`: tolerate xarray builds without `CoordinateValidationError`; read tuple coords entries as xarray's `(dim_name, values)` rather than a bare sequence.
- `linopy/expressions.py`: group by non-dimension coordinate names; faster multi-key groupby path.
- Tests for both.

## Checklist

- [x] AI-generated content is marked (this description; see [`AGENTS.md`](../blob/master/AGENTS.md)).
- [x] Code changes are sufficiently documented (docstrings + `benchmarks/walkthrough.md`).
- [x] Unit tests for new features were added (alignment/groupby; benchmark harness `_tests/`).
- [ ] A note for `doc/release_notes.rst` of the upcoming release is included.


.github/workflows/benchmark-smoke.yml
```diff
@@ -0,0 +1,43 @@
+name: Benchmark smoke
+
+# Builds every spec and fires every phase once under --quick
+# --benchmark-disable: a "did a refactor break a spec?" check, not timing.
```
lkstrp, line 4: How long does this run?

.github/workflows/benchmark-smoke.yml
```diff
@@ -0,0 +1,43 @@
...
+
+    steps:
+    - uses: actions/checkout@v6
+      with:
+        fetch-depth: 0  # setuptools_scm
+
+    - name: Set up Python 3.12
+      uses: actions/setup-python@v6
+      with:
+        python-version: "3.12"
```
lkstrp, line 29: Switch to 3.13?

.github/workflows/codspeed-macro.yml
```diff
@@ -0,0 +1,62 @@
...
+# mode that reflects the real cost of dense-vs-sparse work (cache, allocation,
+# native numpy/scipy), which instruction counting under-weights.
+#
+# Master push (updates the walltime baseline) + manual dispatch + opt-in per-PR
+# via the ``trigger:benchmark`` label. Off every *unlabelled* PR: macro-runner
+# minutes are metered (600/month free), and self-hosted bare-metal shouldn't run
+# arbitrary PR code — the label is a maintainer-controlled gate, so only apply it
+# to trusted (same-repo) PRs.
+#
+# Requires the repo under a GitHub org (macro runners are org-only) with the
```
lkstrp, line 13: Please remove the noise

.github/workflows/codspeed-macro.yml
```diff
@@ -0,0 +1,62 @@
...
+# CodSpeed app connected to the repo (OIDC auth — no token secret needed).
+
+on:
```
lkstrp, line 16: Can we merge both codspeed files?

.github/workflows/codspeed-macro.yml
```diff
@@ -0,0 +1,62 @@
...
+  push:
+    branches: [ master ]
+  pull_request:
+    types: [ labeled, synchronize ]
+    branches: [ master ]
+  workflow_dispatch:
```
lkstrp, line 22: I guess push master + workflow_dispatch is enough

.github/workflows/codspeed-macro.yml
```diff
@@ -0,0 +1,62 @@
...
+  cancel-in-progress: true
+
+jobs:
+  macro:
+    name: CodSpeed walltime (macro runner)
+    # Always on master push / dispatch; on PRs only when explicitly labelled.
+    if: >-
+      ${{ github.event_name != 'pull_request' ||
+          contains(github.event.pull_request.labels.*.name, 'trigger:benchmark') }}
+    runs-on: codspeed-macro
```
lkstrp, line 35: What are costs here. Looks great running memory on github runners and clocktime on codspeed runners, can you point me to more resources? Specially if we wanna use this across multiple repos. How long does this run and how much is for free?

pyproject.toml
```diff
@@ -82,11 +82,27 @@ dev = [
...
+    "netcdf4==1.7.4",
+    # last numpy 1.x — keeps sweep coverage back to linopy <0.5.1 (numpy<2)
+    "numpy==1.26.4",
+    "scipy==1.16.3",
+    "xarray==2025.1.2",
+    "pandas==2.3.3",
+    "polars==1.35.2",
+    "dask==2025.11.0",
+    "pytest==9.0.3",
+    "pytest-benchmark==5.2.3",
```
lkstrp, line 99: See comment above

FabianHofmann:
@FBumann codspeed has now all required permissions. I would like to pull this in before the release 0.8 so we have a clear starting point for the benchmark in upcoming versions

.github/dependabot.yml
```diff
@@ -12,3 +12,30 @@ updates:
...
       patterns:
       - '*'
+
+# Pinned ``[benchmarks]`` extra in pyproject.toml. One PR per dep bump
+# → CodSpeed CI runs and attributes any perf delta to that specific
+# bump. Keeps the cross-version ``sweep`` baseline (lockfile-pinned)
+# stable while still surfacing upstream perf changes per-PR with
+# eyes-open review. Loose ``[project.dependencies]`` (numpy, scipy, ...)
+# have no version specifier so Dependabot leaves them alone — only the
+# ``==`` pins in ``[benchmarks]`` produce PRs.
```
lkstrp, line 22: Pinning like you did in `pyproject.toml` doesn't give us full pins, because the deps of deps aren't pinned. Also, I'm not sure we need these per dependency PRs just to get codspeed stats on them. The user will use the newest versions anyway, and it's very unlikely that we'd pin deps for the user just because of codspeed.

So I'd just fully remove this and the benchmark pins in pyproject.toml. For PRs, we probably want to just commit `uv.lock` and use that one in all PRs (for tests and benchmarks), and on master we always resolve from scratch. But all CI runs need to be updated then. I guess this can rather be done in another PR, and here we just ignore dependency pinning

lkstrp (commented):
Hey @FBumann,
codspeed is great stuff! Thanks, I would wanna use this for other repos as well. I only wen now through the workflows, and have a couple of commits. 

Also note while you guys need to decide that. If I skim the rest, I think it's very verbose and pretty sure this is not necessary

Another point:
Can we remove the comments that Codspeed is creating? Similar to Codcov? In my opinion, it's enough to just get the status checks

.github/dependabot.yml
FBumann: I decided against a uv lock file to be able to sweep across linopy versions. As this is arguably not needed for ci and regression tracking, I'm fine with removing this and maybe even commuting a lock file instead. But lockfile usage needs to stay optional to make sweeping possible.
Another reason to add those here was to catch indirect improvements in upstream repos (xarray etc), as a dependabot or would also run when trying to bump the deps.
But we could probably simplify this.

.github/workflows/codspeed-macro.yml
FBumann: Walltime (macro) is free up to 600 min/month

Cost after that is 0,013 €/min

But I'm not sure if we can go over the 600 minutes without a paid plan...

Everything else can run on a girhub runner and is free

https://codspeed.io/pricing

FabianHofmann:
The PR is very large and will be hard to maintain. I know it does not touch the production code but I would like to have things a bit leaner.  

1. Ship Codspeed CI first 
2. Ship local CLI later
3. Trim down the code a bit, remove comments, think about resusing pytest built-ins where possible, DRY  
4. Reduce the number of test cases. I like the coverage of the individual functionality (merge, sparse KVL and so on), I just think it is too much to really track. Let's think about how to make that more useable


> [!NOTE]
> AI-generated (Claude) analysis, updated to fold in the CodSpeed CLI capabilities. Posted as input for the plan above; decisions and replies are the maintainers'.

**On (1) + (2) — what "CI first, CLI later" maps to in the tree.** Both CodSpeed jobs run only `pytest benchmarks/ --quick --codspeed`, so the CI baseline is `registry → phases.py → test_*.py` + `conftest.py` (~1,900 lines). The local-only analytics stack — memray engine, plotting, cross-version sweep, snapshot/compare, ad-hoc `bench` (~55%) — is exactly the "ship later" layer; none of it runs in CI today. Disposition mapped in the table below.

**On (3) — `pytest-codspeed` built-ins replace a chunk of the local stack.** The pinned plugin (5.0.3) has native `--codspeed-mode {walltime,memory}`, so the local dev story can be *"run the same command CI runs, locally"*:
- Local memory profiling → `pytest benchmarks/ -k <name> --codspeed --codspeed-mode memory`. This makes **`memory.py` (511 lines, the memray engine) redundant** — no bespoke memory harness to maintain.
- Ad-hoc local timing → `--codspeed-mode walltime`, covering most of what `bench.py` (346 lines) exists for.
- What the CLI/plugin does **not** provide: a local A/B compare or local plots — those live on the hosted app. So `sweep.py` / `cli/compare.py` / `plotting.py` have no built-in equivalent. But the hosted app already tracks version-over-version on master, so the only genuinely uncovered case is a one-off "linopy X vs Y" study → a `dev-scripts/` script, not merged infra. (Caveat: the docs don't confirm a fully offline / no-token local mode — worth a quick check before relying on it.)
- DRY quick win: `DEFAULT_PHASES` is an identical longhand copy of `ALL_PHASES` (`registry.py`) and collapses to one line.

**On (4) — the registry makes "fewer tracked cases" a data edit, not a rewrite.** Tracked count = specs × phases × sizes. The builders you want to keep (merge, sparse KVL, rolling, …) are one entry each in `patterns/` / `models/` and are cheap to keep. The dials already exist (`quick_sizes` / `long_sizes` per spec, `--quick` / `--long`). So the question is purely *what the dashboard tracks by default*: e.g. default CodSpeed to **one representative size per spec** (keeping every builder runnable on demand) so the dashboard shows ~1 trend per functionality instead of 3–6 — full coverage stays a flag away without flooding the regression view.

<details>
<summary>Component breakdown — CI baseline vs. ship-later, with disposition</summary>

| Component | ~Lines | In CI baseline? | Disposition under the plan |
|---|---|---|---|
| `registry.py` + `models/` + `patterns/` | ~1,400 | ✅ | ship now (core) |
| `phases.py` | 86 | ✅ | ship now |
| `conftest.py` + `test_*.py` | ~380 | ✅ | ship now |
| `cli/run.py` (smoke) + introspect/notebook | ~450 | ✅ (smoke) | ship now, thin |
| `memory.py` (memray engine) | 511 | ❌ | drop — `--codspeed-mode memory` covers it |
| `bench.py` (ad-hoc timer) | 346 | ❌ | mostly dropped — `--codspeed-mode walltime` |
| `plotting.py` | 533 | ❌ | CLI-later; hosted app covers ongoing |
| `sweep.py` (cross-version via uv) | 424 | ❌ | CLI-later / one-off `dev-scripts/` |
| `snapshot.py` + `cli/compare.py` | 333 | ❌ | CLI-later (feeds plotting/compare) |
| `walkthrough.md` + notebook CI job | 361 | smoke only | trim with the layer it documents |

Counts approximate (`wc -l`); CI usage from the three workflow files. CodSpeed CLI facts from <https://codspeed.io/docs/cli> and the `pytest-codspeed` reference.

</details>

.github/workflows/benchmark-smoke.yml
FBumann: 2 minutes
1 for the actual smoke test, 1 for the notebook

Can be optimized I think.

https://github.com/fluxopt/linopy/actions/runs/27085593476/job/79939326221

FBumann:
I think If we go all in on Godspeed we could trim a lot of things from this pr.
But I think the local benchmarking cli is very valuable during the next few months.

If you rather keep it lean, feel free to drop all the local dev stuff. But note that this will mean that we will probably leverage ci/cod speed more during dev.
If you don't mind the ci minutes, this is ok for me. We can add it back in later if we change minds

FabianHofmann:
> I think If we go all in on Godspeed we could trim a lot of things from this pr. But I think the local benchmarking cli is very valuable during the next few months.
> 
> If you rather keep it lean, feel free to drop all the local dev stuff. But note that this will mean that we will probably leverage ci/cod speed more during dev. If you don't mind the ci minutes, this is ok for me. We can add it back in later if we change minds

Let me think about it once more, this is my first intuition. but I let me try the local benchmarking first and then we decide together

FBumann:
I tried the local benchmarking to try out different approaches for memory reduction, and having a local memory benchmark felt really valuable to me (pytest doesn't have one).
As memory is arguably more important than time in some cases, this was the main reason to add it

FBumann:
I think splitting the registry and patterns etc from the utility suite is the right call.
I means that linopy only keeps the thinks it needs for codspeed. 
And another package, which can be reused in other projects, holds the logic for sewwping, plotting etc!

Working on it here: https://github.com/fluxopt/linopy/pull/34

FabianHofmann:
> And another package, which can be reused in other projects, holds the logic for sewwping, plotting etc!

yes, makes sense! have a look at the edited comment above again. quite some things can be pulled from codspeed CLI (but not sweeping and plotting as you said, so perhaps you are aware of this already)

FBumann:
Leveraging cod speed cli would be nice, but it's only supported on Linux. Not everyone will work on Linux (including me), so I would rule this out

FabianHofmann:
@FBumann I really don't want to occupy you in your holidays, and I think you have plans on the implementation. so instead of stripping out the CI relevant part on my own I would defer that to you whenever you have time (when you are back) and go forward with release of `v0.8`.

FBumann:
Just release 0.8.0 in the meantime...
I think regression testing doesn't need a starting point on a certain release...
An we can always sweep backwards to identify historic regressions and fix them later

FBumann:
I would see this as superseeded by #781

PR #774, opened 2026-06-08
fix: Variable.fix() value alignment on named dimensions

FBumann:
This came up during working on another PR.
Small but important bugfix. Very obvious and easy since we introduced clean unternal helpers for such things.

> [!Note]
> AI-assisted implementation.

## Bug

`Variable.fix(value)` aligned the value with `as_dataarray(value).broadcast_like(self.labels)`, which aligns only by **dimension name**. That happens to work only for the default `dim_0`. On a variable with a *named* dimension, a positional value (list/array) gained a spurious `dim_0` and was broadcast **across** the real dimension instead of aligned **onto** it — silently producing a wrong fix constraint.

```python
import linopy, pandas as pd

m = linopy.Model()
t = m.add_variables(lower=-5, upper=5, coords=[pd.Index([2020, 2030, 2040], name="time")], name="t")
t.fix([1.0, 2.0, 3.0])

# before: __fix__t constraint has dims (time: 3, dim_0: 3) — a 3×3 block fixing
#         every t[time] to all of 1, 2, 3 at once (infeasible); only a soft warning
# after:  __fix__t has dims (time: 3) with rhs [1, 2, 3]
```

This affects the released v0.7.0 `fix()`.

## Fix

Use `broadcast_to_coords` against the variable's own coords — the same coords-aware alignment `add_variables` uses for `lower`/`upper`:

- scalars broadcast across the dimension,
- positional inputs (list / ndarray) land on the right dimension,
- named pandas/xarray inputs align by coordinate value (reordered indexes reindex),
- a value carrying an unknown dimension raises an error that names the variable.

## Tests

`TestFixValueAlignment` in `test_fix_relax.py` is parameterized over the value dtype — scalar (broadcast), list, ndarray, named Series (in order and reordered), and a named DataArray — asserting the fix constraint aligns to the named dimension (`dims == ("time",)`) with correct values, plus an unknown-dimension rejection case.

### DIFF
--- a/linopy/variables.py
+++ b/linopy/variables.py
@@ -31,7 +31,7 @@ from xarray.core.types import JoinOptions
 from xarray.core.utils import Frozen
 
 import linopy.expressions as expressions
-from linopy.alignment import as_dataarray, broadcast_to_coords
+from linopy.alignment import broadcast_to_coords
 from linopy.common import (
     LabelPositionIndex,
     LocIndexer,
@@ -1368,7 +1368,9 @@ class Variable:
                 )
                 raise ValueError(msg) from None
 
-        value = as_dataarray(value).broadcast_like(self.labels)
+        value = broadcast_to_coords(
+            value, self.coords, label=f"fix() for variable '{self.name}'"
+        )
 
         if self.attrs.get("integer") or self.attrs.get("binary"):
             value = value.round(0)
--- a/linopy/variables.py
+++ b/linopy/variables.py
@@ -56,10 +56,12 @@ from linopy.common import (
 )
 from linopy.config import options
 from linopy.constants import (
-    FIX_CONSTRAINT_PREFIX,
     HELPER_DIMS,
     SOS_DIM_ATTR,
     SOS_TYPE_ATTR,
+    STASHED_ATTRS,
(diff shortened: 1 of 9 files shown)


linopy/variables.py
```diff
@@ -1368,7 +1368,9 @@ def fix(
                 )
                 raise ValueError(msg) from None
 
-        value = as_dataarray(value).broadcast_like(self.labels)
+        value = broadcast_to_coords(
+            value, self.coords, label=f"fix() for variable '{self.name}'"
```
FabianHofmann, line 1393: I am just wondering about the `label` value (mostly note for myself, have to checkout locally)

FabianHofmann (commented):
great catch, will pull this in quickly, just have to check one thing

linopy/variables.py
FabianHofmann: okay, it is only there for bookkeeping

FabianHofmann (commented):
all resolved - added a test class for covering more cases of alignment. (could argue it duplicates parts of the tests of broadcast_to_coords, but it's good to have them here as well to avoid regression)

FabianHofmann (approved):

PR #783, opened 2026-06-21
fix(constraints): freeze empty constraint groups without reshape error

Ketchp:
> [!NOTE]
> Code and test in this PR was AI aided.

## Problem description:

`CSRConstraint.from_mutable` reshaped `con.vars` with an inferred `-1` dimension. For an empty constraint group (zero rows and zero terms) the vars array has size `0`, and NumPy refuses to infer a `(0, -1)` reshape, raising "cannot reshape array of size 0 into shape (0,newaxis)".

This is a degenerate edge-case, but can be encountered when slicing or masking is used.

Pass the explicit `_term` count instead of `-1`; NumPy accepts a `(0, 0)` reshape and the rest of the method already handles zero-row input.

Minimal reproducible example:
```py
import linopy

m = linopy.Model(freeze_constraints=True) # does not work
# m = linopy.Model(freeze_constraints=False)  # works
x = m.add_variables(coords=[range(3)], dims=["t"], name="x")
m.add_constraints(x.isel(t=range(0)) == 0)  # raises: "ValueError: cannot reshape array of size 0 into shape (0,newaxis)"
```

## Checklist

### DIFF
--- a/linopy/constraints.py
+++ b/linopy/constraints.py
@@ -1081,10 +1081,7 @@ class CSRConstraint(ConstraintBase):
         # Build active_mask aligned with con_labels (rows in csr)
         # Use same filter as to_matrix: label != -1 AND at least one var != -1
         labels_flat = con.labels.values.ravel()
-        # Explicit term count, not -1: NumPy can't infer a (0, -1) reshape for
-        # an empty group (size-0 vars). _term is the trailing dim of con.vars.
-        nterm = con.vars.sizes.get(TERM_DIM, 0)
-        vars_flat = con.vars.values.reshape(len(labels_flat), nterm)
+        vars_flat = con.vars.values.reshape(len(labels_flat), con.nterm)
         active_mask = (labels_flat != -1) & (vars_flat != -1).any(axis=1)
         rhs = con.rhs.values.ravel()[active_mask]
         sign_vals = con.sign.values.ravel()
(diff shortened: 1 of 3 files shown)


linopy/constraints.py
```diff
@@ -1081,7 +1081,10 @@ def from_mutable(
         # Build active_mask aligned with con_labels (rows in csr)
         # Use same filter as to_matrix: label != -1 AND at least one var != -1
         labels_flat = con.labels.values.ravel()
-        vars_flat = con.vars.values.reshape(len(labels_flat), -1)
+        # Explicit term count, not -1: NumPy can't infer a (0, -1) reshape for
+        # an empty group (size-0 vars). _term is the trailing dim of con.vars.
+        nterm = con.vars.sizes.get(TERM_DIM, 0)
+        vars_flat = con.vars.values.reshape(len(labels_flat), nterm)
```
FabianHofmann:
```suggestion
        vars_flat = con.vars.values.reshape(len(labels_flat), con.nterm)
```

test/test_constraint.py
```diff
@@ -95,6 +95,31 @@ def test_add_constraints_uses_model_freeze_default() -> None:
     )
 
 
+def test_csr_constraint_handles_empty_rows() -> None:
```
FabianHofmann: let's also test this for mutable csr constraints: use a pytest parameter for `freeze_constraints` to iterate over true and false, and rename the the test function

FabianHofmann (changes requested):
Thanks @Ketchp  for the quick bug fix. some requests below

linopy/constraints.py
Ketchp: Applied.

test/test_constraint.py
Ketchp: Done.

FabianHofmann (approved):
great, thanks. pypsa-eur test are failing. I need to take a closer look at this first (not related to your changes), but will merge afterwards

PR #793, opened 2026-06-29
perf: scatter groupby-sum terms directly instead of unstacking

FBumann:
__

> [!NOTE]

## What this does

The fast path of `LinearExpression.groupby(...).sum()` previously did
`ds.unstack(group_dim, fill_value=...)` followed by a stack. That
materializes 2–3 intermediate copies of the padded result
(`n_groups × max_group_size × nterm`) and routes through pandas
`MultiIndex` machinery sized by the number of elements.

This change factorizes the groups and scatters `coeffs`/`vars` directly
into preallocated padded result arrays; constants are group-summed with
`np.add.at`. Peak memory drops to input + result (the minimum for the
padded layout), and the grouping itself gets considerably faster. The
result is unchanged: same dims, coords, term ordering and padding.

The unstack-based implementation is kept as `_sum_by_unstack` and is
still used for chunked (dask-backed) data, which cannot be scattered into
numpy arrays. `NaN` group labels now raise an informative `ValueError`
instead of failing inside `unstack`.

## Notes

- Self-contained: touches only `linopy/expressions.py` and adds tests in
  `test/test_linear_expression.py` (124 new lines).
- `_sum_by_unstack` retains the current `master` `names_to_drop` logic
  (drop every coordinate aligned to `group_dim`), so the slow path keeps
  the existing behavior.

<details>

- `pytest test/test_linear_expression.py` → 309 passed
- groupby/sum subset (`-k "group or sum or scatter or unstack"`) → 73 passed
- `ruff check`, `ruff format --check`, `mypy linopy/expressions.py` → clean

</details>

### DIFF
--- a/linopy/expressions.py
+++ b/linopy/expressions.py
@@ -340,20 +340,13 @@ class LinearExpressionGroupby:
 
             # At this point, group is always a pandas Series
             assert isinstance(group, pd.Series)
-            group_dim = group.index.name
 
-            arrays = [group, group.groupby(group).cumcount()]
-            idx = pd.MultiIndex.from_arrays(arrays, names=[GROUP_DIM, GROUPED_TERM_DIM])
-            new_coords = Coordinates.from_pandas_multiindex(idx, group_dim)
-            # collapsing group_dim invalidates every coordinate aligned to it
-            names_to_drop = [
-                name
-                for name, coord in self.data.coords.items()
-                if group_dim in coord.dims
-            ]
-            ds = self.data.drop_vars(names_to_drop).assign_coords(new_coords)
-            ds = ds.unstack(group_dim, fill_value=LinearExpression._fill_value)
-            ds = LinearExpression._sum(ds, dim=GROUPED_TERM_DIM)
+            if self._can_sum_by_scatter(group):
+                ds = self._sum_by_scatter(group)
+            else:
+                # chunked (e.g. dask-backed) data or exotic coordinates on the
+                # grouped dimension: use xarray's unstack machinery
+                ds = self._sum_by_unstack(group)
 
             if int_map is not None:
                 index = ds.indexes[GROUP_DIM].map({v: k for k, v in int_map.items()})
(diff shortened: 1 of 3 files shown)


FabianHofmann (approved):
@FBumann this is super strong. I added release notes and cleaned up some comments. feel free to merge.

linopy/expressions.py
```diff
@@ -374,6 +365,123 @@ def func(ds: Dataset) -> Dataset:
 
         return self.map(func, **kwargs, shortcut=True)
 
+    def _can_sum_by_scatter(self, group: pd.Series) -> bool:
+        """
+        Whether :meth:`_sum_by_scatter` covers the structure of the data.
+
+        The scatter kernel requires numpy-backed arrays (chunked data cannot be
+        scattered into preallocated numpy arrays) and no coordinates tied to
+        the grouped dimension besides its own index. Everything else falls
```
coroa: i don't understand the second condition. What means no coordinates tied to the grouped dimension besides its own index?

coroa (commented):
Cool, need to wrap my head around it

linopy/expressions.py
coroa: ie. can we please improve the docstring

coroa: My thinking is, that if the second condition does basically not happen, then we could wrap `_can_sum_by_scatter` into an `xr.apply_ufunc`, similar to https://github.com/PyPSA/atlite/blob/966380825089e5f913eb3dd45141abbff4fc075d/atlite/gis.py#L1077-L1089 , and get rid of the groupby path entirely.

coroa:
Alternatively, we might think about deprecating the dask layer.

FBumann:
As im working on the v1 convention and the multiindex stuff, this might even become obsolete indirection. I need to think about it

coroa:
Free to merge from my side, too

FBumann:
> As im working on the v1 convention and the multiindex stuff, this might even become obsolete indirection. I need to think about it

I'm sure this is not correct. I'm also pro merging this

FBumann:
But I would prefer #802 over this for code simplicity (only one path remains)

FBumann:
Superseded by #802

PR #794, opened 2026-06-29
docs: import AGENTS.md into CLAUDE.md

FBumann:
Inject `AGENTS.md` into `CLAUDE.md` via the `@AGENTS.md` import syntax, so Claude Code loads the contribution rules, workflow and conventions automatically instead of depending on the agent to follow the markdown link.

`AGENTS.md` remains the single source of truth — no duplicated prose to keep in sync.

> [!NOTE]
> This change was generated by AI.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,5 +1,8 @@
 # CLAUDE.md
 
 This project's guidance for AI agents — contribution rules, development workflow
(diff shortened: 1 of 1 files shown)


FabianHofmann (approved):
thanks! I've been noticing it and wanted to tackle, but here we already go :)

PR #797, opened 2026-06-29
feat(piecewise): partial `active` gate support via active_gate helper (#796)

### LINKED ISSUE #796: add_piecewise_formulation: support a partial `active` gate (treat missing/null entries as always-active)
We need partial active triggers in pw formulations. 

> [!Note]
> AI written issue

`Model.add_piecewise_formulation(..., active=<var>)` gates a piecewise function on a binary `active` variable: when `active=0`, the auxiliary variables (and thus the output) are pushed to zero. The gating enters through `delta_var <= active` and the `bp0 * active` intercept term in `_add_incremental` (and the analogous disjunctive path).

This works when `active` is defined over the *entire* coordinate of the formulation. It breaks down when `active` is only defined over a **subset** of the indexed dimension.

### Problem

When the `active` expression is null/missing for part of the formulation's dimension, those entries get gated as if `active=0` and their output is forced to zero — rather than being treated as ungated (always on).

This surfaced downstream in PyPSA ([PyPSA/PyPSA#1755](https://github.com/PyPSA/PyPSA/pull/1755)). PyPSA builds one piecewise formulation across a mix of **committable** and **non-committable** components and passes the unit-commitment `status` variable as `active`. The `status` variable only exists for committable components; for the rest it is null. The result: non-committable components — which should always dispatch — were incorrectly forced to zero by the `_active` piecewise constraint.

### Proposed feature

Let `add_piecewise_formulation` accept an `active` expression/variable that is **defined only over a subset** of the formulation's dimension (i.e. null/missing elsewhere). Entries where `active` is missing should be treated as **always active** (gate = 1, ungated), and entries where it is present should be gated as today.

Effectively:
- `where active.isnull()` → no gating
- `where ~active.isnull()` → gated on `active`

### Notes / open questions

- The `active` semantics are already flagged as an evolving API (`EvolvingAPIWarning`), so refining them here fits.
- Decide whether the trigger is "NaN entries in a full-coordinate `active`" vs. "`active` indexed over a strict subset of the dimension" — both should resolve to the same behaviour, ideally without the caller pre-aligning.
- Applies to both the incremental/SOS2 (`_add_incremental`) and disjunctive (`_add_disjunctive`) paths; the LP path already rejects `active`.

FBumann:
_(placeholder — Felix to rewrite)_

Closes #796. See the [design discussion in the issue](https://github.com/PyPSA/linopy/issues/796#issuecomment-4831729593).

> [!NOTE]

## Problem

`add_piecewise_formulation(active=...)` assumed the gate spanned the formulation's full indexed dimension. A gate defined over only a **subset** of coordinate labels — or with **masked** entries — silently forced the uncovered entries to zero, instead of leaving them ungated. This broke formulations that mix gated and ungated entities (e.g. committable + non-committable units sharing a unit-commitment `status`), surfaced in PyPSA/PyPSA#1755.

## Approach

Per the issue discussion, this does **not** make the function silently accept misaligned coords (that would mask genuine coverage bugs and clashes with v1 alignment semantics). Instead:

1. **Validate, don't auto-fix.** `add_piecewise_formulation` now rejects an under-defined `active` (strict subset or masked) with an actionable error. A lower-*dimensional* gate still broadcasts and is accepted.
2. **`linopy.active_gate(active, coords, fill_value=1)`** pads a partial gate to full coverage — missing/masked entries become always-active (`1`) or always-off (`0`). The signature mirrors `active.reindex(coords).fillna(fill_value)` so migration is mechanical.

`active_gate` is a **temporary legacy stopgap**, isolated in `linopy/_active_gate.py`: under the planned v1 arithmetic semantics (#717) the bare `reindex().fillna()` idiom is correct on its own, so the helper is expected to be deprecated and that file removed. The validation guard stays.

<details>
<summary>Why a helper is needed today (legacy footgun)</summary>

The "always-on" pad is conceptually just `reindex(coords).fillna(1)`. Today that idiom is shape- and cast-dependent: a bare `Variable.fillna` resolves a gap to always-*off*, `fillna` can't add absent labels (a strict subset needs `reindex`), and a masked entry has no NaN for `fillna` to fill (needs `.where`). There is no single correct one-liner across both shapes. `active_gate` encapsulates a `.where`-on-term-labels recipe that is correct for both shapes; under v1 it collapses to the bare idiom. Full per-shape/per-regime table in the [issue comment](https://github.com/PyPSA/linopy/issues/796#issuecomment-4831729593).

</details>

## Changes

- `linopy/_active_gate.py` (new) — `active_gate` helper.
- `linopy/piecewise.py` — `_validate_active_coverage` + wiring; `active` docstring.
- `linopy/__init__.py` — export `active_gate`.
- `test/test_piecewise_active_gate.py` (new) — parametrized over both partial shapes × helper/validation/solver (incremental, sos2, disjunctive), plus broadcast and full/scalar acceptance.
- Docs: `api.rst`, piecewise guide (partial-gates section), release notes.

## Tests

`test_piecewise_active_gate.py` + `test_piecewise_constraints.py`: 247 passed. mypy and ruff clean.

### DIFF
--- a/linopy/_active_gate.py
+++ b/linopy/_active_gate.py
@@ -1,0 +1,75 @@
+"""
+Legacy helper for padding a partial piecewise ``active`` gate.
+
+This module is a temporary stopgap.  Under the planned v1 arithmetic
+semantics (#717) the bare idiom ``active.reindex(coords).fillna(fill_value)``
+is correct on its own, so :func:`active_gate` is expected to be deprecated
+and this file removed once v1 lands.  Keeping it isolated makes that
+removal a single-file delete.
+"""
+
+from __future__ import annotations
+
+from collections.abc import Mapping
+from typing import TYPE_CHECKING, Any
+
+from linopy.piecewise import _to_linexpr, _warn_evolving_api
+
+if TYPE_CHECKING:
+    from linopy.expressions import LinearExpression
+    from linopy.types import LinExprLike
+
+
+def active_gate(
+    active: LinExprLike,
+    coords: Mapping[Any, Any],
+    fill_value: float = 1,
+) -> LinearExpression:
+    r"""
+    Pad a partial ``active`` gate to full coverage for piecewise gating.
+
+    Reindexes ``active`` to ``coords`` and fills missing/masked entries with
+    ``fill_value`` (``1`` = always active, ``0`` = always off), so a gate
+    defined over only a subset of :meth:`~linopy.Model.add_piecewise_formulation`'s
+    coordinate does not force the uncovered entries to zero. Equivalent to the
+    v1 idiom ``active.reindex(coords).fillna(fill_value)`` but correct under
+    legacy too (see the module docstring).
+
+    .. code-block:: python
+
+        gate = active_gate(status, {"component": components})
+        m.add_piecewise_formulation((power, xs), (fuel, ys), active=gate)
+
+    Parameters
+    ----------
+    active : Variable or LinearExpression
+        The (possibly partial) gate expression.
+    coords : mapping of dim to labels
+        Reindex target, passed straight to ``reindex``; unlisted dims
+        broadcast.
+    fill_value : float, default 1
+        Value for missing/masked entries (``1`` = on, ``0`` = off).
+
+    Returns
+    -------
+    LinearExpression
+        The padded gate, suitable to pass as ``active=``.
+
+    Warns
+    -----
+    EvolvingAPIWarning
+        Part of the evolving piecewise API; may be refined.
+    """
+    _warn_evolving_api(
+        "active_gate",
+        "piecewise: active_gate is a new API; its signature and the way it "
+        "resolves missing/masked entries may be refined in minor releases.  "
+        "It is primarily a legacy stopgap and may be removed once legacy "
+        "semantics are dropped.  This warning fires once per session; "
+        "silence with "
+        '`warnings.filterwarnings("ignore", category=linopy.EvolvingAPIWarning)`.',
+    )
+    gate = _to_linexpr(active).reindex(coords)
+    term_dims = [d for d in gate.vars.dims if d not in gate.coord_dims]
+    present = (gate.vars >= 0).any(term_dims)
+    return gate.where(present, fill_value)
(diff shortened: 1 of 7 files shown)


linopy/_active_gate.py
```diff
@@ -0,0 +1,75 @@
...
+        "resolves missing/masked entries may be refined in minor releases.  "
+        "It is primarily a legacy stopgap and may be removed once legacy "
+        "semantics are dropped.  This warning fires once per session; "
+        "silence with "
+        '`warnings.filterwarnings("ignore", category=linopy.EvolvingAPIWarning)`.',
+    )
+    gate = _to_linexpr(active).reindex(coords)
+    term_dims = [d for d in gate.vars.dims if d not in gate.coord_dims]
+    present = (gate.vars >= 0).any(term_dims)
+    return gate.where(present, fill_value)
```
FabianHofmann:
```suggestion
    gate = _to_linexpr(active).reindex(coords)
    return gate.where(gate.has_terms, fill_value)
```

FabianHofmann (changes requested):
thanks @FBumann ; the exposure of the new `active_gate` function is not ideal as this will be a throw-away in v1. but I definitely see the point. a potentiially cleaner solution would be adding a explicit boolean `mask` that has prio and active can be a subset. then we have a clear split on masking and gating and align the signature with the other `add_*` functions. what do you think?

linopy/_active_gate.py
FabianHofmann: (note this is independent of my proposal)

FBumann:
> thanks @FBumann ; the exposure of the new `active_gate` function is not ideal as this will be a throw-away in v1. but I definitely see the point. a potentiially cleaner solution would be adding a explicit boolean `mask` that has prio and active can be a subset. then we have a clear split on masking and gating and align the signature with the other `add_*` functions. what do you think?

A `mask` parameter is indeed a good idea. But it should mask the whole constraint, not just active! So its not a fix to this use case.

And yes, the helper isnt ideal, but as you can see from the tests and the AI comment, under legacy, there are many edge cases, and scoping them in a helper keeps `add_piecewise_formulation` clean. Else we need parameters purely scoped to active (which we could do...), and deprecate them gracefully. Maybe this is the right call from a user perspective though...

linopy/_active_gate.py
FBumann: Perfect. THats why we added `has_terms'!!

FBumann:
I'll do another branch with inline parameters, so we can decide from seeing the api!

FabianHofmann:
closed via #798

PR #801, opened 2026-06-30
test: make assert_linequal ignore dimension order

FBumann:
I hit some false positives while working on sth else. This fixes them. Making the testing more robust.

> [!NOTE]

## What this does

linopy expressions are dimension-labelled xarray structures, so they inherit
xarray's broadcast order-sensitivity: `x + y` produces dims `(x_dim, y_dim)`
while `y + x` produces `(y_dim, x_dim)`. That difference is not mathematically
meaningful, but xarray's `assert_equal` is order-sensitive, so
`assert_linequal(x + y, y + x)` failed.

`assert_linequal` already normalizes *term* order before comparing (its
docstring promises "semantically equal"). This aligns *dimension* order the
same way: a new `_align_dim_order` helper transposes the second operand to the
first's dimension order when they share a dimension set. If the dimension sets
differ, the expressions are genuinely unequal and the comparison is left
untouched so `assert_equal` reports the real difference.

## Notes

- Touches only `linopy/testing.py`; adds `test/test_testing.py` (dim-order
  insensitivity, plus genuinely-different expressions still fail).
- Full `test_linear_expression.py` + `test_quadratic_expression.py` pass
  unchanged (333), i.e. nothing was relying on the order-sensitive behavior.

<details>
<summary>Reproducer</summary>

```python
import pandas as pd
from linopy import Model

m = Model()
a = m.add_variables(coords=[pd.Index([0, 1], name="i")], name="a")
b = m.add_variables(coords=[pd.Index([0, 1, 2], name="j")], name="b")

(a + b).data.coeffs.dims   # ('i', 'j', '_term')
(b + a).data.coeffs.dims   # ('j', 'i', '_term')
assert_linequal(a + b, b + a)   # before: AssertionError; after: passes
```

</details>


FBumann:
canceled pypsa-tests

FabianHofmann (approved):

PR #802, opened 2026-06-30
perf(groupby): unify scatter kernel over numpy and dask via apply_ufunc

FBumann:
I use apply_ufunc to make this dask capable. As we dont have the reference unstack implementation anymore, i introduced quite a heavy testing part (fast though), as I find the apply_ufunc version harder to understand personally. Happy to strip it down.

> [!NOTE]

**Stacked on #793.** Until #793 merges, this PR's diff includes its commit too — review only the top commit `perf(groupby): unify scatter kernel ...`.

## What this does

#793 split groupby-sum into a numpy kernel and a dask `unstack` fallback. This
collapses them into a **single** kernel (`_grouped_sum`) wrapped in
`xarray.apply_ufunc`:

- numpy-backed data scatters terms into the padded result arrays as before;
- chunked (dask) data runs the *same* scatter lazily via `dask="parallelized"`,
  after gathering the grouped dimension into a single chunk (which unstacking
  required too).

This removes the last `pd.MultiIndex`/`unstack` usage in groupby-sum, drops the
numpy-vs-dask branch in `sum()`, and keeps peak memory at input + result on both
backends. Multi-key / DataFrame grouping and its `MultiIndex` result are
unaffected — that logic sits above the kernel (existing tests cover it).

## Tests

The kernel is verified **from first principles** — for every group and every
slice over the non-grouped dims, the result's live terms must equal the multiset
of its members' terms and the constant their NaN-skipping sum — across every
case shape on **both numpy and dask** backends. Three explicit anchors pin the
exact padded layout (member order, fill position, `(nterm, max_size)`
interleaving, and the `_factor` axis) for the linear, multidim and quadratic
cases.

<details>
<summary>Benchmark (300k elem × 8 dim × 1000 groups, numpy)</summary>

| kernel | mean time | peak memory |
| --- | --- | --- |
| `_sum_by_scatter` (#793) | 46.5 ms | 114.6 MiB |
| `_sum_by_unstack` (#793 dask path) | 49.9 ms | 254.8 MiB |
| **apply_ufunc (this PR)** | 48.3 ms | **114.6 MiB** |

The unified kernel matches the scatter kernel's memory and time; the old dask
path cost 2.2× peak.

</details>

## Notes

- Touches only `linopy/expressions.py` and the groupby kernel tests; full
  `test_linear_expression` + `test_quadratic_expression` pass (366), broader
  suite green.
- Gathering `group_dim` into one chunk is unavoidable for a scatter (a group's
  members can sit in any chunk) and is exactly what the old unstack path forced.

### DIFF
--- a/linopy/expressions.py
+++ b/linopy/expressions.py
@@ -388,9 +388,9 @@ class LinearExpressionGroupby:
         group_dim = group.index.name
         fill_value = LinearExpression._fill_value
 
-        scatter_core_dims = {group_dim: -1, TERM_DIM: -1}
+        _scatter_core_dims = {group_dim: -1, TERM_DIM: -1}
         if data.chunks:
-            data = data.chunk(scatter_core_dims)
+            data = data.chunk(_scatter_core_dims)
 
         codes, unique_groups = pd.factorize(group, sort=True)
         if (codes == -1).any():
--- a/linopy/expressions.py
+++ b/linopy/expressions.py
@@ -74,7 +74,6 @@ from linopy.constants import (
     FACTOR_DIM,
     GREATER_EQUAL,
     GROUP_DIM,
-    GROUPED_TERM_DIM,
     HELPER_DIMS,
     LESS_EQUAL,
     STACKED_TERM_DIM,
@@ -341,18 +340,7 @@ class LinearExpressionGroupby:
             # At this point, group is always a pandas Series
             assert isinstance(group, pd.Series)
 
-            numpy_backed = all(
-                isinstance(self.data[k].data, np.ndarray)
-                for k in ("coeffs", "vars", "const")
-            )
-            if numpy_backed:
-                ds = self._sum_by_scatter(group)
-            else:
-                logger.debug(
-                    "groupby-sum: non-numpy-backed (e.g. dask) data, "
-                    "falling back to the unstack kernel."
-                )
-                ds = self._sum_by_unstack(group)
+            ds = self._grouped_sum(group)
 
             if int_map is not None:
                 index = ds.indexes[GROUP_DIM].map({v: k for k, v in int_map.items()})
@@ -373,20 +361,18 @@ class LinearExpressionGroupby:
 
         return self.map(func, **kwargs, shortcut=True)
 
-    def _sum_by_scatter(self, group: pd.Series) -> Dataset:
+    def _grouped_sum(self, group: pd.Series) -> Dataset:
         """
         Sum groups by scattering all terms directly into the final padded arrays.
 
         Every group member keeps its block of ``nterm`` terms, so the resulting
         term dimension has size ``max_group_size * nterm`` and smaller groups are
-        padded with fill values. In contrast to :meth:`_sum_by_unstack` only the
-        result arrays are allocated, without intermediate copies of that size.
+        padded with fill values. Only the result arrays are allocated, keeping
+        peak memory at input + result.
 
-        Only the term and constant values are computed with numpy; the result
-        structure (dimensions, coordinates and their order) is assembled by
-        xarray itself and thereby matches the result of unstacking the group
-        dimension. The caller dispatches here only for numpy-backed data
-        (chunked data uses :meth:`_sum_by_unstack`).
+        The scatter runs inside :func:`xarray.apply_ufunc`, so it covers numpy
+        and chunked (dask) data alike: for dask the grouped dimension is gathered
+        into a single chunk and the scatter is applied lazily.
         """
         data = self.data
         group_dim = group.index.name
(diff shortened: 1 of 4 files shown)


FBumann:
I ran the benchmark locally with pytets-benchmem to also check timing. Small improvement there too!

> [!NOTE]

Local verification of the performance claim on the `nodal_balance` reference model from `benchmarks/` (the `groupby(bus).sum()` idiom), comparing `master` (`fe798b1`, pre-#793) against this branch. Measured with `pytest-benchmark` + `pytest-benchmem` (`--benchmark-memory`).

| `test_build[nodal_balance]` | time: master → branch | peak mem: master → branch |
| --- | --- | --- |
| severity=0 | 10.24 → 9.78 ms (−5%) | 1.43 → 1.30 MiB |
| severity=50 | 11.16 → 10.14 ms (−9%) | 16.60 → 6.92 MiB (×2.4) |
| severity=100 | 11.70 → 10.39 ms (−13%) | 31.82 → 12.72 MiB (×2.5) |

Both peak memory (~2.4–2.5× lower) and build time (~5–13% faster) improve, and the gain grows with group skew — the pathological case the scatter targets. Consistent with the CodSpeed report (×2–3 memory on build/to_lp/to_solver).

<details>
<summary>Method</summary>

Each version's `linopy` selected via `PYTHONPATH` (master checked out in a throwaway worktree; branch from this checkout), same `benchmarks/` suite, then:

```
benchmem compare bench_master.json bench_branch.json --columns time,peak --stat mean
```

A `benchmem sweep linopy <master-ref> <branch-ref> --suite benchmarks/ --memory` reproduces this in isolated uv venvs.

</details>

FabianHofmann:
`apply_ufunc` is of course master class :) this implementation is quite complex and I would run it on pypsa tests first. I would say this should have priority over #793 and should directly go into master. @coroa is it realistic for you to take a look?

coroa:
I'll put in on my stack, but no promises this week. I also think this is not time critical. The actual use of dask, is non-existent in my expectation.

linopy/expressions.py
```diff
@@ -74,7 +74,6 @@
     FACTOR_DIM,
     GREATER_EQUAL,
     GROUP_DIM,
-    GROUPED_TERM_DIM,
```
coroa, line 77: I think GROUPED_TERM_DIM is then unused and could be removed

linopy/expressions.py
```diff
@@ -374,6 +361,91 @@ def func(ds: Dataset) -> Dataset:
...
+            moved = np.moveaxis(values, -1, 0)
+            out = np.zeros((n_groups, *moved.shape[1:]), dtype=values.dtype)
+            np.add.at(out, codes, np.where(np.isnan(moved), 0, moved))
+            return np.moveaxis(out, 0, -1)
+
+        def single_chunk(da: DataArray) -> DataArray:
+            # the scatter's core dims must each sit in one chunk
+            if da.chunks is None:
+                return da
+            return da.chunk({d: -1 for d in (group_dim, TERM_DIM) if d in da.dims})
```
coroa: `Dataset.chunk` forwards to all `DataArray.chunk` and that ignores any unknown dims, so it's easier to add before the xr.apply_ufunc calls:

```python
if data.chunks is not None:
    data = data.chunk({group_dim: -1, TERM_DIM: -1})
```

linopy/expressions.py
```diff
@@ -374,6 +361,91 @@ def func(ds: Dataset) -> Dataset:
...
+        positions = pd.Series(codes).groupby(codes).cumcount().to_numpy()
+        nterm = data.sizes[TERM_DIM]
+
+        def scatter_terms(values: np.ndarray, fill: Any) -> np.ndarray:
+            # (..., n_elem, nterm) -> (..., n_groups, nterm * max_size); each
+            # member's nterm block is kept together, padding at the block's end
+            rest = values.shape[:-2]
+            out = np.full((*rest, n_groups, nterm, max_size), fill, dtype=values.dtype)
+            out[..., codes, :, positions] = np.moveaxis(values, -2, 0)
+            return out.reshape((*rest, n_groups, nterm * max_size))
```
coroa: Took me a while to work that out, but it is sound. Very nice

linopy/expressions.py
```diff
@@ -374,6 +361,91 @@ def func(ds: Dataset) -> Dataset:
...
+
+        def group_sum(values: np.ndarray) -> np.ndarray:
+            # (..., n_elem) -> (..., n_groups), summing within groups, skipping NaN
+            moved = np.moveaxis(values, -1, 0)
+            out = np.zeros((n_groups, *moved.shape[1:]), dtype=values.dtype)
+            np.add.at(out, codes, np.where(np.isnan(moved), 0, moved))
+            return np.moveaxis(out, 0, -1)
```
coroa: Also ideal

linopy/expressions.py
```diff
@@ -374,6 +361,91 @@ def func(ds: Dataset) -> Dataset:
...
+        def single_chunk(da: DataArray) -> DataArray:
+            # the scatter's core dims must each sit in one chunk
+            if da.chunks is None:
+                return da
+            return da.chunk({d: -1 for d in (group_dim, TERM_DIM) if d in da.dims})
+
+        def scatter(da: DataArray, fill: Any) -> DataArray:
+            return xr.apply_ufunc(
+                scatter_terms,
+                single_chunk(da),
```
coroa: chunk dataset instead

linopy/expressions.py
```diff
@@ -374,6 +361,91 @@ def func(ds: Dataset) -> Dataset:
...
+                dask="parallelized",
+                dask_gufunc_kwargs={
+                    "output_sizes": {GROUP_DIM: n_groups, TERM_DIM: nterm * max_size}
+                },
+                output_dtypes=[da.dtype],
+            )
+
+        const = xr.apply_ufunc(
+            group_sum,
+            single_chunk(data.const),
```
coroa: same

linopy/expressions.py
coroa: Must have been part of a previous PR, but i find this doc string problematic. It does not tell you what keyword arguments will be ignored, or could be given when you use_fallback. Should be improved.

coroa: `sort of default xarray implementation` is also not helpful at all

linopy/expressions.py
```diff
@@ -374,6 +361,91 @@ def func(ds: Dataset) -> Dataset:
...
+        Sum groups by scattering all terms directly into the final padded arrays.
+
+        Every group member keeps its block of ``nterm`` terms, so the resulting
+        term dimension has size ``max_group_size * nterm`` and smaller groups are
+        padded with fill values. Only the result arrays are allocated, keeping
+        peak memory at input + result.
+
+        The scatter runs inside :func:`xarray.apply_ufunc`, so it covers numpy
+        and chunked (dask) data alike: for dask the grouped dimension is gathered
+        into a single chunk and the scatter is applied lazily.
```
coroa: implementation detail. remove

test/test_linear_expression.py
```diff
@@ -1908,6 +1911,332 @@ def test_linear_expression_groupby_from_variable(v: Variable) -> None:
...
+        assert_linequal(grouped, fallback)
+
+        # every group carries exactly the variables of its members, rest is fill
+        for letter in ["a", "b", "c"]:
+            members = np.where(np.array(labels) == letter)[0]
+            vars_of_group = grouped.data.vars.sel(letter=letter).values
+            present = set(vars_of_group[vars_of_group >= 0])
+            assert present == set(v.labels.values[members])
+            assert (vars_of_group >= 0).sum() == len(members) * expr.nterm
+            assert grouped.const.sel(letter=letter).item() == 5 * len(members)
```
coroa, line 2016: I find this the most relevant test.

test/test_linear_expression.py
```diff
@@ -1908,6 +1911,332 @@ def test_linear_expression_groupby_from_variable(v: Variable) -> None:
...
+    ) -> None:
+        """
+        Each group's terms and constant must match its members, from first
+        principles, on both numpy and dask backends. See ``GROUPBY_SUM_CASES``
+        for the structures covered.
+        """
+        if backend == "dask":
+            pytest.importorskip("dask")
+        expr, groups = build(groupby_ctx)
+        _assert_grouped_sum_correct(expr, groups, chunked=backend == "dask")
```
coroa: Wouldn't the simpler pattern to compare fallback to scatter implementation, like the skewed one above, be the better comparison. Didn't check in detail what assert_grouped_sum_correct actually does, but it looks too complicated.

coroa (changes requested):
Cool, the core stuff is very good. I find the tests a bit bloated and have a few small comments.

Great work

test/test_linear_expression.py
FBumann: The fallback errors on quadratics (no ``_factor`` handling) and diverges on masked and auxiliary-coordinate cases, so it cannot validate the scatter kernel in general.

But we should split the cases up i think!

FBumann:
@coroa Im done here. About the bloated tests. I agree, but thought its better to over cover for now. If you agree, im open to removing some tests.

1. test_grouped_sum_matches_fallback's dask backend
2. test_exact_padded_layout
3. test_chunked_runs_lazily
4. test_skewed_unsorted_groups

FabianHofmann (approved):
I would keep the tests as they are. they cover some edge cases, but with the upcoming changes in v1 I prefer a high test coverage.  feel free to pull in I'd say

FBumann:
@coroa I would merge this tomorrow morning if you have no objections to my latest changes.

linopy/expressions.py
```diff
@@ -242,7 +268,6 @@ def groupby(self) -> xarray.core.groupby.DatasetGroupBy:
     def map(
         self,
         func: Callable[..., Dataset],
-        shortcut: bool = False,
```
coroa, line 245: This might be a potentially breaking API change? Even though it is dead API.

linopy/expressions.py
```diff
@@ -359,7 +364,7 @@ def func(ds: Dataset) -> Dataset:
             ds = ds.assign_coords({TERM_DIM: np.arange(len(ds._term))})
             return ds
 
-        return self.map(func, **kwargs, shortcut=True)
+        return self.map(func)
```
coroa, line 376: @FabianHofmann Can you clarify what kwargs were meant to be passed here? Shortcuts was dead code, so fine to remove.

linopy/expressions.py
```diff
@@ -369,15 +374,15 @@ def _grouped_sum(self, group: pd.Series) -> Dataset:
...
-        and chunked (dask) data alike: for dask the grouped dimension is gathered
-        into a single chunk and the scatter is applied lazily.
         """
         data = self.data
         group_dim = group.index.name
         fill_value = LinearExpression._fill_value
 
+        scatter_core_dims = {group_dim: -1, TERM_DIM: -1}
+        if data.chunks:
+            data = data.chunk(scatter_core_dims)
```
coroa: Isn't it simpler to inline the dict?

test/test_linear_expression.py
coroa: Wouldn't it be easier then to do a simpler test example for quadratics, and fix the divergences by re-ordering or whatever it would take?

I don't understand the test and where it would fail and find this concerning. I agree with large test coverage is good. With claude i often can't tell whether the tests actually test what we would want them to

linopy/expressions.py
FBumann: @coroa We could also keept it with a Future Warning instead. I did not realize `def map()` is actually used. But this is arguably a wrong assumption.
I will change it just to make sure and mark it for removal in v1

linopy/expressions.py
FBumann: Yes. But naming the dict makes the code more readable i think. Ill prefix it with an underscore to indicate it as very local

linopy/expressions.py
FabianHofmann: yes, it was a no-op that was mainly there for mirroring the signature of the internal xarray groupby.map function. it was never explicitly used as I (guess I) wanted to tackle this later which never happened. so fine to remove. the kwargs seems to be dead anyway and setting it would have let to errors. so this is fixing it

FBumann:
@coroa I changed the testing from the generic but hard to read oracle to a few hand rolled cases. I dropped the oracle. We could also keep the oracle if you value the test surface, but I think its fine like this.

FabianHofmann:
@FBumann I had a another view now. I think it's fine. let's pull this in

coroa:
Sorry for not keeping up with review speed.

Yes, thanks, the hand-rolled cases are clearer to me.

Both: But, I would appreciate if you waited for me once i am reviewing things until i give the ok.

Issue #40, opened 2022-09-02
Linopy v0.0.10 breaks with xarray v2022.06


FabianHofmann:
Due to some newly introduced bugs in xarray, the most recent linopy version `v0.0.10` breaks with xarray `v2022.06`. We recommend to downgrade xarray to `v2022.03` if users encounter problems.

PR #805, opened 2026-07-01
bench: arithmetic operation micro-benchmarks

FBumann:
I really want to make sure the v1 convention doesnt introduce regressions to linopy. And i think as soon as we decide on what to do with the multiindex stuff, there will be room for improval. THis should lay a nie foundation for that.

Codspeed cost is really small, but coverage (and granularity) really improves.

> [!NOTE]
> AI-assisted (Claude Code): implementation and this description; reviewed by me.

Adds an **op-level benchmark tier** to `benchmarks/`, alongside the whole-model build benchmarks. One benchmark per **operation**, with operands built **outside** the measured region so a run isolates a single arithmetic op.

**Why.** Whole-build benchmarks catch a regression but can't attribute it — a build says *"kvl got heavier"*, an op benchmark says *"`expr+expr` broadcast got heavier"*. (Motivated by a real regression hunt where attribution needed exactly this granularity.)

**What.**
- `benchmarks/ops.py` — op registry (`OpSpec`) + a single 3-D `grid` size profile (dims 3×4×1000; the asymmetric shape also catches dim-order/transpose bugs) + **35 ops** across scaling, var/expr arithmetic, quadratic, reductions, masking, groupby, merge, and constraint construction. Binary labelled ops carry **`match`/`broadcast`** variants — the alignment-path axis where the interesting regressions live.
- `benchmarks/drivers/test_ops.py` — parametrized driver, one benchmark per op.
- `conftest.py` — `test_ops` added to `CODSPEED_MODULES` (tracked; memory advisory).

**Cost.** 35 benchmarks; the memory run stays ~2–2.5 min including the model builds that dominate the job — cheap.

**Signal validated.** `broadcast ≈ 5× match` on the alignment axis (the §9 cross-product) — well above the noise floor.

Memory is **report-only** to start (op-scale memory can be noisy); op-*time* is the natural gate candidate once the signal proves stable.

### DIFF
--- a/benchmarks/conftest.py
+++ b/benchmarks/conftest.py
@@ -20,6 +20,7 @@ CODSPEED_MODULES = (
     "test_build",
     "test_to_lp",
     "test_to_solver",
+    "test_ops",
 )
 
 
--- a/benchmarks/drivers/test_ops.py
+++ b/benchmarks/drivers/test_ops.py
@@ -12,16 +12,12 @@ from collections.abc import Callable
 
 import pytest
 
-from benchmarks.ops import OpSpec, Profile, iter_op_params
+from benchmarks.ops import GRID, OpSpec, iter_ops
 
-_CASES = iter_op_params()
+_OPS = iter_ops()
 
 
-@pytest.mark.parametrize(
-    "op, profile",
-    _CASES,
-    ids=[f"{op.name}[{profile.key}]" for op, profile in _CASES],
-)
-def test_op(benchmark: Callable[..., object], op: OpSpec, profile: Profile) -> None:
-    operands = op.setup(profile)
+@pytest.mark.parametrize("op", _OPS, ids=[op.name for op in _OPS])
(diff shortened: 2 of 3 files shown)


FBumann:
> [!NOTE]

**CodSpeed cost check.** This PR adds 35 arithmetic micro-benchmarks (one `GRID`
profile each, dims 3×4×1000). Under CodSpeed's per-PR **memory** instrument the
full `benchmarks/` run (build + ops) stays at ~2.5 min — no cost blow-up:

| Run | Instrument | Duration |
|-----|-----------|----------|
| fork PR (ops only) | memory | 2m36s |
| fork PR #40 (build + ops) | memory | 2m44s |

The bare-metal **walltime** job remains gated to `master` + the `trigger:benchmark`
label, so PRs don't incur it. Verified on the fork: fluxopt/linopy#40 ran the ops
under CodSpeed and the memory comparison worked without inflating cost.

FabianHofmann (approved):

FabianHofmann:
nice idea! and great work on v1

PR #806, opened 2026-07-01
fix: support adding/removing variables after adding frozen constraint

koen-vg:
I like the new `CSRConstraint` and the memory benefits it comes with. The current implementation, however, breaks when you add or remove variables after adding a frozen constraint. This is because existing CSRConstraints record the number of active variables (matrix columns) in a `shape` parameter, which becomes stale when a new variable is added. When a variable (column) is removed, the CSR indices change and have to be rebuilt.

This PR makes it so CSR representations are updated when variables are added and removed. Especially in the case of adding variables, this is very low-impact, since the actual CSR data doesn't change; only the "shape" does.

I guess there is an argument to be made for a stricter kind of behavior: that no variables may be added or removed after any CSRConstraint has been added to the model. I believe that may be overly strict since the solution in this PR seems to work fine, but happy to discuss the alternative if anyone else has any insights. In any case, though, if the intention is that no variables may be added or removed when CSRConstaints are present, then this should be enforced properly at variable addition/removal. (Currently, you just get an obscure matrix alignment error when passing the model to a solver.)

Code and tests were written with the help of Claude Code.

- Added a function to reshape CSR matrix returned by `to_matrix` in case any variables were added and matrix shape is stale.
- Added a function to remap CSR columns for when a variable is removed.
- Added tests for different orders of adding variables and CSRConstraints.

## Checklist


linopy/model.py
```diff
@@ -1281,8 +1281,23 @@ def remove_variables(self, name: str) -> None:
...
+        old_vlabels = self.variables.label_index.vlabels.copy() if frozen_cons else None
+
         self.variables.remove(name)
 
+        if frozen_cons:
+            assert old_vlabels is not None  # captured whenever frozen_cons is non-empty
+            new_label_index = self.variables.label_index
+            for con in frozen_cons:
+                con._remap_columns(old_vlabels, new_label_index)
+
```
coroa: Not the right place for the remapping, this will lead to spaghetti eventually.

coroa (commented):
Oh, damn. I fucked up. I don't know anymore why i suggested active variables. I had an intermediate representation where i was going for the variables shape being mapped to the full xCounter range and then somehow ended up with changing to the active constraints first and then variables, but this remapping need is bad.

Thanks for seeing that @koen-vg .

I don't think the n.remove_variable is a good place for the remapping to happen. In principle you could also imagine variable mask changes (how do we do those?) to have the same effect! Should we have a short call how to fix this, @FabianHofmann ?

coroa:
I suggest that we hold something like a reference to the `vlabels` on the CSRConstraint and when accessing `_csr` we do:

```python
if self.vlabels is self.model.variables.index.vlabels:
   return self._csr

csr = self._csr
if (self.vlabels != self.model.variables.index.vlabels[:self.shape[1]]).any():
   # more changes than additions, ie. remap
   csr = ...

csr = sp.sparse.csr_matrix(self._csr..., ..., shape=(self.shape[0], self.self.model.variables.index.vlabels[:]))
self._csr = csr

return csr
```

koen-vg:
@coroa thanks for picking up this thread. I will admit that I was just trying to apply a simple fix that made my local modelling work, and wasn't too sure about the wider context and "morally correct" place to apply the fix.

Happy to update the PR along the lines you suggest though! I thought a little bit about it, and it seems to make sense. But I'm not super familiar with linopy internals. If you want to work on this yourself, feel free.

coroa:
> @coroa thanks for picking up this thread. I will admit that I was just trying to apply a simple fix that made my local modelling work, and wasn't too sure about the wider context and "morally correct" place to apply the fix.
> 
> Happy to update the PR along the lines you suggest though! I thought a little bit about it, and it seems to make sense. But I'm not super familiar with linopy internals. If you want to work on this yourself, feel free.

@FabianHofmann assigned the review to himself, so i was kinda hoping with the limited breadcrumbs i left, he'd whip up an amazing solution 😀

coroa:
but feel free to give it a go yourself.  the only additional comment i have would be that it probably makes sense to test two/three remapping strategies and/or think about whether there is an easy remapping cache strategy.

coroa:
my intuition would be that something like

new_colinds = np.searchsorted(old_vlabels, vlabels)[colinds] 

would be the most efficient remapping and even more so if you cache this intermediate

koen-vg:
Okay so this was my attempt at the suggestion solution. I think it holds up.

One comment, in the new implementation there is this ValueError:

```
+            referenced = old_vlabels[csr.indices]
+            new_positions = label_index.label_to_pos[referenced]
+            if (new_positions < 0).any():
+                missing = np.unique(referenced[new_positions < 0])
+                    f"Frozen constraint '{self._name}' references variables that "
+                    f"are no longer part of the model (labels {missing.tolist()}). "
+                    "Use `Model.remove_variables` to remove variables; it also "
+                    "removes the constraints referencing them."
+                )
```

Now, this isn't really supposed to happen, but if a user were to call `Variables.remove` (instead of `Model.remove_variables`), this could happen. It's probably a good idea to rename `Variables.remove` to `Variables._remove` (or delete the function entirely and inline it; it's only called from `remove_variables` right?) so nobody thinks this is a public function (even though it's not documented).

Issue #813, opened 2026-07-02
CodSpeed regression: highspy 1.15.0 addRows ~4x slower in direct HiGHS handoff (#809 bump)


FBumann:
_Intent placeholder — @FBumann to replace with your own words._

> [!NOTE]
> The following was generated by AI (investigation + reproduction of the CodSpeed failure).

## What happened

CodSpeed went red on `master` with a −16.66% regression: [failing run](https://github.com/PyPSA/linopy/runs/84771255430). All 7 regressed benchmarks are `test_to_solver[highs-*]`; gurobi and every non-solver benchmark are unaffected.

The only relevant change between the compared commits is the **highspy bump `1.13.1 → 1.15.0`** (#809). The `benchmarks` extra pins highspy exactly *specifically so dep bumps don't move the numbers* — dependabot bumped that very pin, so the run measured the dependency, and it revealed a real upstream slowdown.

## Root cause

The regression is in **`highspy.Highs.addRows`**, called by linopy's direct HiGHS handoff (`linopy/solvers.py`, `Highs._build_solver_model`). On `sparse_network(250)` the real `to_highspy` path is ~2× slower end-to-end (the `addRows` call itself ~4×):

| highspy | `to_highspy(sparse_network(250))` |
| ------- | --------------------------------- |
| 1.13.1  | ~22 ms |
| 1.15.0  | ~46 ms |

## Reproduction

Single PEP 723 script, linopy-native, runnable directly with `uv` — mirrors the CI benchmark via the pytest-benchmark `benchmark` fixture (no manual timing, no hand-built matrices). Lives at `dev-scripts/highs-addrows-repro/`; every dep but `highspy` is pinned in the accompanying `.lock`, and `highspy` is injected per run so the two versions share an identical env.

```bash
uv run --locked --with highspy==1.13.1 highs_handoff_bench.py --benchmark-columns=median,mean
uv run --locked --with highspy==1.15.0 highs_handoff_bench.py --benchmark-columns=median,mean
```

```
matrices.A rows=6000 cols=12000 stored_nnz=1506000 structural_nnz=18000 (98.8% explicit zeros)

  Name (time in ms)    Median
  test_to_highspy  [1.13.1]  22.2
  test_to_highspy  [1.15.0]  46.4
```

<details>
<summary>highs_handoff_bench.py</summary>

```python
# /// script
# requires-python = ">=3.12,<3.13"
# dependencies = [
#     "linopy",
#     "pytest",
[...]

PR #815, opened 2026-07-02
perf: drop explicit zeros from the constraint matrix (#814)

### LINKED ISSUE #814: Speed up direct solver handoff by dropping explicit zeros from the constraint matrix
_Intent placeholder — @FBumann to replace with your own words._

> [!NOTE]
> The following was generated by AI (found while investigating #813).

## Motivation

The direct solver handoff hands HiGHS a constraint matrix full of **explicitly stored zeros**. Models whose expressions broadcast against a dense coordinate (e.g. a `bus × line` incidence block) store one coefficient per coordinate pair, most of them zero. For the `sparse_network(250)` benchmark:

- `model.matrices.A`: **1,506,000 stored entries, 18,000 structural nonzeros → 98.8% explicit zeros.**

`highspy.Highs.addRows` cost scales with *stored* nnz, so we spend the handoff describing zeros (and materialise them in the solver — the peak-memory column below). This is independent of the highspy version regression in #813 — it's pure wasted work on every version.

## Proposal

Drop explicit zeros before the direct handoff. Minimal version in `linopy/solvers.py` (`Highs._build_solver_model`, ~L1668):

```python
A = A.tocsr()
A.eliminate_zeros()
...
h.addRows(num_cons, lower, upper, A.nnz, A.indptr, A.indices, A.data)
```

Mathematically identical — a zero coefficient doesn't change the constraint.

Better still, consider eliminating zeros centrally in `model.matrices.A` (or when the coefficient array is assembled) so the LP/MPS writers and the other direct backends (gurobi, xpress, …) all benefit, not just HiGHS.

## Impact

Measured with the repro in #813 (`dev-scripts/highs-addrows-repro/`, runnable via `uv run`). Timings via pytest-benchmark; peak memory via pytest-benchmem (memray). Effect of `eliminate_zeros()` on the `addRows` handoff:

| highspy | explicit zeros (1.5M nnz) | `eliminate_zeros()` (18k nnz) | speedup |
| ------- | ------------------------- | ----------------------------- | ------- |
| 1.13.1  | ~13 ms | ~1.6 ms | **8×** |
| 1.15.0  | ~51 ms | ~1.7 ms | **30×** |

Peak memory for the handoff also drops (currently ~65 MiB, dominated by the ~1.5M redundant coefficients). As a side effect this makes linopy insensitive to the #813 highspy regression.

## Notes / to check

- Confirm no backend relies on structural zeros being present.
- Decide the right layer (per-backend vs central in `matrices`).
- Add a regression benchmark on stored-vs-structural nnz.

FBumann:
This opportunity was discovered while exploring the regression introduced by highspy==1.15
Depeneding on the model, this is quite a impressive improvement. But id only merge it after someone tested it with real world models and maybe also hardened tests with other solvers.

Closes #814.

> [!NOTE]

## What & why

Expressions that broadcast against a dense coordinate (e.g. a `bus × line`
incidence block) store one coefficient per coordinate pair, most of them
structurally zero. Those explicit zeros were carried into `model.matrices.A`
and handed to every solver. `highspy.Highs.addRows` — and the other direct
backends' matrix loaders — scale with *stored* nnz, so the handoff spent most
of its work describing zeros (`sparse_network(250)`: 1.5M stored entries for
18k structural nonzeros, 98.8% zeros).

## Change

Prune explicit zeros once, centrally, in `matrices._stack` via
`eliminate_zeros()`. This covers both `A` and `indicator_A`, so **all**
consumers benefit — HiGHS, gurobi, xpress, copt, mosek and the LP/MPS
writers — not just HiGHS. A zero coefficient never changes a constraint, so
the result is mathematically identical; row count (and thus `clabels`
alignment) is preserved, and the `dual`/`sol` paths recompute independently of
`A`.

Chose the central `matrices` layer over the per-backend `_build_solver_model`
patch from the issue precisely so every backend and writer shares the win.

## Tests

- New `test_matrices_drops_explicit_zeros` asserts a dense-broadcast constraint
  yields only structural nonzeros (stored-vs-structural nnz regression).
- Existing `test_matrices.py` and `test_solvers.py` pass unchanged.

<details>
<summary>Expected addRows speedup (from #813 standalone repro)</summary>

| highspy | explicit zeros (1.5M nnz) | eliminate_zeros (18k nnz) | speedup |
| ------- | ------------------------- | ------------------------- | ------- |
| 1.13.1  | ~13 ms | ~1.6 ms | ~8× |
| 1.15.0  | ~51 ms | ~1.7 ms | ~30× |

Side effect: makes linopy insensitive to the #813 highspy `addRows` regression.
</details>

### DIFF
--- a/linopy/matrices.py
+++ b/linopy/matrices.py
@@ -22,10 +22,20 @@ if TYPE_CHECKING:
 
 
 def _stack(csrs: list) -> scipy.sparse.csr_array | None:
-    """Vertically stack CSR blocks, or None when there are none."""
+    """
+    Vertically stack CSR blocks, or None when there are none.
+
+    Explicit zeros are dropped: expressions that broadcast against a dense
+    coordinate store one coefficient per pair, most of them zero, and a zero
+    coefficient never changes a constraint. Keeping them only inflates the
+    stored nnz handed to the solvers/writers (e.g. ``highspy.addRows`` scales
+    with stored nnz), so we prune them once, centrally, for every backend.
+    """
     if not csrs:
         return None
-    return cast(scipy.sparse.csr_array, scipy.sparse.vstack(csrs, format="csr"))
+    stacked = cast(scipy.sparse.csr_array, scipy.sparse.vstack(csrs, format="csr"))
+    stacked.eliminate_zeros()
+    return stacked
 
 
 def _concat(arrays: list, dtype: type | None = None) -> ndarray:
--- a/test/test_matrices.py
+++ b/test/test_matrices.py
@@ -68,6 +68,27 @@ def test_matrices_duplicated_variables() -> None:
     assert np.isin(np.unique(np.array(A)), [0.0, 2.0]).all()
 
 
+def test_matrices_drops_explicit_zeros() -> None:
+    # https://github.com/PyPSA/linopy/issues/814
+    # Expressions that broadcast against a dense coordinate store one coefficient
+    # per pair, most of them structurally zero. Those must not reach A, whose
+    # stored nnz drives the direct-solver handoff (e.g. highspy.addRows).
+    m = Model()
+
+    x = m.add_variables(coords=[range(4)], name="x")
+    coeff = xr.DataArray(
+        np.eye(4), dims=["j", "i"], coords={"j": range(4), "i": range(4)}
+    )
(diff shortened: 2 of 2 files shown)


FabianHofmann (approved):
thanks @FBumann for detecting and fixing this!

PR #816, opened 2026-07-02
perf: prune explicit zeros at matrix assembly (cut build peak memory)

FBumann:
_Intent placeholder — @FBumann to replace with your own words._

> [!IMPORTANT]
> **Draft — depends on #815. Merge #815 first, then mark this ready.**
> This targets `master`, so until #815 merges the diff below also includes #815's commit (the `_stack` `eliminate_zeros()`). Once #815 is in, GitHub collapses the diff to just this PR's source-level prune.

Follow-up that completes the zero-drop story on the **memory** axis.

> [!NOTE]

## Why a second PR

#815 eliminates zeros from the *stacked* constraint matrix with `eliminate_zeros()`. That runs **after** `scipy.vstack` has already materialised the full dense-with-zeros block, so the **peak** allocation — exactly what the CodSpeed memory instrument tracks — is unchanged; only the resting matrix size and the solver-ingest cost drop. That's why the memory job shows no movement even though the handoff is measurably faster.

## Change

Prune at the source: fold `coeffs != 0` into the valid-entry mask in `Constraint._matrix_export_data`, so broadcast zeros never enter `cols`/`data`/the CSR at all. Snapshot capture and the matrix build both go through this path, so they stay consistent. #815's `eliminate_zeros()` stays as cheap safety for the one case a pre-filter can't catch — duplicate variable terms that cancel to zero *after* `sum_duplicates`.

## Two ripples (both handled)

1. **Dual alignment.** `matrices.dual` derived its active rows from *stored nnz* (`np.diff(csr.indptr)`). Once zeros are pruned, an all-zero-coefficient row (e.g. `0·x ≤ 5`) keeps its row in `A` but stores no entry, so it would silently lose its dual slot → misaligned dual vector. Fixed by deriving active rows from *row activity* via a new `ConstraintBase.active_row_mask` (coefficient-independent; also skips a redundant CSR rebuild).
2. **Warm-start semantics.** A zero-coefficient term no longer changes the sparsity pattern, so the persistent-solver path no longer forces a `SPARSITY` rebuild when one is added — correct, since the matrix is unchanged. `test_shape_mismatch_triggers_sparsity_rebuild` updated to use a non-zero coefficient (still exercises the real path), and `test_zero_coefficient_term_needs_no_rebuild` added for the new behaviour.

## Impact

<details>
<summary>Peak allocation building <code>m.matrices.A</code> — sparse_network(250)</summary>

| variant | build peak | final nnz |
| ------- | ---------- | --------- |
| master (keep zeros) | 50.1 MB | 1,506,000 |
| #815 (`_stack` eliminate_zeros) | 49.9 MB | 18,000 |
| this PR (prune at source) | **36.3 MB** | 18,000 |

~27% lower peak. The residual is the unavoidable dense-coeffs flatten — removing that needs expression-level sparsity, out of scope here. Solver result is identical; `addMConstr`/`addRows` stay ~2× faster from the smaller nnz.
</details>

Full test suite green (3770 passed, 45 skipped).


FabianHofmann (approved):
this looks all very sensible to me and make the structure more aligned with the dense representation of constraints. any blockers? I would merge this

FBumann:
@FabianHofmann I dont have one, but I merely filed this because I saw the opportunity. I cant confidently say that I know all possible implications fo this change.

FabianHofmann:
> @FabianHofmann I dont have one, but I merely filed this because I saw the opportunity. I cant confidently say that I know all possible implications fo this change.

it definitely make the data model and flow cleaner, so I don't think this is harmful. before the next release I will do a wider test sweep with pypsa models anyway

coroa:
> > @FabianHofmann I dont have one, but I merely filed this because I saw the opportunity. I cant confidently say that I know all possible implications fo this change.
> 
> it definitely make the data model and flow cleaner, so I don't think this is harmful. before the next release I will do a wider test sweep with pypsa models anyway

Does it, though?

There are now at least three places (this one, as well as #815, as well as the explicit sanitize_zeros)  where zeros are eliminated without us ever discussing how we want to handle zeros.

And i wanted to start exactly this discussion with asking about that in https://github.com/PyPSA/linopy/issues/814#issuecomment-4879210422 , i put that in the issue, because it felt like the best place, since it concerns both PRs rather than a single one.

Now i am checking in on the comments and the question hasn't been answered and everything is merged.

Why is this a difficult question per se? This pbb makes persistent solvers a lot less useful. Imagine a rolling horizon operations run with solar, which is time-varying. Sun-rise/sets shift over the year so the p_max_pu value which is either a right-hand side or a coefficient receives night time 0 values in different hours of the week, which leads to frequent structural changes leading to unnecessary rebuilds, so we need a way to split unnecessary from necessary 0s (and now we need this in all three places).

The impact is low at the moment, but might increase in time, so we need this discussion. I'd like if we find a way to have these discussions. Ideally without slowing down improvement speed.

FBumann:
@coroa Yes, maybe we should slow down performance work a bit and discuss those intrusive, complex things more.

PR #824, opened 2026-07-14
fix: groupby-sum fast path reduces multi-dim DataArray grouper over all its dims

### LINKED ISSUE #823: groupby().sum() returns wrong dims for a multi-dimensional DataArray grouper (fast path)
Likely a regression from #802 

> [!Note]
> created by AI

### Bug

`Variable.groupby(...).sum()` returns wrong dimensions when grouping by a **multi-dimensional** `DataArray` grouper on its default (fast) path. The reduction happens over only one of the grouper's dimensions, and the other survives in the result — producing incorrect (and silently wrong) grouped constraints/expressions.

The `use_fallback=True` escape hatch (native xarray groupby) gives the correct result.

### Root cause

In `LinearExpressionGroupby.sum` (`linopy/expressions.py`), the fast path does `group = group.to_pandas()` for a `DataArray` grouper. For a 2-D grouper this yields a `DataFrame`, which is then routed through the *multikey* encoding path — reducing over only the row axis and leaving the column axis in the output. A genuinely N-D grouper (whose values should define the groups over all its dims jointly) is not handled.

### Minimal reproducer

```python
import linopy, xarray as xr

# 2x2 grid, grouped into 2 groups by a 2-D grouper; extra dim 'k' should be preserved
grp = xr.DataArray([[1, 1], [2, 2]], dims=["i", "j"],
                   coords={"i": [0, 1], "j": [0, 1]}).rename("g")

m = linopy.Model()
v = m.add_variables(coords=[[0, 1], [0, 1], [0, 1]], dims=["i", "j", "k"], name="v")

print(v.groupby(grp).sum().dims)                    # ('j', 'k', 'g', '_term')  <- stray 'j', WRONG
print(v.groupby(grp).sum(use_fallback=True).dims)   # ('g', 'k', '_term')        <- correct
```

Plain xarray reduces the same 2-D grouper correctly over both dims, so the expected result is `('g', 'k', '_term')` (group dim + the untouched `k`).

### Impact

Silent incorrectness: expressions/constraints built via `groupby().sum()` with a multi-dim grouper are wrong unless the user knows to pass `use_fallback=True`. Surfaced while writing the `examples/sudoku.ipynb` example (grouping a `(row, column, digit)` variable by a 2-D `(row, column)` square-index grouper), which currently has to carry `use_fallback=True` as a workaround.

### Environment

- linopy `0.8.0.post1.dev109+ge87efc1f4`
- xarray `2026.4.0`

### Suggested fix

Detect a multi-dimensional `DataArray` grouper on the fast path and either handle it directly or route it to the native-xarray (fallback) implementation instead of `to_pandas()`-ing it into a multikey `DataFrame`.

FabianHofmann:
Closes #823.

This aligns the grouping mechanism across fallback and non-fallback groupby operations. 

> [!NOTE]

`LinearExpression.groupby(...).sum()` returned wrong dimensions when grouping by a **multi-dimensional** `DataArray` grouper on its default (fast) path: the reduction happened over only one of the grouper's dimensions and the others survived in the result. Only `use_fallback=True` gave the correct result.

Root cause: the fast path did `group.to_pandas()`, which turns a 2-D grouper into a `DataFrame` that is then routed through the *multikey* encoding path — reducing over only the row axis. A genuinely N-D grouper (whose values define the groups jointly over all its dims) was never handled. This is a long-standing bug present since v0.8.0, not a regression from #802.

The fix:

- Detect a genuine N-D grouper — a `DataArray` with `ndim > 1` whose dims are all data dims — and stack the data over those dims into a single key, so the fast path reduces over every grouper dim at once. This preserves the fast path's memory benefit.
- Grouper `DataArray`s whose extra axis is *not* a data dim (i.e. multikey frames built from a `DataFrame`) keep flowing through the existing `DataFrame`/multikey handling.
- `_grouped_sum` now accepts the (stacked) dataset to operate on.

## Tests

- `test_linear_expression_groupby_ndim` now exercises the fast path too (`use_fallback=[True, False]`) and asserts it equals the fallback; the stale TODO was removed.
- New `test_linear_expression_groupby_multidim_preserves_extra_dim` covers the issue's reproducer (2-D grouper over `(i, j)`, preserved `k`).

## Checklist


FBumann:
@FabianHofmann Great that you found that bug right away. Im really sorry about the regression!

EDIT: Nevermind, seems like it wasnt my PR...

FBumann:
@FabianHofmann I dived a bit deeper in and found some other issues (positional alignment when using pd.Series or xr.DataArray). This is arguably a bug, but could also be part of v1. Ill create issues for both. We should decide if we treat this as a bugfix or a semantics thing!

FBumann:
See #827

FBumann (approved):

FBumann:
@FabianHofmann This should be merged no matter what happens in #827

Issue #459, opened 2025-05-20
Properly free resources after solve with gurobipy


FBumann:
### Version Checks (indicate both or one)

- [x] I have confirmed this bug exists on the lastest [release](https://github.com/pypsa/linopy/releases) of Linopy.

- [x] I have confirmed this bug exists on the current [`master`](https://github.com/pypsa/linopy/tree/master) branch of Linopy.

### Issue Description

Using gurobipy with a non-academic license leads to issues with the license not being freed after the model was solved.
I would 

Im not finished investigating the issue, but it probably arises from not closing the gp.Model before closing the gp.Env
The gurobipy [website](https://docs.gurobi.com/projects/optimizer/en/current/reference/python/env.html) says the following: 
>It is good practice to use the with keyword when dealing with environment (and model) objects. That way the resources tied to these objects are properly released even if an exception is raised at some point. The following example illustrates two typical use patterns.

```python
import gurobipy as gp
with gp.Env("gurobi.log") as env, gp.Model(env=env) as model:
    # Populate model object here...
    model.optimize()

with gp.Env(empty=True) as env:
    env.setParam("ComputeServer", "myserver1:32123")
    env.setParam("ServerPassword", "pass")
    env.start()
    with gp.Model(env=env) as model:
        # Populate model object here...
        model.optimize()
```

I will correct this and open a PR.
Is there any reason to not close the model after the solve? Maybe because of IIS calculation?

### Reproducible Example

```python
Unfortunately I can provide a reproducible example without a usage restricted license
```

### Expected Behavior

I would expect to be able to solve a linty model in a new python thread/console, after the model in another console finished.
This is not possible, due to the license not being released properly.

### Installed Versions

<details>
bottleneck==1.5.0
click==8.2.0
cloudpickle==3.1.1
dask==2025.5.0
deprecation==2.1.0
fsspec==2025.3.2
importlib-metadata==8.7.0
linopy==0.5.5
locket==1.0.0
numexpr==2.10.2
numpy==2.2.6
packaging==25.0
pandas==2.2.3
partd==1.4.2
polars==1.29.0
python-dateutil==2.9.0.post0
pytz==2025.2
pyyaml==6.0.2
scipy==1.15.3
six==1.17.0
toolz==1.0.0
tqdm==4.67.1
tzdata==2025.2
xarray==2025.4.0
zipp==3.21.0
</details>

FabianHofmann:
[...]

PR #826, opened 2026-07-14
fix(gurobi): dispose the solver model before the env on close

FBumann:
I treid to verify that the gurobi license issue #459 is fully solved. As far as i can see its not perfect yet. I also took the opportunity to make the .close() more discoverable!

> [!NOTE]
> The following content was generated by AI (Claude Code).

Follow-up to #459. The env has been context-managed since the license-leak audit, but `Solver.close()` only *dereferenced* the `gurobipy.Model` — it disposed the env first and relied on refcount GC to free the model. Since Gurobi keeps the underlying environment (and with it the license) alive until every model created from it is freed, any user-held `model.solver_model` reference silently kept the license acquired after `close()` / `model.solver = None`.

**What changed**

- `Gurobi` now registers the `gurobipy.Model` on the env `ExitStack` (as Xpress and Mosek already do). `close()` unwinds LIFO, disposing the model before the env — the ordering the Gurobi docs require — and explicit `dispose()` releases the license even while user code still holds a reference.
- `to_gurobipy()` detaches the built model from its throwaway solver so the caller keeps ownership (its teardown would otherwise now dispose the returned model).
- `Model.solve()` docstring documents when the license is released: `model.solver.close()`, `model.solver = None`, the next `solve()` call, or model GC.

**Compatibility with older gurobipy**

The fix relies on `gurobipy.Model` supporting the context-manager protocol. Env and Model have been context managers since Gurobi 9.0, and the oldest gurobipy with wheels for Python 3.11 (linopy's floor) is 10.0 — so every installable combination is covered. Verified identical behavior (context protocol + `Model has already been freed` on disposed access) on gurobipy 12.0.1 and 13.0.2.

**Proof it was a footgun**

The new test `test_gurobi_close_disposes_held_solver_model` holds a `solver_model` reference across `close()` and asserts the model is disposed. On `master` it fails — the held model is still fully queryable, i.e. the license is still acquired.

<details>
<summary>Empirical demonstration (gurobipy 13.0.2)</summary>

```python
import contextlib
import gurobipy as gp

stack = contextlib.ExitStack()
env = stack.enter_context(gp.Env())
m = gp.Model(env=env)
m.addVar(); m.update()

held = m          # user-held reference, e.g. model.solver_model
stack.close()     # disposes the env — what Solver.close() did on master

print(held.NumVars)   # -> 1: model (and license) still alive
held.dispose()        # only now is everything released
```

New test against `master` (pre-fix):

```
FAILED test/test_solvers.py::test_gurobi_close_disposes_held_solver_model[direct]
1 failed, 5 passed, 100 deselected in 1.31s
```

With the fix: `19 passed` (`-k "close or gurobi"`), and the full gurobi/persistent sweep passes (`523 passed, 11 skipped`).

</details>

### DIFF
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -522,6 +522,7 @@ Solvers
    :toctree: generated/
 
    solvers.available_solvers
+   solvers.Solver
    solvers.CBC
    solvers.COPT
    solvers.Cplex
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -522,7 +522,6 @@ Solvers
    :toctree: generated/
 
    solvers.available_solvers
-   solvers.Solver
    solvers.CBC
    solvers.COPT
    solvers.Cplex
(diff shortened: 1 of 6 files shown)


FabianHofmann (approved):
thanks for covering this remaining bit

PR #828, opened 2026-07-14
refactor(int32): per-model label dtype instead of sticky process-global

FBumann:
This implements int dtype control per Model instead of through a global config option (global kept).
This resolves all potential issues with the global for little complexity.
I think this is the right approach.

> [!NOTE]

Builds on the auto-widening behavior from #822 (now closed; this PR rebased onto `master`), but moves the label dtype from the sticky process-global option onto the `Model` itself.

## What changed

- The label dtype becomes a `Model()` argument: `Model(dtypes={"labels": np.int64})`, default `np.int32`. It is exposed **read-only** as `Model.dtypes` — a `types.MappingProxyType` mapping (a frozen dict) with a single `"labels"` field for now, extensible to more fields later (per review feedback). Unknown keys and non-`int32`/`int64` dtypes are rejected. The `linopy.options["label_dtype"]` setting is removed entirely — embedders forward constructor kwargs instead (PyPSA: `n.optimize(model_kwargs={"dtypes": {"labels": np.int64}})`), so `config.py` returns to its pre-#566 state.
- A model widens **only itself** to `int64` when its labels would exceed the int32 maximum; the `UserWarning` recommends passing `dtypes={"labels": np.int64}` upfront.
- All cast sites read the dtype from their object's model reference; `common.save_join` takes it as an explicit `fill_dtype` parameter instead of reading the global.
- `read_netcdf` restores the widened dtype per model (from the counters, as before). Pickle round-trips preserve it automatically since it is an instance attribute.
- `OptionSettings.widen_label_dtype()` and its sticky-`reset()` semantics are removed, along with the test-suite fixture that guarded against global-state leakage.

## Why per-model

#822 chose the process-global because "the four float→int cast sites … have no reliable model reference (expressions can be detached)". That premise doesn't hold in the current codebase: `BaseExpression.__init__` raises unless `model` is a `Model` instance, and every `label_dtype` cast site is a method on an object with a validated model reference (`BaseExpression`, `Variable`, `Variables`, `Constraints`, `CSRConstraint`). The only model-less site, `save_join`, is called with `integer_dtype=True` from exactly three container properties that all have `self.model` in scope.

With a reliable model reference, a per-model monotonic widen gives the same no-narrowing guarantee at the same zero hot-path cost (an attribute read), and removes the global design's edge cases:

- **No revert footgun.** Under the global design, following the documented revert (`options["label_dtype"] = np.int32`) while a widened model is alive lets the cast sites silently narrow `>2^31` labels — `int64 → int32` `.astype()` wraps to negative values without any error. Per-model, there is nothing to revert.
- **Widening travels with the model.** A widened model shipped to another process via pickle/cloudpickle (dask, multiprocessing) keeps its dtype; the global flag does not cross process boundaries.
- **No cross-model contamination.** #822 documents that one huge model permanently costs every later small model in the process its int32 memory/speed win. Per-model, small models are unaffected.
- One knob in one place: the constructor argument. Nothing needs test isolation.

<details>
<summary>Cast-site → model-reference verification</summary>

| Site | Enclosing scope | Model access |
|---|---|---|
| `expressions.py` vars cast in init | `BaseExpression.__init__` | `model` param, enforced `isinstance(model, Model)` (check moved above the cast) |
| `expressions.py` `sanitize` | `BaseExpression` | `self.model` |
| `expressions.py` simplify (2×) | `LinearExpression` | `self.model` |
| `variables.py` `ffill`/`bfill`/`sanitize` | `Variable` | `self.model` (init enforces `isinstance(model, Model)`) |
| `variables.py` `flat` remap | `Variables` | `self.model` |
| `constraints.py` `_to_dataset` (2×) | `CSRConstraint` | `self._model` |
| `constraints.py` `flat` remap | `Constraints` | `self.model` |
| `common.py` `save_join` | free function | dtype passed by callers (`Variables.labels`, `Constraints.labels`, `Constraints.vars`) |

</details>

## Tests

`test/test_dtypes.py`: the widen tests assert other models stay `int32`; the init-arg test drives the `dtypes={"labels": ...}` mapping (+ invalid-dtype and unknown-key rejection), plus `test_dtypes_is_read_only_mapping` (mutating `m.dtypes` raises `TypeError`), `test_widen_applies_to_expressions`, and `test_auto_widen_survives_pickle`; the netcdf test asserts the loaded model is widened. The `restore_label_dtype` fixture and the option tests are gone. Full suite passes.


linopy/model.py
```diff
@@ -495,6 +508,36 @@ def solver_dir(self, value: str | Path) -> None:
...
+            "its label dtype was widened to int64. Pass label_dtype=np.int64 "
+            "to Model() when building large models to avoid the mid-build "
+            "upcast.",
+            UserWarning,
+            stacklevel=4,
+        )
+
+    def _allocate_labels(self, start: int, end: int) -> np.ndarray:
+        """Return the label range ``[start, end)``, widening the dtype on overflow."""
+        if end > np.iinfo(self._label_dtype).max:
```
FabianHofmann: should only be true if dtype!=int64

FabianHofmann (changes requested):
generalize `label_dtype` accessor to `dtypes` which is property pointing to a frozen dict with fields "labels" at first stage. can be expanded later.

FabianHofmann (approved):
great!

coroa:
This implemented auto-widening the dtypes now?

FBumann:
> This implemented auto-widening the dtypes now?

Yes! 
And this led to storing it on the model to avoid issues

PR #829, opened 2026-07-14
fix: don't leak aux coords of summed dims onto the term dimension

FBumann:
Fixes the remaining alignment-leak half of #295.

> [!NOTE]

`LinearExpression._sum` drops a summed dimension's index coordinates via `reset_index`, but an auxiliary (non-dimension) coordinate on that dimension — e.g. a `B(A)` group label — rode the `.stack()` onto the internal `_term` helper dimension. The leaked coordinate then broke later arithmetic:

```python
>>> v = v.assign_coords({"B": ("A", [311, 311, 322])})
>>> (1 * v).sum("A") >= xr.DataArray(10)
CoordinateValidationError: coordinate B has dimensions ('_term',), but these are not a subset of the DataArray dimensions ()
```

Aux coords whose dims intersect the summed dims are now dropped before stacking. Aux coords on the remaining dimensions still propagate unchanged. `QuadraticExpression.sum` and the groupby fallback route through the same `_sum` and are covered by the same change.

Only the leak is fixed here — it is a plain bug under both semantics. The other half of #295's residual (a conflicting aux coord being silently kept in binary operations) is a semantics change and stays with the v1 work (convention §11).

## Checklist

- [x] Code changes are sufficiently documented.
- [x] Unit tests for new features were added: aux coord on a summed dim is dropped (incl. the #295 reproducer), aux coord on a kept dim propagates.


FabianHofmann:
@FabianHofmann to check for groupby and other operations that collapse axes

FabianHofmann (approved):

FabianHofmann:
FYI I would like to include this in a patch release v0.8.1

PR #830, opened 2026-07-15
fix(groupby): align groupers by label, raise on mismatch (#827)

### LINKED ISSUE #827: groupby fast path matches Series/DataArray groupers by position, not by label
We have one more place where positional alignment leads to silent errors. Should we fix this immediately or as part of the v1 convention?

> [!NOTE]
> This issue was written by AI while reviewing #824. All behaviors were verified by running reproducers against current `master` and against `12e54d9~1` (the commit before #802).

### Bug

On the default (fast) path, `groupby(...).sum()` matches a `pd.Series` or `DataArray` grouper to the expression **by position**, ignoring its labels. A grouper whose index is ordered differently from the expression's coordinates silently regroups. The other grouper paths behave differently, so there are three alignment semantics side by side:

| grouper | labels reordered relative to the data |
|---|---|
| `pd.Series` / `DataArray`, fast path | silently grouped by position — **wrong result** |
| `pd.DataFrame`, fast path | reindexed by label — correct |
| any grouper, `use_fallback=True` | raises (xarray exact join) |

A grouper with a *mismatching* label set either fails deep in the scatter kernel with an unhelpful `ValueError: array is not broadcastable to correct shape`, or silently succeeds positionally when the sizes happen to match.

### Reproducer

Both groupers below encode the identical mapping `i -> g` (`{0: 1, 1: 1, 2: 2, 3: 2}`), just stored in reversed order — so all four calls should build the same expression.

```python
import pandas as pd
import xarray as xr
import linopy

m = linopy.Model()
v = m.add_variables(coords=[[0, 1, 2, 3]], dims=["i"], name="v")
e = 1 * v
```

`pd.Series` grouper:

```python
>>> s = pd.Series([1, 1, 2, 2], index=pd.Index([0, 1, 2, 3], name="i"), name="g")
>>> e.groupby(s).sum()
LinearExpression [g: 2]:
------------------------
[1]: +1 v[0] + 1 v[1]
[2]: +1 v[2] + 1 v[3]

>>> e.groupby(s.iloc[::-1]).sum()   # same grouper, index reversed
LinearExpression [g: 2]:
------------------------
[1]: +1 v[2] + 1 v[3]                # <- silently flipped
[2]: +1 v[0] + 1 v[1]
```

`DataArray` grouper — same silent flip:

```python
>>> da = xr.DataArray([1, 1, 2, 2], coords={"i": [0, 1, 2, 3]}, name="g")
>>> e.groupby(da).sum()
LinearExpression [g: 2]:
------------------------
[1]: +1 v[0] + 1 v[1]
[2]: +1 v[2] + 1 v[3]

>>> rev = da.isel(i=slice(None, None, -1))   # identical mapping i -> g, reordered
>>> e.groupby(rev).sum()
LinearExpression [g: 2]:
------------------------
[1]: +1 v[2] + 1 v[3]                # <- silently flipped
[2]: +1 v[0] + 1 v[1]
```

The fallback refuses the reordered grouper outright instead of regrouping:

```python
>>> e.groupby(rev).sum(use_fallback=True)
xarray.structure.alignment.AlignmentError: cannot align objects with join='exact' where index/labels/sizes are not equal along these coordinates (dimensions): 'i' ('i',)

The above exception was the direct cause of the following exception:

ValueError: the group variable's length does not match the length of this variable along its dimensions
```

(The final `ValueError` misdiagnoses the reorder as a length mismatch — both sides have 4 entries along `i`. xarray's `_resolve_group` re-raises *every* `AlignmentError` from its exact join with that hardcoded length message; the accurate diagnosis only survives as the chained cause.)

The same holds for N-D `DataArray` groupers (after #824) with reordered coordinates.

### History

Not a regression from #802 — verified at `12e54d9~1`: positional matching dates back to the introduction of the fast path in 769e540 (June 2023). Related to, but distinct from, #823/#824 (which fixed the *dimensions* of multi-dim groupers, not alignment).

### Suggested fix

Align indexed groupers to the expression's indexes before grouping:

- same label set, different order → reorder by label (a reindex),
- different label set → raise `ValueError`,
- grouper without an index (nothing to align by) → keep positional matching.

This makes all three grouper paths consistent and matches the v1 arithmetic convention's §8 ("shared dimensions must carry the same labels": align by label, never by position; only a label-set difference raises), so groupby groupers would follow the same rule as operands.

FBumann:
Closes #827.

@FabianHofmann We decided to NOT reindex. As all groupers should behave similarly, they now all raise if the grouper is misaligned. 
**But Dataframes were reindex before**. We could also change this to reindex all groupers instead of raising if its a pure reorder. Users relying on the reindex will get an error now...

> [!WARNING]
> **Potentially breaking.** Groupers that were previously accepted (silently regrouping by position, or — for `DataFrame` groupers — silently reindexing) now raise `ValueError`. Code that relied on a reordered grouper "just working" must reorder the grouper to match the expression's coordinates.

> [!NOTE]

### What & why

On the default (fast) path, `groupby(...).sum()` matched a `pd.Series`/`DataArray` grouper to the expression **by position**, ignoring labels — a reordered grouper silently regrouped and produced a wrong result (#827). The three grouper paths also disagreed: `DataFrame` reindexed by label, the fallback raised.

This makes all grouper types follow one rule, per the issue decision (*"Don't reindex, just check size and index fits — else raise"*): **match by label, never reindex, raise on mismatch.**

- New `_check_grouper_alignment(group, data)` validates each shared dimension's labels against the expression's coordinates and raises a clear `ValueError`, distinguishing *"same labels in a different order"* from *"a different set of labels"*. Runs on both the fast path and `use_fallback=True` (replacing the fallback's misleading *"length does not match"* message).
- Unified across `pd.Series`, `pd.DataFrame` and 1-D/N-D `DataArray`. A grouper without an index along a dim has nothing to align by and keeps the positional match.
- `_encode_multikey_group` no longer reindexes (alignment is now guaranteed upstream).

### Breaking change

`DataFrame` groupers previously reindexed a reordered index silently; they now raise like the others. `test_linear_expression_groupby_with_dataframe_non_aligned` was updated to assert the raise.

### Tests

New `TestGroupbyGrouperAlignment`: reordered Series/DataFrame/DataArray raise, N-D reorder raises, mismatched label-set raises, aligned + unindexed groupers still work (parametrized over `use_fallback`). Full `test_linear_expression.py` and all `-k group` tests pass; `ruff` and `mypy` clean.

### DIFF
--- a/doc/release_notes.rst
+++ b/doc/release_notes.rst
@@ -39,6 +39,7 @@ Upcoming Version
 * Freezing an empty constraint group (e.g. an empty ``isel`` slice) no longer raises ``ValueError: cannot reshape array of size 0``. ``Model(freeze_constraints=True)`` and ``Constraint.freeze()`` now round-trip zero-row constraints losslessly.
 * ``Variable.where`` no longer raises ``ValueError: exact match required for all data variable names`` once a solution is attached (after ``Model.solve``) or the variable is fixed. The fill value now covers auxiliary data variables (``solution``, stashed bounds) instead of only ``labels``/``lower``/``upper``.
 * ``LinearExpression.groupby(...).sum()`` with a multi-dimensional ``DataArray`` grouper now reduces over all of the grouper's dimensions on the default (fast) path, instead of leaking one of them into the result.
+* ``LinearExpression.groupby(...).sum()`` now raises when a grouper's labels are reordered or a different set relative to the expression, instead of silently regrouping by position. Reorder the grouper to match the expression's coordinates before grouping. (https://github.com/PyPSA/linopy/issues/827)
 * ``linopy.testing.assert_linequal`` now aligns dimension order before comparing, so mathematically identical expressions built in different orders (e.g. ``x + y`` versus ``y + x``, which inherit different dimension orders from xarray broadcasting) are correctly treated as equal. Genuinely different expressions still fail.
 
 Version 0.8.0
--- a/linopy/expressions.py
+++ b/linopy/expressions.py
@@ -215,19 +215,55 @@ def _unstack_multikey(ds: Dataset, dim: str) -> Dataset:
     return ds.unstack(dim, fill_value=LinearExpression._fill_value)
 
 
+def _check_grouper_alignment(group: Any, data: Dataset) -> None:
+    """
+    Ensure an indexed grouper's labels match the data along each shared dim.
+
+    The fast path matches a ``pd.Series``, ``pd.DataFrame`` or ``DataArray``
+    grouper to the expression by position, so a grouper whose coordinates are
+    reordered -- or a different set entirely -- relative to the expression would
+    silently regroup. linopy does not reindex the grouper: it checks that the
+    labels match and raises otherwise, leaving the caller to align the grouper
+    explicitly. A grouper without an index along a dimension has nothing to align
+    by and keeps the positional match.
+    """
+    shared: list[tuple[Hashable, pd.Index]]
+    if isinstance(group, (pd.Series, pd.DataFrame)):
+        shared = [(group.index.name, group.index)]
+    elif isinstance(group, DataArray):
+        shared = [
+            (dim, group.get_index(dim)) for dim in group.dims if dim in group.indexes
+        ]
+    else:
+        return
+    for dim, index in shared:
+        if dim not in data.indexes or index.equals(data.indexes[dim]):
+            continue
+        detail = (
+            "the same labels in a different order"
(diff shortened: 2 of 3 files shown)


FabianHofmann:
@FBumann one thought that came to my mind. since here we don't have the ability to set the join convention - in contrast to `x.mul(y, join="...")` I wonder if our choice of "exact" join is the correct choice here. I will take a look at this again tomorrow morning.

FabianHofmann (approved):

PR #860, opened 2026-07-21
fix(groupby): preserve grouped-dim position in groupby-sum

FabianHofmann:
> [!NOTE]

`LinearExpression.groupby(...).sum()` moved the grouped dimension to the trailing
position instead of replacing it in place, diverging from xarray's native
groupby-reduce and from linopy <= 0.8.0 (regressed in #802). `assert_linequal`
aligns dimension order, so existing tests missed the reorder.

Root cause: both the `apply_ufunc` scatter kernel (fast path) and the `groupby.map`
fallback append the new group dimension at the end. The `LinearExpression`
constructor's `xr.broadcast` then reorders every variable to the dataset's
canonical dim order — which is driven by coordinate-insertion order — so a plain
`transpose` alone doesn't stick.

The fix adds `_restore_group_dim_position`, applied to both paths:

- Transposes the terms so the group dimension(s) sit where the grouped
  dimension(s) were, surviving dims keeping their relative order.
- Reorders the dimension coordinates so the constructor's canonical order matches.
- Leaves `MultiIndex` group coordinates untouched, so multikey / `observed=True`
  results (which are order-agnostic) are unaffected and no `FutureWarning` is
  emitted.

The strict `xfail` marker on `test_linear_expression_groupby_preserves_dim_position`
is removed; the test now passes for both `use_fallback=False` and `True`.

## Checklist


FBumann:
@FabianHofmann Good catch, but it seems like this needs to be fixed differently to keep the memory low

FBumann:
Ill fix the memory issue!

FBumann:
@FabianHofmann fixed the memory issue in c5d08f422c367cef3cd6d9ec53a972b8cfef1d3e

FabianHofmann:
@FBumann great, thanks!

Issue #756, opened 2026-06-04
Umbrella: long-format / sparse `_term` kernel (dense-`_term` memory cluster)


FBumann:
This issue collects information about how to handle sparsity in linopy, while keeping the convenience of xarray broadcasting and backwards compat.

> [!NOTE]
> Generated with AI

## Root cause

A `LinearExpression` stores `coeffs`/`vars` over a **dense `coord_dims × _term` rectangle**. Every op that combines terms re-pads that axis — which is the peak allocation in real PyPSA builds. The same root cause surfaces at four entry points:

- **#748** — `@`/`dot` vs a sparse matrix
- **#745** — `groupby().sum()` pads to the largest group
- **#749** — `merge` of ragged expressions (the build peak)
- **#757** — multi-key `groupby([names])` densifies to a cartesian grid (the fast-path half, #753, shipped in #751)

Context: **#740** (polar-high benchmark — the motivation), **#712/#713** (absent/NaN semantics a sparse store must keep), **#744** (flat dims — prerequisite for a tabular core), **#630** (`CSRConstraint` — existing compressed path, export-side only).

## Direction

Two backends were weighed (full note below): `sparse.COO` under xarray (all-or-nothing; densifies at the first unsupported op) vs. a Polars long-format kernel (the form `to_polars` already emits). **Recommendation: incremental long-format on the hot ops, behind the existing dense xarray API** — convert back to dense `_term` only at the boundary, validate bit-for-bit, ship one op at a time:

1. `__matmul__`/`dot` (#748) — most self-contained, biggest single win
2. `groupby().sum()` (#745) — transient win first (`FBumann/linopy#25`), then the inherent padding
3. `merge` (#749) — concat long, rectangularize once

A full tabular core or `sparse.COO` remain open later, but are larger bets (the core is gated on #744).

*(Earlier siblings referred to "#740" as the kernel — that was a mis-reference to the polar-high benchmark; this issue is the kernel.)*

<details>
<summary><b>Full investigation note</b></summary>

# Sparsity support in linopy — investigation

> Investigation note (Claude Code, prompted by @FBumann), 2026-06-04. Assesses
> two candidate backends for sparse expression storage — sparse xarray
> (`sparse.COO` duck-arrays) vs. a tabular/long-format (Polars) kernel — and
> recommends an incremental path. Code anchors are against the
> `docs/ai-contribution-guide` working tree.

## What the issue tracker already says

[...]

Issue #745, opened 2026-06-03
groupby-sum pads every group to the largest group size


FBumann:
> **Note:** This issue was drafted by an AI agent (Claude Code) during a profiling/investigation session with @FBumann. The measurements (memray C-level profiling, benchmarks, the 284× / 132 MB figures) and the git-history archaeology are real and reproducible, but the framing and proposed directions are a starting point for discussion — please sanity-check before acting on them.

`LinearExpression.groupby(...).sum()` pads every group to the largest group size along `_term` (via `unstack(group_dim, fill_value=...)` in `expressions.py`). For skewed groups — e.g. incidence/balance constraints where a few buses host most units — the result blows up: a 300k-element expression over zipf-distributed groups produces a ~12 GiB padded result that is >99% fill.

Two separable costs:
- **Transient** (the unstack/stack machinery making extra copies) — addressable in the current representation. Prototype: FBumann/linopy#25 scatters terms directly into the result (~32× faster, ~2.6× less peak on the skewed case).
- **Inherent** (the padded result itself) — forced by the dense rectangular `_term` layout; only a long-format/sparse kernel removes it. See #756.

### Real-world evidence: PyPSA already works around this

PyPSA splits the nodal-balance constraint into **two separate constraints** — strongly-meshed vs weakly-meshed buses — purely to contain this padding (`pypsa/optimization/optimize.py`, with its own comment *"This reduces memory usage for large networks"*):

```python
meshed_buses = get_strongly_meshed_buses(n, threshold=45)   # buses with >45 attached components
define_nodal_balance_constraints(n, sns, buses=weakly_meshed_buses)             # small groups
define_nodal_balance_constraints(n, sns, buses=meshed_buses, suffix="-meshed")  # large groups
```

A single grouped balance over all buses would pad *every* bus to the largest hub's term count; bucketing by meshedness keeps each rectangle padded only to its own bucket max. This is the *"eventually do a separation of short and long linear expressions"* noted in the original groupby commit (PyPSA #557, 2023) — and it is actively maintained: PyPSA #1591 (2026) promoted `meshed_thresholds` to a tunable user parameter.

Note: on **typical** PyPSA networks the realistic group skew is small (max ≈ 8 generators/bus, ~2.7× padding), so `groupby` is not the build's peak allocation there — `merge` is (#749). The pathological skew matters for detailed unit-commitment / rooftop-PV aggregation and for the meshed hubs the split above exists to handle.

### Related

[...]

PR #867, opened 2026-07-24
perf: sparse-aware dot/matmul against a mostly-zero constant

### LINKED ISSUE #748: `@`/`dot` against a sparse matrix densifies the result to full `_term`
> **Note:** This issue was drafted by an AI agent (Claude Code) during a profiling/investigation session with @FBumann. The measurements (memray C-level profiling, benchmarks, the 284× / 132 MB figures) and the git-history archaeology are real and reproducible, but the framing and proposed directions are a starting point for discussion — please sanity-check before acting on them.

`LinearExpression.__matmul__` is `(self * other).sum(dim=common_dims)` (expressions.py). When `other` is a **sparse** matrix, the result still gets one `_term` per element of the contracted dimension — almost all of them zero-coefficient.

### Where it bites

PyPSA's Kirchhoff Voltage Law constraint (`pypsa/optimization/constraints.py`, `define_kirchhoff_voltage_constraints`):

```python
flow = m[f"{c}-s"].sel(...)        # (snapshot, branch), nterm = 1
exprs.append(flow @ C_branch * 1e5)  # C_branch: (branch, cycle), sparse
```

The cycle matrix `C` is sparse — each cycle contains only a handful of branches (~3), each branch sits in 1–2 cycles. But `flow @ C_branch` produces an expression with `_term = n_branches`, because the sum over `branch` stacks **every** branch into `_term` regardless of whether `C[b,c] == 0`.

### Measured (852 branches, 268 cycles, 24 snapshots, ~3 branches/cycle)

| | `_term` | result cells |
|---|---|---|
| current `flow @ C_branch` | 852 | 5,480,064 |
| same constraint via sparse entries + `groupby(cycle).sum()` | 3 | 19,296 |

**284× fewer cells.** Memray attributes the **single largest allocation** in a full SciGRID-DE `create_model()` (132 MB of a 351 MB peak) to the `concat`/`merge` of these densified KVL expressions — see companion issue on `merge`.

### Reframe (proof it is inherently sparse)

A branch belongs to multiple cycles, so it is not a clean `groupby` over branches. But exploding `C` to its nonzero `(branch, cycle, value)` entries makes each entry belong to exactly one cycle — a clean partition — so the constraint *is* expressible as a sparse-entry `groupby(cycle).sum()` with `nterm = max(branches per cycle)`. The prototype above does exactly this and is bit-equivalent.

### Possible directions

- A sparse-aware `dot`/`@`: when the constant operand has many zeros, drop zero-coefficient terms instead of materializing them (cf. `densify_terms`, but never building them in the first place).
- Or the general long-format/sparse `_term` kernel (#756), which subsumes this.

Sibling of #745 (groupby padding) — same root cause, different entry point. Part of the dense-`_term` memory story.

FBumann:
> [!NOTE]

Closes #748.

`LinearExpression.__matmul__` is `(self * other).sum(common_dims)`, which stacks one term per element of the contracted dimensions regardless of zeros in the operand — a sparse constant (e.g. PyPSA's cycle-incidence matrix in the KVL constraint) densifies the result to full `_term` (284× more cells than needed, the largest single allocation in a SciGRID-DE build, see #748/#749).

This adds a sparse-aware kernel using the reframe from the issue: each **nonzero entry** of the operand belongs to exactly one output group, so the contraction is a clean sparse-entry `groupby(free_dim).sum()`. The kernel selects the nonzero entries (`isel` with a flat entry dimension), multiplies each with its slice of the expression, and group-sums via the existing scatter kernel (#802). Only contributing terms are materialized; the represented expression is identical to the dense path's — only the positional term layout (count, order, padding) differs, which linopy already treats as non-semantic (`densify_terms`, zero-coeff filtering at export).

**When it engages** (all other cases fall back to the dense path unchanged):
- constant operand with at most one free (non-contracted) dimension, unique labels,
- contracted labels match the expression exactly — alignment semantics are never re-implemented, the sparse path only fires when alignment is a no-op,
- nonzero density ≤ `linopy.options["sparse_dot_max_density"]` (new option, default `0.5`; `0` disables the sparse path),
- numpy-backed data (dask keeps the dense path).

NaN entries in the operand are kept as terms (`!= 0`), matching the dense path's NaN propagation under both current and future (v1) semantics. The constant part is reduced with the same skipna semantics as `_sum`.

<details>
<summary>Benchmark (memray, KVL-shaped contraction: 852 branches × 268 cycles, 24 snapshots, 3 branches/cycle)</summary>

Op-only measurement (`memray.Tracker` around just the `expr @ C` call, native traces):

| | result `_term` | coeff+var cells | op peak | op total allocated |
|---|---:|---:|---:|---:|
| dense path (`sparse_dot_max_density=0`) | 852 | 10,960,128 | 189.2 MB | 288.9 MB |
| sparse kernel (this PR) | **3** | **38,592** | **3.9 MB** | **4.7 MB** |

284× fewer result cells (matching the figures in #748), ~48× lower op peak. Whole-process peak (interpreter + model build + op) drops 282.3 MB → 97.1 MB.

</details>

**Tests**: equivalence against the dense path via a canonical (term-layout-insensitive) comparison — incidence matrix, vector contraction, expression with constant, NaN entries, aux coords on the free dim, coordinate-less free dim — plus fallback tests (reordered labels, dense operand, duplicate free labels, >1 free dims, option disabled) asserting the dense path's exact layout is unchanged. Full suite: 3800 passed, 45 skipped.

Sibling issues #745/#749/#756 (groupby padding, ragged merge, long-format umbrella) are not touched; this is the self-contained first step of the #756 plan.

### DIFF
--- a/doc/release_notes.rst
+++ b/doc/release_notes.rst
@@ -25,6 +25,7 @@ Upcoming Version
 
 **Performance**
 
+* ``LinearExpression.__matmul__``/``dot`` against a mostly-zero constant operand (e.g. a cycle-incidence matrix) no longer materializes one term per element of the contracted dimension. A sparse-aware kernel pairs each nonzero entry of the operand with its slice of the expression and group-sums the entries, so the result only holds contributing terms — on a Kirchhoff-voltage-law shaped contraction this shrinks the result by orders of magnitude. The represented expression is unchanged; only the term layout (count, order, padding) differs. The kernel engages when the operand's nonzero density is at most ``linopy.options["sparse_dot_max_density"]`` (default ``0.5``, ``0`` disables it) and the contracted labels match the expression exactly; all other cases keep the dense path unchanged. (`#748 <https://github.com/PyPSA/linopy/issues/748>`__)
 * ``LinearExpression.groupby(...).sum()`` now scatters terms directly into the padded result arrays via ``xarray.apply_ufunc``, avoiding intermediate copies and speeding up the grouping. A single kernel covers both numpy and chunked (dask) data, the latter staying lazy. On representative models this lowers build and export peak memory by up to ~3x. The kernel emits the grouped result in its final axis order in one contiguous allocation; on dask inputs the reduction now runs over a single chunk (it no longer parallelises over the surviving dimensions). (`#802 <https://github.com/PyPSA/linopy/pull/802>`__)
 
 **Deprecations**
--- a/linopy/config.py
+++ b/linopy/config.py
@@ -62,4 +62,5 @@ class OptionSettings:
 options = OptionSettings(
     display_max_rows=14,
     display_max_terms=6,
+    sparse_dot_max_density=0.5,
 )
--- a/linopy/expressions.py
+++ b/linopy/expressions.py
@@ -101,6 +101,8 @@ if TYPE_CHECKING:
 
 FILL_VALUE = {"vars": -1, "coeffs": np.nan, "const": np.nan}
 
+ENTRY_DIM = "_entry"
+
 
 def exprwrap(
     method: Callable, *default_args: Any, **new_default_kwargs: Any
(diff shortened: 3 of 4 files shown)


FBumann:
> [!NOTE]
> AI-generated note (Claude Code, prompted by @FBumann).

Benchmark coverage: #868 adds a matmul op group to `ops.py` (`expr_matmul_dense` / `expr_matmul_sparse`) — the suite previously had no benchmark exercising `@` at all (`kvl_cycles` uses the expanded `(flow * C).sum(...)`, which bypasses this kernel). It should merge before this PR so the ops have a master baseline; this PR's CodSpeed run should then show `expr_matmul_sparse` dropping sharply while `expr_matmul_dense` stays put. Since CI runs on the merge commit, no rebase of this branch is needed once #868 lands.

FabianHofmann:
superseded by #961

Issue #757, opened 2026-06-04
Multi-key groupby([names]).sum() densifies to a dense cartesian grid (memory)


FBumann:
> [!NOTE]
> AI-written (Claude Code, prompted by @FBumann). memray numbers verified on the #751 branch.

Follow-up to #753, splitting out the part that #751 did **not** solve. #753's fast-path half shipped in #751 (multi-key `groupby([names])` now takes the reindex path, one dimension per key, non-breaking); this issue tracks the memory half.

## Problem

`expr.groupby(["period","season"]).sum()` returns separate `period` × `season` dims. In **dense** xarray that's a full cartesian grid — every absent key combination is a real fill cell. For a sparse/correlated crossing it blows up. memray, diagonal crossing (N=1000 observed combos):

| grouping | output | peak memory |
|---|---|---|
| `groupby([names])` | `{period:1000, season:1000}` dense grid | **33.3 MB** |
| `groupby(df)` | `{group:1000}` MultiIndex, observed-only | **0.33 MB** |

~100×, scaling as N. The whole difference is the final densification.

## Why it's inherent (in dense xarray)

Separate dims *are* a dense grid; the only compact form is a stacked `MultiIndex` (what the `DataFrame` grouper returns). "Separate dims **and** compact" is impossible without a genuinely sparse store.

## What #751 shipped (mitigation, not a fix)

- A `UserWarning` when the grid ≫ observed combinations, nudging users to the `DataFrame` grouper.
- The `DataFrame` grouper as the compact (observed-only, stacked) escape hatch.

## Real fix → #756

The sparse / long-format `_term` kernel (umbrella **#756**, which lists `groupby` densification as an entry point). Under a long-format kernel, `groupby(k).sum()` is a relational `group_by().agg()` over observed combinations only — no grid, no padding.

Issue #749, opened 2026-06-04
`merge` of ragged expressions is the peak allocation in PyPSA model builds


FBumann:
> **Note:** This issue was drafted by an AI agent (Claude Code) during a profiling/investigation session with @FBumann. The measurements (memray C-level profiling, benchmarks, the 284× / 132 MB figures) and the git-history archaeology are real and reproducible, but the framing and proposed directions are a starting point for discussion — please sanity-check before acting on them.

`merge` (expressions.py, `merge(...)`) concatenates several `LinearExpression`s along a new/shared dimension by aligning their `_term` axes. Because the `_term` axis is a dense rectangle, alignment pads **every** block to the **global** maximum term count before concatenation, and the concatenation itself allocates the full padded result.

### Evidence

Memray on a full SciGRID-DE `create_model()` (585 buses, 24 snapshots, 59,640 vars, 142,968 cons) — peak C-level memory **351 MB**, and the single largest live allocation at the high-water mark:

```
132.6 MB  concatenate (xarray duck_array_ops)
          <- concat (xarray)
          <- merge (linopy/expressions.py)
          <- define_kirchhoff_voltage_constraints (pypsa/.../constraints.py)
```

So `merge` is the **peak operation** of the whole build — i.e. the allocation that actually sets the OOM ceiling for large networks, more than `groupby` (#745) on realistic group-size distributions.

Two contributing factors:
1. The inputs are already densified (see the KVL `@` issue) — `merge` inherits that bloat.
2. `merge` pads each block to the global-max `_term` before concat; blocks with few terms waste the rest.

### Why it matters

PyPSA works around exactly this by **splitting the nodal-balance constraint into two** (strongly- vs weakly-meshed buses) so each `merge`/`groupby` operates on a bucket of similar term counts — see #745 for that evidence. A `merge` that handled ragged `_term` (or operated in long format) would remove the need for that manual bucketing.

### Possible directions

- Concatenate in long/COO form and only rectangularize once, padding to the per-result max rather than per-input-then-global.
- Or the long-format/sparse `_term` kernel (#756).

Sibling of #745 and the KVL `@` issue — all three are the dense-`_term` representation surfacing at different ops.

FBumann:
**Note:** Same session (AI agent + @FBumann); env as in the #745 comment (linopy 0.7.0 / pypsa 1.2.2.post1.dev2+g30e4ed0ef / xarray 2026.4.0, build-only).

[...]

PR #870, opened 2026-07-24
feat(v1): CSR-backed sparse groupby-sum - skew-independent build memory (6-9x on the #745 hub case)

FBumann:
This is a drive by PR done with CLAUDE FABLE. Its here to explore how we can better ahdnle sparsity.

Not sure where this goes yet.

> [!NOTE]

**pytest-benchmem, the #745 hub scenario (nodal balance, 120 buses × 24 snapshots, one hub bus with `hub` generators, 100 on each other bus; per-benchmark allocator peak of the build):**

| hub generators | dense groupby build peak | `sparse=True` + `freeze=True` | median build time |
|---:|---:|---:|---|
| 8,000 | 277.1 MiB | **43.5 MiB** (6.4×) | 19.4 ms → **12.9 ms** |
| 16,000 | 546.1 MiB | **61.0 MiB** (9.0×) | 38.5 ms → **16.5 ms** |

Doubling the skew doubles the dense peak but only grows the sparse one by its actual new terms: the sparse path is O(nnz) and independent of the group-size distribution, so the padding cost that PyPSA's meshed-bus banding exists to contain (see #745) is gone rather than capped. Constraint export also gets cheaper, because the result already *is* the export representation.

## What this adds

Addresses the inherent half of #745 (and #757) on the v1 line, following the umbrella plan in #756: `groupby(g).sum()` must pad every group to the largest group's term count because a `LinearExpression` stores terms in a dense `cells × _term` rectangle. The padded cells are pure intermediate waste for constraint-bound results — the LP/matrix export drops them again.

`expr.groupby(g).sum(sparse=True)` (or `linopy.options["sparse_groupby"] = True`) builds the grouped sum in **CSR form behind the unchanged `LinearExpression` type** — modeled on dask-backed xarray objects: same public class, different backing, no new public primitive. The payload (`linopy/csr.py`, `CSRPayload`) stores the expression as `A @ x + c`: one CSR row per result cell, one column per variable label, plus a dense per-cell constant. The `_term` axis is ragged by construction, so group-size padding has no analog, and the operations between a groupby and its constraint become sparse linear algebra:

- grouping scatters members into group rows (conceptually `G @ A` with a 0/1 grouping matrix),
- `merge` along `_term` — and therefore `+`/`-` — is sparse matrix addition,
- unary minus / scalar multiplication scale values,
- `== rhs` is carried as a pending payload on the `Constraint`,
- `Model.add_constraints(..., freeze=True)` (or `Model.freeze_constraints`) staples sign and rhs on and registers a `CSRConstraint` directly — the existing frozen backend (#630), so the LP writer and matrix export work unchanged.

Anything without a sparse branch transparently expands to the dense rectangle through the `.data` property and proceeds exactly as today, so compatibility is a fallback, not a constraint.

## The contract, and why it is v1-gated

The CSR form is canonical: duplicate variables within a cell are summed (`2x + 3x → 5x`) and terms are ordered by variable label. Expanding back to the dense form therefore yields the **mathematically identical expression in canonical term layout**, not the eager kernel's exact positional layout. Term layout is non-contractual under v1 (linopy already treats it as such at export: zero-coeff filtering, `maybe_group_terms_polars`, `densify_terms`), which is why the feature requires `options["semantics"] = "v1"` — under legacy, `sparse=True` raises and the option is ignored. Equivalence is asserted at the level that matters: identical polars constraint rows and identical LP files (the test suite includes an end-to-end LP diff, term order within a row canonicalized).

v1 parity is kept in the direct realization: NaN in the rhs raises (§5), a reordered/differing rhs index raises with the standard alignment message (§8), and rows whose constraint is absent realize as masked (§12). Labels and the constraint grid bit-match the dense path.

## Reproduce

The suite gains a sparse sibling of the existing `nodal_balance` pattern (the #745 severity sweep), so the padding cost is CI-visible on CodSpeed and one command locally:

```sh
pytest benchmarks/ -k "nodal_balance and build" --benchmark-memory
```

| build peak (KiB) | severity 0 | 50 | 100 |
|---|---:|---:|---:|
| `nodal_balance` (dense) | 951 | 5,460 | 9,909 |
| `nodal_balance_sparse` | 1,524 | 1,524 | 1,524 |

<details>
<summary>Standalone repro (single pytest file — the headline table above is its output)</summary>

```python
# repro745.py — the #745 hub scenario, dense vs sparse (CSR) groupby-sum.
# run:  pytest repro745.py --benchmark-memory
import numpy as np
import pandas as pd
import pytest
import xarray as xr

import linopy


def build_balance(sparse: bool, hub: int) -> linopy.Model:
    linopy.options["semantics"] = "v1"
    n_bus, n_snap = 120, 24
    buses = pd.RangeIndex(n_bus, name="bus")
    gen_of_bus = np.repeat(np.arange(n_bus), [hub] + [100] * (n_bus - 1))
    gens = pd.RangeIndex(len(gen_of_bus), name="gen")
    snaps = pd.RangeIndex(n_snap, name="snapshot")

    m = linopy.Model()
    gen_p = m.add_variables(coords=[gens, snaps], name="gen_p")
    load = xr.DataArray(np.ones((n_bus, n_snap)), coords=[buses, snaps])

    grouper = pd.Series(gen_of_bus, index=gens, name="bus")
    supply = (1 * gen_p).groupby(grouper).sum(sparse=sparse)
    m.add_constraints(supply == load, name="balance", freeze=sparse)
    return m


@pytest.mark.parametrize("hub", [8000, 16000])
@pytest.mark.parametrize("sparse", [False, True], ids=["dense", "sparse"])
def test_hub_balance_build(benchmark, sparse, hub):
    benchmark(build_balance, sparse, hub)
```

Needs `pytest-benchmark` and `pytest-benchmem` (both in the `benchmarks` extra). Measured on this branch, macOS arm64, python 3.13. Both paths produce identical constraints — the test suite asserts identical polars term rows and identical LP files for this construction.

</details>

## Scope and current limitations

- Single-key groupers over an existing dimension; multi-key / multidim groupers, `use_fallback` and `observed` take the eager path.
- The sparse branches cover the constraint-building chain (neg, scalar mul, merge/`+`/`-` on the same grid, comparison with a constant rhs); everything else materializes canonically.
- `freeze=False` falls back to a dense `Constraint` (mathematically equal, canonical layout).
- Quadratic expressions are untouched.

Natural follow-ups on this seam (not in this PR): `dot` as a payload op (the v1-side counterpart of #867, which stays the master-era #748 fix) and ragged/KVL-style merge (#749).

## History

Two commits kept deliberately: the first implements the same public behavior with a deferred-recipe payload (ungrouped parts + groupers, bit-identical fallback via replaying the eager kernel); the second swaps the payload to CSR. The recipe variant measured within ~5 % of the CSR numbers but accumulates unbounded part lists, holds all input expressions alive until realization, and every future op would need its own replay logic — CSR is the representation the rest of #756 (dot, merge) composes on. The swap trades the recipe's bit-identical fallback for the canonical-form contract above.

Full suite: 6336 passed, 557 skipped — unchanged from the base branch. #717 (`feat/arithmetic-convention`) has landed on master, so this branch is now rebased directly onto master and is no longer stacked.


FabianHofmann:
> [!NOTE]
> The following content was generated by AI (Claude Code), reviewed by @FabianHofmann.

## Follow-up once #926 (label columns for `CSRConstraint`) lands

Parts of this PR are a stopgap for the fact that `CSRConstraint` stores dense variable positions. With label columns, the following is deleted or simplified, nothing else moves:

- **Delete `Constraint._from_pending`, the lazy branch in `Constraint.data`, and `extract_pending`** (`linopy/constraints.py`). `LinearExpression.to_constraint` on a payload-backed expression returns an unassigned `CSRConstraint` (`cindex=None`) directly, so no pending `(lhs, sign, rhs)` state is needed.
- **Simplify `CSRConstraint.from_payload`**: drop the label-to-position mapping and the label-index dependency. It becomes a thin constructor from payload, sign and rhs; the `has_terms` / `eliminate_zeros` ordering stays.
- **Reduce the dispatch in `Model.add_constraints`** to "if the incoming constraint is an unassigned `CSRConstraint`, assign labels", which is the step the dense path already performs.

Unchanged: `linopy/sparse_expression.py` (`CSRPayload`, `try_csr_merge`), the `_payload` hooks on `LinearExpression`, `from_payload` as the sparse sibling of `from_mutable`, and `_rhs_grid_values`.

Note that frozen constraints produced here inherit the bug in #926: adding or removing variables after a sparse `freeze=True` constraint fails at matrix assembly, exactly as for the pre-existing dense `freeze=True` path.

FabianHofmann (approved):
highly needed for the math-spec integration in linopy (for pypsa networks with loads of connections). gave it a push. merging now

PR #882, opened 2026-08-06
Expressions container

### LINKED ISSUE #881: Attach expressions (incl. lazy ones) to linopy.Model
I would like to be able to keep some linear/quadratic expressions close to the rest of the model representation, in part so I can have equivalents to the expressions I would define in text-based math (#561) and in part so I can have a store of expressions useful to my analysis that persist with the model when it is passed around or dumped to file. 

I can see two types of expressions being useful to define:

1. Formalising an `Expressions` dataclass that sits as an attribute of the Model class, which stores LinearExpression or QuadraticExpression objects in the same way `Variables` and `Constraints` work at the moment.
2. Adding some lazy equivalent (`LazyExpressions`?) which doesn't evaluate the expression but rather takes a callable and only promotes it to an `Expression` member on request. This would work well with the text math as you could define any number of expressions and have only those which will be used by downstream constraints or objectives actually evaluated; the rest only need evaluating on-the-fly when requesting their solutions (they could be statistics of your model runs that you just want to calculate in postprocessing). 

I had a few thoughts on the LazyExpressions that I haven't fully resolved in my head:
1. Should they always be promoted on first evaluation or not? I'm leaning towards having a separate `promote` method which explicitly promotes them, otherwise they can spit out the evaluated expression on calling `evaluate` but the linopy model still stores them as a LazyExpression.
2. Should it be possible for them to store an explicit AST or is a callable OK? From the perspective of having a consistent Python API that allows these to be created interactively in Python or (in future) via the text math, I was thinking of keeping it as a callable. But if we limited it to just working with the text math, it could also hold the AST, knowing that it is a pyparsing object and having explicit functionality to deal with that.
3. How to handle arithmetic operations? I guess easiest is to always evaluate them if any kind of arithmetic is applied, but then you can't have `lazy + lazy -> lazy`. One could also defer the operations by storing the callables and their connecting arithmetic, then you could get a LazyExpression out of combining two lazy expressions (you probably _wouldn't_ want to support `lazy + variable -> lazy`).

I've implemented some of this already in my text math draft PR (#871) but think it warrants separate discussion and a separate PR, to precede #871.

brynpickering:
Partially closes #881 

> [!NOTE]
> Initial implementation in #871 was human-generated. I've cherry-picked that and then iterated on it with Claude to close some implementation gaps and to add tests/docs.

> [!NOTE]
> This does not address the concept of lazy expressions from #881 as I wanted to keep this PR lightweight and less controversial. 

``Model.add_expressions`` registers a ``LinearExpression`` or ``QuadraticExpression`` under a name (auto-generated as ``expr0``, ``expr1``, ... if omitted), accessible afterwards via ``Model.expressions`` (an ``Expressions`` container mirroring ``Model.variables``/``Model.constraints``) and removable via ``Model.remove_expressions``.
  Named expressions are persisted by ``Model.to_netcdf``/``linopy.read_netcdf`` and preserved by ``Model.copy``, ``copy.copy``, ``copy.deepcopy``, and pickling.

## Checklist


test/test_expressions.py
```diff
@@ -0,0 +1,141 @@
...
+        m.add_expressions(x + 1, name="expr_x")
+
+
+def test_add_expressions_from_variable_and_tuples() -> None:
+    m = Model()
+    x = m.add_variables(name="x")
+
+    expr = m.add_expressions(x, name="from_var")
+    assert isinstance(expr, LinearExpression)
+    assert_linequal(expr, x.to_linexpr())
```
FabianHofmann, line 83:
```suggestion
    assert_linequal(expr, x.to_linexpr())
    assert_linequal(expr, m.expressions["from_var"])
```

test/test_expressions.py
```diff
@@ -0,0 +1,141 @@
...
+
+    expr = m.add_expressions([(2, x)], name="from_tuples")
+    assert isinstance(expr, LinearExpression)
+    assert_linequal(expr, 2 * x)
```
FabianHofmann, line 88:
```suggestion
    assert_linequal(expr, 2 * x)
    assert_linequal(expr, m.expressions["from_tuples"])
```

FabianHofmann (changes requested):
this is extremely non-controversial. let's pull this in. Just to test potential future regressions let's also catch the case that returned and stored expression are the same (as suggested below)

brynpickering:
@FabianHofmann @FBumann I can't really understand the memory benchmarking results. It's showing a regression in one and massive improvements in others. I can understand regressions _if_ the expressions container is being used but I can't understand performance improvements: there should be none.

FabianHofmann:
@brynpickering I think the benchmarks are in a dirty state where the reference run is from master *before* several package version updates and not the latest master. At least this is how I understood @FBumann. So let's ignore it and just look at the other tests

FBumann:
Merging master into here should resolve this
But I'm having some flakiness issues with codspeed in lpspec too at the moment, so might be a problem on their side

FabianHofmann (approved):

PR #885, opened 2026-08-09
fix(piecewise): declare ragged breakpoints with mask= instead of inferring NaN

### LINKED ISSUE #884: v1 semantics: ragged (NaN-padded) breakpoints raise "NaN found in a user-supplied constant" in add_piecewise_formulation
likely the last issue I found for fully integrating v1 into pypsa

> [!NOTE]
> The content of this issue was generated with AI assistance.

## Version Checks

- [ ] Latest release — n/a, the `semantics` option does not exist there.
- [x] Current [`feat/arithmetic-convention`](https://github.com/PyPSA/linopy/pull/717) branch (`34694331`).

## Issue Description

Under `linopy.options["semantics"] = "v1"`, `add_piecewise_formulation` raises

```
ValueError: NaN found in a user-supplied constant. linopy treats this as ambiguous: ...
```

for **ragged breakpoints** — i.e. entities with different numbers of breakpoints, NaN-padded along `_breakpoint`. That padding is linopy's own ragged representation: `_pad_ragged` builds it (`np.full((len(keys), max_len), np.nan)`), and `_strip_nan` / `_has_trailing_nan_only` consume it. So the NaN is structural, not a data error.

The formulation code already knows which pieces are padding — `_add_incremental` computes a `piece_mask` from `bp_valid` (`piecewise.py:1930-1932`) — but the arithmetic that trips the v1 NaN guard runs **before** that mask is applied:

- `_incremental_weighted` (`piecewise.py:1806`) evaluates `(delta_var * steps).sum(dim=piece_dim) + bp0_term`, where `steps` are consecutive breakpoint differences and so are NaN in the padded slot.
- `_tangent_lines_impl` (`piecewise.py:769`, called from `_add_incremental` at `:1937`) evaluates `slopes * _to_linexpr(x) + intercepts` with NaN `intercepts` from the same padding.

Under legacy this is silently tolerated, and linopy's own deprecation notice points straight at the site:

```
piecewise.py:1806: LinopySemanticsWarning: NaN in the multiplicative factor was silently
treated as 0 by legacy (so the variable was zeroed out at that slot). Under v1 this raises
ValueError.
  return (delta_var * steps).sum(dim=piece_dim) + bp0_term
```

All three formulation paths are affected (`lp`, `incremental`, and `incremental` with `active=`), so it is not specific to a method or to unit-commitment gating.

This blocks PyPSA's v1 migration ([PyPSA/PyPSA#1829](https://github.com/PyPSA/PyPSA/pull/1829)): piecewise cost/efficiency curves are naturally ragged across components, and the two remaining v1 test failures there both land in these two functions.

## Reproducible Example

```python
import numpy as np
import pandas as pd
import xarray as xr
import linopy
from linopy import Model, breakpoints
from linopy.constants import BREAKPOINT_DIM

names = pd.Index(["a", "b"], name="name")
dims = ["name", BREAKPOINT_DIM]

# 'a' has 3 breakpoints, 'b' only 2 -> trailing NaN pad
X = xr.DataArray([[0.0, 50.0, 100.0], [0.0, 100.0, np.nan]], coords={"name": names}, dims=dims)


def build(y_values, use_active):
    y = xr.DataArray(y_values, coords={"name": names}, dims=dims)
    m = Model()
    x_var = m.add_variables(0, 100, coords=[names], name="x")
    y_var = m.add_variables(coords=[names], name="y")
    kwargs = {}
    if use_active:
        status = m.add_variables(binary=True, coords=[names], name="s")
        kwargs["active"] = status.to_linexpr()
    m.add_piecewise_formulation(
        (y_var, breakpoints(y), ">="), (x_var, breakpoints(X)), name="pw", **kwargs
    )


CASES = {
    "lp (convex)": ([[0.0, 30.0, 100.0], [0.0, 60.0, np.nan]], False),
    "incremental (concave)": ([[0.0, 70.0, 100.0], [0.0, 60.0, np.nan]], False),
    "incremental + active": ([[0.0, 70.0, 100.0], [0.0, 60.0, np.nan]], True),
}

for mode in ("legacy", "v1"):
    linopy.options["semantics"] = mode
    for label, (y_values, use_active) in CASES.items():
        try:
            build(y_values, use_active)
            print(f"{mode:7s} | {label:22s} | OK")
        except Exception as e:
            print(f"{mode:7s} | {label:22s} | {type(e).__name__}")
```

Output:

```
legacy  | lp (convex)            | OK
legacy  | incremental (concave)  | OK
legacy  | incremental + active   | OK
v1      | lp (convex)            | ValueError
v1      | incremental (concave)  | ValueError
v1      | incremental + active   | ValueError
```

## Expected Behavior

All six rows print `OK`. NaN padding along `_breakpoint` marks *absent pieces*, which linopy itself created and already detects via `bp_valid` / `piece_mask` — it should not be reported to the user as an ambiguous NaN constant.

Applying the existing `piece_mask` (or resolving the padded slots) before the arithmetic in `_incremental_weighted` and `_tangent_lines_impl` would fix both sites. If instead ragged input is meant to be rejected under v1, a dedicated error naming the ragged entity would be far easier to act on than the generic NaN-constant message, which gives no hint that padding is the cause.

## Installed Versions

<details>

```
linopy   0.8.0.post1.dev140+g346943317  (PR #717 head, 34694331)
xarray   2026.7.0
pandas   2.3.3
numpy    2.4.6
python   3.12.3
```

</details>

FBumann:
This PR fixes #884. It takes the simplest, strict path, disallowing nan like everywhere else, except if a matching mask is passed. This works without adding a new primitive like `Breaks`, which would be another option but a bigger api change and less flexibility (or a complex, dataarray-like class).

> [!NOTE]
> The content below this line was generated by AI.

_Stacked on #717 (`feat/arithmetic-convention`). Fixes #884, which blocks PyPSA/PyPSA#1829._

## What this changes

Ragged curves — entities with different numbers of breakpoints — are stored densely along `_breakpoint` with the surplus slots left absent. Under v1 that tripped the §5 user-NaN guard from deep inside the formulation's arithmetic:

```
ValueError: NaN found in a user-supplied constant. ... if you meant *absent at this slot*,
mark it on the variable instead (mask=, .where(cond), .reindex(...), .shift(...))
```

— advice that does not apply to breakpoint data, raised from `(delta_var * steps).sum(...)` rather than from the input.

**§5 is right to refuse the input.** A shorter curve and a stray NaN look identical, and linopy trusts NaN only from its own structural operations (§4). What was missing is a way to *declare* the absence. So:

- **`add_piecewise_formulation(..., mask=...)`** — §4's mechanism, already the vocabulary of `add_variables`/`add_constraints`. It declares which `(entity, breakpoint)` slots hold a real breakpoint and becomes the authoritative breakpoint mask, replacing the `isnull()` inference. For NaN-padded input the declaration is `mask=x_pts.notnull()`.
- **Legacy is unchanged** except for a `LinopySemanticsWarning`: it keeps inferring from NaN placement, so nothing breaks before opt-in.
- **Padding must still not reach the arithmetic as a *constant*.** Provenance is gone by the time a breakpoint table multiplies a variable, so §5 applies to it there. Those coefficients are zeroed — the absence is already carried by the masked variable and propagates on its own (§6).
- **`tangent_lines`** has no mask to declare with, so it can only report; it now names raggedness instead of surfacing the generic user-NaN message.

## Scope: five paths, not three

The issue lists `lp`, `incremental`, and `incremental + active`. `sos2` and disjunctive fail identically — they reach the padded slot through `(lambda_var * links.eq_bp)` instead. All five are covered.

## Resulting behaviour

| | legacy | v1 |
|---|---|---|
| ragged, undeclared | builds (+ deprecation warning) | raises at the boundary, naming `mask=` |
| ragged, `mask=` declared | builds | builds |

Verified across `lp` / `sos2` / `incremental` / `incremental + active` / disjunctive.

<details>
<summary>Verification detail</summary>

- Objectives match legacy exactly on all three solvable cases; the `lp` path's LP file is byte-identical.
- The `incremental` LP file differs from legacy only by three redundant rows (`-1.0 x5 <= -0.0`) that legacy emits for fully-masked `delta_bound` / `fill_order` constraints and v1 correctly drops. Pre-existing v1 improvement, not from this PR.
- Analytic check: on the short curve `y: 0 → 60` over `x: 0 → 100`, `x=50` gives `y=30`.
- `test_declared_mask_matches_legacy_oracle` reuses the oracle from the existing legacy-only padding tests (`f_b(10) = 12.5`) to assert the declared model *is* the model legacy inferred.
- Full suite: 5074 passed, 446 skipped. `ruff` clean; `mypy` clean (the one `solvers.py:297` error pre-exists on the base branch).

</details>

## Notes for review

- **The existing tests already anticipated this.** `test_lp_per_entity_nan_padding` and `test_sos2_per_entity_nan_padding` were marked `@pytest.mark.legacy` with the comment *"Legacy-only: NaN-as-mask in user input (see convention.md §5)"*. Those models now run under v1 too, via `mask=`.
- **No spec change.** This is an application of §4/§5, not a new rule, so `doc/design/convention.rst` is untouched.
- **`Slopes(align="leading")` is arguably inconsistent** with §5 — it *requires* `values[0]` to be NaN as a positional sentinel in user data. Out of scope here; worth a separate issue if you agree it is one.
- **PyPSA impact**: the migration becomes "declare the absent breakpoints" (`mask=x_pts.notnull()`), not "linopy tolerates the NaN".

## Alternatives considered

- *Accept trailing NaN as a documented positional contract* — smallest change, but has linopy infer meaning from a user NaN, which is what §5 exists to prevent.
- *A `Breaks` carrier returned by `breakpoints()`/`segments()`* — carries provenance so ragged `dict` input stays undeclared, but changes what those public factories return. Rejected in favour of zero new public types; the cost is that ragged `dict`/list input must now be declared too.


linopy/piecewise.py
```diff
@@ -830,6 +836,13 @@ def tangent_lines(
         "entirely with "
         '`warnings.filterwarnings("ignore", category=linopy.EvolvingAPIWarning)`.',
     )
+    # No mask to declare absence on this low-level helper, so a ragged curve
+    # can only be reported (§5), not resolved — say so here rather than let
+    # the generic user-NaN message surface from the chord arithmetic.
```
FabianHofmann: drop this

linopy/piecewise.py
```diff
@@ -983,6 +996,69 @@ def _paired_valid_points(*points: DataArray) -> DataArray:
     return points[0].where(~invalid)
 
 
+def _drop_absent(values: DataArray, mask: DataArray | None) -> DataArray:
+    """
+    Make a breakpoint table safe to use as a *constant* operand.
+
+    Absent slots are marked with ``NaN`` (§2), but as soon as the table
```
FabianHofmann: reference to convention in code.

FabianHofmann (commented):
wonderful and faster than allowed. There is some references in the code (docstrings & comments) which point to the paragraphs of the convention. I think we should keep those out and all these would need to be updated once we change the convention. let's keep this information absolute

doc/release_notes.rst
```diff
@@ -35,6 +35,7 @@ Upcoming Version
...
 * ``add_variables(binary=True, ...)`` now accepts ``lower``/``upper`` bounds, as long as they are 0 or 1. Previously binary bounds could only be set via the ``.lower``/``.upper`` setters after creation. (https://github.com/PyPSA/linopy/issues/776)
+* ``add_piecewise_formulation`` gained a ``mask`` parameter declaring which breakpoint slots hold a real breakpoint. It is needed for **ragged** curves — entities with different numbers of breakpoints — which are stored densely with the surplus slots left absent. Under v1 that absence must be declared (``mask=x_pts.notnull()``) rather than read off the NaN padding, since §5 does not let linopy tell a shorter curve from a data error; legacy keeps inferring it, with a ``LinopySemanticsWarning``. Previously ragged curves failed under v1 with a generic "NaN found in a user-supplied constant" raised from deep inside the formulation's arithmetic, on all of the ``lp``, ``sos2``, ``incremental`` and disjunctive paths. (https://github.com/PyPSA/linopy/issues/884)
```
FabianHofmann:
```suggestion
* ``add_piecewise_formulation`` gained a ``mask`` parameter declaring which breakpoint slots hold a real breakpoint. It is needed for **ragged** curves — entities with different numbers of breakpoints — which are stored densely with the surplus slots left absent. Under v1 that absence must be declared (``mask=x_pts.notnull()``) rather than read off the NaN padding. ```

File: linopy/piecewise.py (hunk 2 above)
Comment (FabianHofmann) [reply linopy/piecewise.py]: make the docstring a one-liner

File: linopy/piecewise.py [hunk 4]
```diff
@@ -1183,6 +1260,25 @@ def add_piecewise_formulation(
...
+        rather than read off the data: a shorter curve and a stray NaN look
+        identical, so linopy refuses to guess (see the convention, §4/§5).
+        For breakpoints that are already NaN-padded, the declaration is
+        ``mask=x_pts.notnull()``.
+
+        Slots marked absent are excluded from the formulation: no auxiliary
+        variable is created for them and no constraint references them.
+        Passing a mask that hides a *present* value is allowed and drops
+        that breakpoint; a NaN at a slot the mask calls present is a data
+        error and raises.
```
FabianHofmann: should be more compact

FabianHofmann (changes requested):
some more things

FabianHofmann (approved):

FabianHofmann:
wonderful! let's merge this

PR #886, opened 2026-08-11
fix: ignore masked variable labels when matching constraints

### LINKED ISSUE #883: `remove_variables` deletes unrelated constraints: `-1` means "masked" in a variable but "empty term slot" in a constraint

- [x] I have confirmed this bug exists on the current [`master`](https://github.com/pypsa/linopy/tree/master) branch of Linopy.

## Summary

Since 0.8, `Model.remove_variables(name)` deletes every constraint for which `Constraint.has_variable(variable)` is true. That test is

```python
# linopy/constraints.py
def has_variable(self, variable: variables.Variable) -> bool:
    return bool(self.data["vars"].isin(variable.labels.values.ravel()).any())
```

`-1` appears on both sides of that comparison with two unrelated meanings:

- in `variable.labels`, `-1` marks an entry masked out by `mask=`;
- in `constraint.data["vars"]`, `-1` marks an unused term slot in the padded  `_term` dimension.

So **any masked variable tests as "used by" any constraint that has a padded term slot**, and removing the former silently deletes the latter. Models that use `mask=` at all are affected, and the more masking there is, the more constraints are destroyed.

The failure is quiet: the model still builds, it has simply lost constraints, so it typically goes on to solve unbounded or to a wrong optimum. The emitted `UserWarning` names constraints that genuinely have nothing to do with the removed variable, which makes it read like a false alarm rather than a symptom.

### Reproducible Example

```python
import pandas as pd
import xarray as xr
import linopy
from linopy import Model

print("linopy", linopy.__version__)

idx = pd.Index([0, 1, 2], name="i")
m = Model()

a = m.add_variables(
    name="a", lower=0, coords=[idx],
    mask=xr.DataArray([True, False, True], coords=[idx]),
)
b = m.add_variables(
    name="b", lower=0, coords=[idx],
    mask=xr.DataArray([True, True, False], coords=[idx]),
)

m.add_constraints(b >= 1, name="only_b")

print("a labels   :", a.labels.values)                                  # [ 0 -1  2]
print("b labels   :", b.labels.values)                                  # [ 3  4 -1]
print("only_b vars:", m.constraints["only_b"].data["vars"].values.ravel())  # [ 3  4 -1]

m.remove_variables("a")
print("only_b still in model:", "only_b" in m.constraints)
```

### Expected Behavior

`only_b` should survive: it does not use `a`. More generally, `has_variable` should ignore the `-1` sentinel on both sides.

## Suggested fix

Restrict the membership test to real labels:

```python
def has_variable(self, variable: variables.Variable) -> bool:
    labels = variable.labels.values.ravel()
    labels = labels[labels >= 0]
    used = self.data["vars"]
    return bool(used.where(used >= 0).isin(labels).any())
```

(or filter `used` to `>= 0` first — the point is that neither side should let `-1` match `-1`.)

### Installed Versions

<details>
- linopy 0.7.0 (correct), 0.8.0 and 0.9.0 (both affected)
- Python 3.13, xarray 2025.12.0, pandas 2.x
</details>

YassineAbdelouadoud:
I encountered this issue while updating to the latest linopy version and running the [POMMES framework ](https://git.persee.minesparis.psl.eu/energy-alternatives/pommes)

Fixes #883

> [!NOTE]

### The bug

`Model.remove_variables` removes every constraint for which
`constraint.has_variable(variable)` is true. `has_variable` compared the
constraint's `vars` against the variable's labels without filtering the `-1`
sentinel, which means two different things on the two sides:

* on the variable side, `-1` marks a **masked entry**;
* in a constraint's `vars`, `-1` marks an **empty (padded) term slot**.

So any masked variable matched any constraint that has a padded term slot, even
when the two are entirely unrelated, and the constraint was silently deleted.
Models built with `mask=` can therefore lose constraints and solve to a wrong
optimum or unbounded.

### The fix

Filter the `-1` entries out of the variable labels before matching, through a
new `common.assigned_labels` helper, applied at the three places that match a
variable's labels against another object's labels:

* `Constraint.has_variable` (dense path) — the one that was actually broken;
* `CSRConstraint.has_variable` — a CSR matrix stores no empty slots, so this
  path was already correct; the guard is added for symmetry;
* the objective cleanup at the end of `remove_variables`, which had the same
  latent mismatch (there it only dropped already-empty terms).

### Test

`test_remove_masked_variable_keeps_unrelated_constraints` in `test/test_model.py`,
parametrized over `freeze` to cover both constraint representations. It builds a
constraint over a masked variable `b` (hence with padded term slots) plus an
unmasked `c`, and a second constraint that genuinely uses the masked `a`, then
removes `a`: the first constraint must survive, the second must still go.

On master the `freeze=False` case fails at `assert not without_a.has_variable(a)`;
the `freeze=True` case passes before and after.

<details>

```
$ uv run pytest -q -p no:randomly --ignore=test/remote
2829 passed, 38 skipped, 192 warnings in 146.41s

$ uv run mypy .
Success: no issues found in 112 source files

$ uv run ruff check .
All checks passed!
```

`test/remote` was skipped locally because the `oetc` extra
(`google-cloud-storage`) is not installed in this environment; it is unrelated
to this change.

Reproduction on master, before the fix:

```python
import pandas as pd
import linopy

i = pd.Index(range(3), name="i")
m = linopy.Model()
a = m.add_variables(coords=[i], name="a", mask=[True, False, True])
b = m.add_variables(coords=[i], name="b", mask=[True, False, True])
c = m.add_variables(coords=[i], name="c")
m.add_constraints(b.sum() + c >= 0, name="no_a")

m.remove_variables("a")          # UserWarning: ... also removes constraints ['no_a']
print(list(m.constraints))       # [] -- 'no_a' never referenced 'a'
```

</details>


FabianHofmann:
thanks @YassineAbdelouadoud I found another bug that I fixed in #895 with these changes in one go ; closing this. thanks for raising and proposing the fix. I will make a bug fix release

PR #893, opened 2026-08-15
fix(sos): order a set by its declaration, not by its coordinate labels

### LINKED ISSUE #892: add_sos_constraints conflates coordinate labels with SOS weights: string dims rejected, unsorted numeric labels silently reorder SOS2 adjacency
Not yet sure if this is a bug or just an api decision.

> [!NOTE]
> This bug report was generated by AI (Claude), and the reproduction was verified by running it against current `master` (`0.8.0.post1.dev141+g0402431f6`).

`Model.add_sos_constraints` takes no `weights=` parameter and uses the coordinate labels of `sos_dim` as the SOS weights. That has two consequences:

1. **A string-labeled dimension cannot carry an SOS constraint at all** — `add_sos_constraints` raises `ValueError: SOS constraint requires numeric coordinates`.
2. **Numeric labels that are not in sorted order silently redefine SOS2 adjacency.** Solvers order SOS members by weight, so "consecutive" means consecutive by *label value*, not by array position. Two models that are element-for-element identical (same declaration order, same bounds, same objective coefficients) reach different optima depending only on what the coordinates are named — both reported as optimal.

## Reproduction

```python
import pandas as pd
import linopy

# 1. string-labeled dimension — SOS impossible
m = linopy.Model()
x = m.add_variables(lower=0, upper=1, coords=[pd.Index(["s1", "s2", "s3"], name="size")])
m.add_sos_constraints(x, sos_type=1, sos_dim="size")
# ValueError: SOS constraint requires numeric coordinates for dimension 'size', but got object

# 2. coordinate values silently define SOS2 adjacency
def solve_sos2(labels):
    m = linopy.Model()
    x = m.add_variables(lower=0, upper=1, coords=[pd.Index(labels, name="size")], name="x")
    m.add_sos_constraints(x, sos_type=2, sos_dim="size")
    m.add_objective((x * pd.Series([1.0, 1.0, 0.1], index=x.indexes["size"])).sum(), sense="max")
    m.solve(solver_name="highs", reformulate_sos=True, output_flag=False)
    return m.objective.value, m.solution["x"]

solve_sos2([0, 1, 2])     # objective 2.0 — nonzero members at positions 0, 1
solve_sos2([30, 10, 20])  # objective 1.1 — nonzero members 10, 20
```

Output:

```
case 1: SOS constraint requires numeric coordinates for dimension 'size', but got object

case 2: identical model, members declared in the same order
  labels [0, 1, 2]   -> objective 2.0  nonzero: [0, 1]
  labels [30, 10, 20] -> objective 1.1  nonzero: [10, 20]
```

In case 2 the objective coefficients are (1.0, 1.0, 0.1) by position in both runs. With labels `[30, 10, 20]` the weight order is 10 < 20 < 30, so the consecutive pairs are {10, 20} and {20, 30} — the first-two-declared pair {30, 10} is no longer expressible, and the optimum drops from 2.0 to 1.1 with no error or warning.

This is not an artifact of the big-M reformulation — all three emission paths agree on value order:

- `sos_reformulation.py:285-288` argsorts members by coordinate value before building adjacency,
- `solvers.py:138` passes `labels.coords[sos_dim].values` as weights to the native `addSOS` APIs (Gurobi, Xpress),
- `io.py:456` writes the coordinate values as weights into the LP file's `sos` section.

## Suggested fix

An optional `weights=` parameter on `add_sos_constraints` (an array along `sos_dim`, defaulting to the current behavior) would resolve both: string-labeled dims get positional or user-supplied weights, and callers who mean array-order adjacency can say so. At minimum the docstring — currently just "The dimension values are used as SOS" (`model.py:928`) — should state that SOS2 adjacency follows *sorted* coordinate values, not array order.

FBumann:
*(your intent here, handwritten)*

> [!NOTE]

Closes #892.

An SOS set now runs in the order its members are declared along `sos_dim`, rather than the order its coordinate labels sort into — the array order linopy reads everywhere else (`isel`, `shift`, `roll`, `diff`).

## Why

Making the labels the SOS weights had two consequences:

- a string-labelled dimension could carry no set at all (`ValueError: SOS constraint requires numeric coordinates`), reaching `piecewise` the day its breakpoints are labelled `[low, mid, high]`;
- numeric labels declared out of order silently redefined which members are adjacent. Two models identical member-for-member reached different optima depending only on what the coordinates were named, both reported optimal.

## What does not change

Ascending numeric coordinates state the same order and are still passed through as the member weights, so a model whose SOS dimension is sorted is unaffected down to the bytes of its LP file — every piecewise formulation among them, since `_validate_numeric_breakpoint_coords` already requires strictly increasing breakpoints.

Descending coordinates are also meaning-preserving: reversing a set maps its consecutive pairs `{i, i+1}` to `{n-i, n+1-i}`, which are consecutive too, so SOS2's feasible set is unchanged. Only a `sos_type=2` set whose numeric coordinates *neither* ascend *nor* descend genuinely regroups; it now warns, and sorting the index restores the previous behaviour. An `sos_type=1` set cannot change meaning at all — its feasible region never depended on the order, `reformulate_sos1` being `x <= M * y` and `sum(y) <= 1` — so it stays silent.

## Notes for review

- The numeric-coordinate requirement is dropped for both SOS types.
- `reformulate_sos_constraints` loses its `argsort`: that permutation is now always the identity, so it is deleted rather than rewritten. `reformulate_sos2` selects the neighbouring segment indicators with `isel` instead of `sel`.
- `test_sos2_unsorted_coords_matches_sorted` passes both before and after, its best pair happening to coincide either way. That is why this was never caught, and why the new tests pick coefficients that separate the two orders.

## Commits

1. `test(sos)` — five tests pinning the overlap where labels and declaration agree, plus seven strict `xfail`s marking where they part. Self-verifying at that commit.
2. `fix(sos)` — the rule, with the `xfail` markers removed.

### DIFF
--- a/doc/release_notes.rst
+++ b/doc/release_notes.rst
@@ -11,6 +11,14 @@ Upcoming Version
 * ``Model.add_expressions`` registers a ``LinearExpression`` or ``QuadraticExpression`` under a name (auto-generated as ``expr0``, ``expr1``, ... if omitted), accessible afterwards via ``Model.expressions`` (an ``Expressions`` container mirroring ``Model.variables``/``Model.constraints``) and removable via ``Model.remove_expressions``.
   Named expressions are persisted by ``Model.to_netcdf``/``linopy.read_netcdf`` and preserved by ``Model.copy``, ``copy.copy``, ``copy.deepcopy``, and pickling.
 
+*SOS constraints*
+
+* ``Model.add_sos_constraints`` no longer requires the SOS dimension to have numeric coordinates.
+
+**Breaking Changes**
+
+* An SOS set is ordered by the positions its members are declared in, rather than by the values of its coordinates. Ascending coordinates state the same order, so a model whose SOS dimension is sorted is unaffected, every piecewise formulation among them; descending ones reverse it, which leaves an SOS set's meaning intact. Only a ``sos_type=2`` set whose numeric coordinates neither ascend nor descend changes meaning, and it now emits a ``UserWarning``; sorting the index restores the previous behaviour. (`#892 <https://github.com/PyPSA/linopy/issues/892>`__)
+
 
 Version v0.9.0
 --------------
--- a/doc/release_notes.rst
+++ b/doc/release_notes.rst
@@ -17,7 +17,7 @@ Upcoming Version
 
 **Breaking Changes**
 
-* An SOS set is ordered by the positions its members are declared in, rather than by the values of its coordinates. Ascending coordinates state the same order, so a model whose SOS dimension is sorted is unaffected, every piecewise formulation among them; descending ones reverse it, which leaves an SOS set's meaning intact. Only a ``sos_type=2`` set whose numeric coordinates neither ascend nor descend changes meaning, and it now emits a ``UserWarning``; sorting the index restores the previous behaviour. (`#892 <https://github.com/PyPSA/linopy/issues/892>`__)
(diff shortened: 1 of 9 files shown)


FabianHofmann:
this looks good. I would include this in a bugfix release

FabianHofmann (approved):
I added a commit about for negative values and some smaller refactorings. this is good to go

Issue #898, opened 2026-08-18
timezone-aware coords do not build after version 0.8.0


peterhron:
### Version Checks (indicate both or one)

- [x] I have confirmed this bug exists on the lastest [release](https://github.com/pypsa/linopy/releases) of Linopy.

- [ ] I have confirmed this bug exists on the current [`master`](https://github.com/pypsa/linopy/tree/master) branch of Linopy.

### Issue Description

since version 0.8.0, a model fails to build if timezone-aware time coords are used. The regression happens in add_variables(), where a strict `broadcast_to_coords() `validation was introduced by https://github.com/PyPSA/linopy/pull/732.

The expected and actual indexes for validation are constructed differently:
```
expected_idx = pd.Index(coord_values) # this happens in alignment.py::_as_index() helper
actual_idx = arr.coords[dim].to_index()
```

For a timezone-aware xarray DataArray, the first conversion produces an object index of integer nanoseconds, while the second produces the original timezone-aware pandas DatetimeIndex. The equality check therefore fails even though both coordinates originate from the same input.  

### Reproducible Example

```python
import linopy
import pandas as pd
import xarray as xr

time = pd.date_range("2026-01-01", periods=2, tz="Europe/Prague", name="time")
coords = xr.Coordinates({"time": time})

model = linopy.Model()
model.add_variables(coords=coords, name="x") # ValueError
```

### Expected Behavior

validation should pass and model should be created with the timezone-aware coord. A minimal fix could be to adapt `_as_index()` helper in alignment.py to call `to_index()` for DataArrays, which retrieves the pandas index directly (as compared to the pd.Index constructor, which goes through numpy arrays, deleting the timezone in the process). 

I propose this small change in https://github.com/PyPSA/linopy/pull/897. 

### Installed Versions

<details>
Python:  3.12.10
Linopy:  0.8.0
xarray: 2025.9.1
pandas:  2.3.3
NumPy:   2.3.5
</details>

PR #897, opened 2026-08-18
fix: _as_index() calls to_index() for DataArrays

peterhron:
Closes #898 (if applicable).

## Changes proposed in this Pull Request
helper in alignment.py uses `DataArray.to_index()` to retrieve the actual pandas index object instead of re-building it via `pd.Index`. This fixes regression where timezone-aware indexes cannot be used to build a model.

Unit tests are added to validate that models can be built with timezone-aware coords.

## Checklist

- [ ] AI-generated content is marked (see [`AGENTS.md`](../blob/master/AGENTS.md)).
- [ ] Code changes are sufficiently documented; i.e. new functions contain docstrings and further explanations may be given in `doc`.
- [ ] Unit tests for new features were added (if applicable).
- [ ] A note for the release notes `doc/release_notes.rst` of the upcoming release is included.
- [ ] I consent to the release of this PR's code under the MIT license.

PR #902, opened 2026-08-19
test(alignment): assert coord values, not just dims

FBumann:
The test also compares the coord values, on top of the existing dim names now

> [!NOTE]

Follow-up to #897, now rebased onto master since that merged — the diff here
is one commit touching only `test/test_alignment.py`.

`TestAddVariablesCoords.test_coords_set_variable_dims` asserted which
dimensions each coords form produces, but never the values behind them. A
coords form that silently reordered, re-typed or truncated its entries passed.

Each case now carries a `dim -> values` mapping instead of a dims tuple, and
the test asserts the resulting index against it, so both halves of the
contract are pinned for all 14 forms — including the three timezone-aware
ones from #897, where the value assertion is what actually catches #898.
Reverting #897's `_as_index` change fails the `xarray-datetime-utc` and
`xarray-datetime-dst` cases under both the legacy and v1 conventions; with it,
the file is 228 passed / 14 skipped and the suite 5253 passed / 493 skipped.

The DST-spanning hourly range is @peterhron's from #897, kept as-is and
lifted into a module-level `_hourly_index` so the expected column can reuse it.

## Checklist

- [ ] A note for the release notes `doc/release_notes.rst` of the upcoming release is included. — n/a, test-only; the note for the fix landed with #897.

### DIFF
--- a/test/test_alignment.py
+++ b/test/test_alignment.py
@@ -565,41 +565,49 @@ class TestCoordsToDict:
 
 
 # ---------------------------------------------------------------------------
-# add_variables — coords / dims map to the variable's dimensions
+# add_variables — coords / dims map to the variable's dimensions and values
 # ---------------------------------------------------------------------------
 
 
-class TestAddVariablesCoords:
-    """End-to-end: each coords / dims form sets the variable's dimensions."""
+def _hourly_index(timezone: str | None) -> pd.DatetimeIndex:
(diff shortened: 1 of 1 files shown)


FBumann:
@peterhron-sf Just for your information, We are doing a follow up also touching your code. We kept it out of your PR because it was not an issue with your tests. This rather improves the whole test, including other cases.

FabianHofmann (approved):
doesn't hurt

PR #907, opened 2026-08-27
fix: sum() over a dimension crashes when a sibling dimension is empty

### LINKED ISSUE #906: sum() over a dimension crashes when a sibling dimension is empty
I found this edge case while working on lpspec. I think its better handled here than in lpspecs linopy lane

> [!NOTE]
> The following content was generated by AI, on behalf of a user of linopy.

Summing an expression over one dimension raises inside the term reshape when *another* dimension of the expression has size 0 — even though summing over the empty dimension itself works, and the result (an expression with no terms) is well-defined.

```python
import linopy, pandas as pd

m = linopy.Model()
x = m.add_variables(
    coords=[pd.RangeIndex(4, name='t'), pd.Index([], name='cycle'), pd.Index(['a', 'b'], name='line')],
    name='x',
)
x.sum('cycle').sizes   # Frozen({'t': 4, 'line': 2}) — fine
x.sum('line')          # ValueError: cannot reshape array of size 0 into shape (4,0,newaxis)
```

Traceback ends in `expressions.py`, `BaseExpression._sum` (`variables.py:1243 → expressions.py:1633 → expressions.py:2141`). Reproduced on `0.9.1.post1.dev3+g09c34dd9d` (master).

Where it bites: a PyPSA-style model whose cycle basis is empty (no meshed network) — `sum(..., 'line')` of a term still carrying the empty `cycle` dimension. Expected: the same result as `LinearExpression.from_constant(m, xr.zeros_like(...))` over the kept coordinates, i.e. `{'t': 4, 'cycle': 0}` with zero terms — which is what `x.sum('cycle')` already returns for the mirror case. Presumably the stacking reshape needs the term count taken from the summed dimension rather than inferred with `-1`, which fails once the array has zero elements.

FBumann:
Closes #906.

> [!NOTE]

`Dataset.stack` reshapes with an inferred `-1`, which numpy cannot resolve once the array has zero elements — e.g. `x.sum('line')` when `x` also carries an empty `cycle` dimension. An empty expression has no terms, so it is now stacked onto a `_term` dimension of size 0, consistent with `x.sum('cycle')` and `LinearExpression.from_constant(...)` over the kept coordinates.

Three commits, best reviewed one by one:
1. strict-xfail regression test
2. pure refactor: extract `_stack_into_term_dim` from `BaseExpression._sum` (no behaviour change)
3. the edge case inside `_stack_into_term_dim`, un-xfail, release note

### DIFF
--- a/doc/release_notes.rst
+++ b/doc/release_notes.rst
@@ -53,6 +53,7 @@ Upcoming Version
 
 **Bug fixes**
 
+* ``sum()`` over a dimension no longer raises when another dimension of the expression has size 0; it returns an expression without terms over the kept coordinates, as summing over the empty dimension itself already did. (https://github.com/PyPSA/linopy/issues/906)
 * A multi-key ``groupby`` now returns its groups sorted by key tuple, like the single-key path. The key combinations were numbered by iterating a ``set``, so the group order was arbitrary and changed between processes with ``PYTHONHASHSEED``.
 
 
--- a/linopy/expressions.py
+++ b/linopy/expressions.py
@@ -181,6 +181,11 @@ def _expr_unwrap(
 logger = logging.getLogger(__name__)
 
 
+def _stack_into_term_dim(ds: Dataset, dims: list[Hashable]) -> Dataset:
+    """Stack ``dims`` into the term dimension."""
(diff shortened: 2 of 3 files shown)


FabianHofmann (approved):
makes all sense!

Issue #919, opened 2026-09-03
Math-spec integration


FabianHofmann:
While the math-spec package (https://github.com/energy-models/math-spec/) is actively evolving, ground work for integrating it into linopy and shaping the API surface can already be tackled. The following proposal comes with couple of logics allow to define 1) a whole optimization problem, 2) parts of math through the spec. In 3) I envisioned a logic  that uses the expressions from math-spec and reports their values. 

1) a function that defines a linopy model from the spec: 

```python
class Model:
    ...
    @classmethod
    def from_spec(cls, spec, sources: Mapping[str, Any]) -> Model:
    ...
```

It takes spec as math-spec-compatible, ie. anything that `math_spec.to_program` accepts (YAML path, YAML string, dict, `Spec`, `Program`) and input data sources as a Mapping. The explicit format of the latter is to be discussed. I could be compatible with long form tables as from [datarecord](https://github.com/energy-models/datarecord) or even work `model.parameters`. The function returns a linopy Model instance. 

2) a function that incrementally updates a model with a spec
```python
class Model:
    ...
    def add_spec(self, spec: mathspec.Buildable, sources: Mapping[str, Any] | None) -> None:
    ...
```
Same inputs, but now extending an already initialized model. Likely, `sources` could be optional as math could run on existing variables. 

3a) a function to report numeric results of expressions that are defined by the spec. When math-spec defines expression may they be data-only, constraint expressions or post-solve expressions (see https://github.com/energy-models/math-spec/pull/287), linopy does explicitly not load them into memory but keeps them in the spec layer, the same as lpspec does today. Such a setup follows the lazy spirit from #888, but now with the math+data as the only reference. It returns a dataarray.

```python
class Model:
    ...
    def report(self, str, sources: Mapping[str, Any] | None) -> xarray.DataArray:
    ...
```

3b) **alternatively** a Report class that is attached to the linopy model and points to all math expressions from the spec as well as in-memory expressions contained by the model. you can materialize the report by selecting an item. 

```
linopy.model.report["lcoe"] -> xr.DataArray
```
While I personally like such an accessor more than the function, here the question arises how to keep the input data tight to the model and make references to the data work out. Therefore you would likely want to initialize the `Report` class with the `sources` as argument.

[...]

PR #922, opened 2026-09-03
Build and change models via math-spec

FabianHofmann:
Follow the plan in #919; support math-spec for model definition and modification.

> [!NOTE]

A math-spec (fluxopt/lpspec#1518's linopy lane) builds a linopy model, or extends one, from `linopy/spec/`. math-spec is pinned to a git tag (v0.0.0-alpha.88) through the PEP 735 dependency group `spec`, Python >= 3.12; `import linopy` never imports math_spec, the model methods import it lazily.

### Building

- `Model.add_spec(spec, sources, retain="report", name=None, build_expressions=True)` builds variables, SOS, constraints and objective into a v1 model; `Model.from_spec(spec, sources, retain=..., **model_kwargs)` is sugar over it on a fresh model. `spec` is a path, YAML text, dict or `math_spec.Spec`. Both raise under legacy semantics and emit `EvolvingAPIWarning` once per session while the API stabilises.
- `linopy.spec.attach(program, sources, retain=...)` turns a lowered `Program` plus user data into an `Attached`: master coordinates, lookups and on-demand parameters. `sources` is any `Mapping` pulled by key and never iterated (a lazy view over PyPSA's `n.c.<component>.da` works) or a single `xr.Dataset`. math-spec's binding rules are enforced: members only from the dimension's own source, source order never sorted, parameters never read for labels, every lookup's values are labels of the dimension it maps into. Unknown labels, duplicate rows, wrong rank or dtype raise `SpecDataError`. Aligned data is never copied.
- Absence is uniform: a missing parameter row is refused wherever it is used (coefficient, bound, constant side, divisor); mask the coordinate out with a `where` or fill the data. Divisor coverage is checked on named-expression bodies, unaligned operands raise `SpecDataError`, a constraint with its term on the right is swapped with a flipped sense.
- Modules: `attach`, `builder` (declarations), `evaluate` (recursive evaluator, fold), `coverage` (one tree walk collecting divisor, constant-side and coefficient obligations), `operators` (`sum_over`, `grouped_sum`, `at`, `shift`, `sum_back`), `where`, `curves` (piecewise), `terms`, `parameters` (resolution and derivation), `groups`, `context`, `nodes` (the parameters a node names and the dimensions it spans, over math-spec's own `walk`), `netcdf`, `testing` (`synthetic_sources`).

### Spec layers

A spec may be layered onto an already-built model, for example a PyPSA model with hand-built variables, to add custom constraints, caps and named expressions that read the existing variables.

- **Binding through `sources`.** A declared variable whose `sources` entry is a linopy `Variable` is bound to it instead of built. The declaration must match the model variable: same dimension names, default bounds, no `where:`, same domain. Everything else in `sources` stays ordinary data.
- **Collision guards.** A spec that introduces a variable, constraint, SOS or named expression the model already holds is refused before anything is built ("bind it or rename it"), and so is an objective on a model that already has one. The guard runs both ways: `remove_variables`, `remove_constraints` and `remove_expressions` refuse to drop a name any layer builds or binds.
- **Stamping and eager expressions.** Every variable, constraint and expression a layer builds carries the layer's name in `attrs["spec"]`, read through a `spec` property (`str | None`) on `Variable`, `Constraint`, `LinearExpression` and `QuadraticExpression`. A bound variable carries `None`, and so does a derivative of a stamped object (`expr * 2`, `linopy.merge`, `add_constraints(expr >= 0)`). A named expression whose body holds a variable term is built into `model.expressions` under its declared name at `add_spec` time and stamped, so `model.expressions[name]` and `model.spec.expressions[name].expression` are one object; a data-only body or one reading a constraint's `dual` stays spec-only and folds on read. `build_expressions=False` keeps a layer's named expressions lazy; the flag is not persisted, a model read from a file folds whatever the file does not hold.
- **Layers.** `model.spec` is a `ModelSpec` over ordered, named `Layer`s, one per `add_spec` (`name=`, else the file stem, else `"spec"`). `model.spec[name]` reads a layer; `model.spec.expressions` is the union across layers; typesetting joins the layers and `standalone=True` needs a single layer. The single-layer passthroughs (`program`, `text`, `parameters`, `coords`, `lookups`) raise on several layers and point to `model.spec[name]`. `unspecified` reports what no layer declares, plus a `bound` bucket for the model variables a layer binds rather than builds. A hand-built variable that shares a bound spec name stays visible in `unspecified`. The objective is tracked by `objective_owner`, cleared when the objective is replaced by hand.
- **Dimension reconciliation.** A dimension's master index comes from `sources`, else from an earlier layer, else from the bound variables spanning it, which must all agree; a disagreeing claimant raises "the same dimension name means the same axis". A bound variable that spans a strict subset of the master in master order is reindexed onto it lazily, with absent slots at `labels == -1`; labels the master does not hold are refused. SOS declarations always act on the model-owned variable.

### Reading back

- `model.spec.expressions[name]` is a `NamedExpression` with `.node` (lowered formula), `.dims` (static, no data bound), `.expression` (the unsolved linopy expression: the stored `model.expressions[name]` where the layer built one, else folded on read) and `.solution` (the fold over the solved model, substituting `Variable.solution`, retained parameters and a constraint's `dual`). `model.spec.evaluate(name, sources)` returns the same object with parameters attached afresh (for `retain="none"`), refusing sources labelled unlike the model.
- Typesetting: `model.spec.to_latex/to_markdown/to_typst` render the whole model; `model.spec.declaration(name)` and a `NamedExpression` render one declaration as a bare line. `to_markdown` is math-spec's GitHub-flavoured Markdown; in a notebook the accessor renders itself with the `$$` and `$` pairs MathJax reads. The typeset and repr name the bound variables and note that the spec reads variables whose bounds and mask the render leaves out.
- The `Model` repr says the model is built from or extended by math-spec layer(s), prints the spec description, and tags spec-owned variables, constraints and expressions per layer once hand-added items of that kind exist beside the spec; a whole-spec model's repr carries no tags.

### Persistence

- `to_netcdf` writes each layer under its own `spec-{name}` prefix: the spec text and the binding map as layer-named attributes, master coordinates and lookups, and object-dtype parameters as `pandas.factorize` codes and categories, so partial maps keep holes and dtypes on both netcdf engines. Every parameter records its in-memory dtype and is cast back on read; master coordinates are stamped onto every container. Layer order, `whole` and the objective owner are model-level attributes. `read_netcdf` re-lowers each `Program` from its text, and a file without a spec loads without math-spec installed.
- `Model.copy()` carries every layer; `assert_model_equal` compares every layer's text, bindings and parameters including dtypes.
- pandas extension strings (Arrow-backed under pandas 3) are attached as numpy object arrays.

### Docs, CI, benchmarks

- Notebook `examples/building-models-from-specs.ipynb`, wired into the user guide and executed on Read the Docs and in the notebook CI job with the `spec` group installed. API pages for `linopy.spec`, release notes added.
- The 3.12 and 3.13 test jobs install the `spec` group; mypy runs with math-spec installed. Tests skip without it.
- `benchmarks/models/spec_pypsa.py`: `from_spec` on math-spec's `examples/pypsa.yaml` plus the two netcdf phases; skips without math-spec or `MATH_SPEC_EXAMPLES`.

### Hardening (second pass, `dev-scripts/spec-hardening/plan.md`)

- **Refusals instead of guesses.** A constraint row whose variable terms the data all masked out used to build as `0 sense rhs` and vanish from the LP; it is refused (warned about for `absence: zero` against a zero other side). A quadratic term in a constraint raises `NotImplementedError` before anything is built. `sum_back` refuses a window wider than its axis unless `edge="wrap"`. An unnamed square wide table is no longer orientation-guessed. A `pd.MultiIndex` dimension, a mistyped dimension label and a `Variable` bound from another `Model` raise `SpecDataError`.
- **Ownership registry.** `Model._ownership` records which layer built, bound or declared each name. `Variables/Constraints/Expressions.remove`, `add_*`, the objective setter and `remove_sos_constraints` consult it, so a spec-owned name cannot be dropped or shadowed through any door; a list removal is refused as a whole before any name goes. `Model.remove_spec(name)` unwinds one layer and refuses when another layer still binds one of its variables. `ModelSpec.layers` is read-only.
- **One front door.** `linopy.spec.normalize_spec` tells YAML text, a path, a mapping/`Spec` and a `Program` apart the same way everywhere; an empty spec and a layer name netcdf cannot carry (`/`, `-`) are refused at `add_spec`. A `Sources` adapter reads each key once; keys the spec does not declare are ignored and listed on `Attached.unused` (`strict=True` restores the refusal), and a near-miss key warns as a likely typo.
- **Files.** The netcdf header carries the math-spec version and a layout number; a mismatch warns on read. Reading a spec file without math-spec installed warns and loads a plain model instead of raising.
- Duals: `.expression` on a dual raises `TypeError` (no symbolic form), folding an unsolved one raises the spec's own "no dual yet".

### Known limits

- Variable divisors and degree-2 named expressions are refused by math-spec at lowering.

<details>
<summary>Verification (69507135)</summary>

```
uv run pytest -q --deselect test/test_io.py::test_to_xpress
1541 passed, 519 skipped (xpress licence failures deselected, present on master too)
uv run mypy linopy
Success: no issues found in 50 source files
```

</details>

## Checklist


linopy/model.py
```diff
@@ -437,6 +440,91 @@ def solution(self) -> Dataset:
...
+            If the model was not built from a spec.
+        """
+        if self._spec is None:
+            raise AttributeError(
+                "This model was not built from a spec. Use `Model.add_spec` or "
+                "`Model.from_spec` to build one."
+            )
+        return self._spec
+
+    def add_spec(
```
FBumann, line 460: @FabianHofmann This method is doing the same as `Model.from_spec()`, but with a misleading name.

WHat this method is trying to do seems much harder than building a linopy Model from scratch.

FBumann:
@FabianHofmann I fixed some bugs and improved the IO.

Now I have a PR left that improves the reviewability of a Model built from a spec: #951 

Furthermore, `add_spec()` in its current form should be removed. Adding a spec to an existing model can be a follow up. Although i guess it will conflict with mathspecs purpose (fully self contained Spec vs fragments mentioning outside variables...?)

FabianHofmann:
I want to skip this layered approach as it does not feel safe to me. we stick to the whole math.yaml approach in #958 at this stage

PR #921, opened 2026-09-03
feat(spec): xarray data binder for math-spec programs

FabianHofmann:
Step 2 of the plan in #919: the data layer that `Model.from_spec` / `Model.add_spec` will build on. No public model API yet; `linopy.spec` is importable only when math-spec is installed and `import linopy` never loads it.

> [!NOTE]
> The following content was generated by AI.

## Changes proposed in this Pull Request

`linopy/spec/binder.py` turns a lowered `math_spec.Program` plus user data into what a builder needs: master coordinates, lookup arrays and on-demand parameter arrays.

- `bind(program, sources, retain=...)` returns a `Bound` with `coords`, `lookups`, `parameter(name)` and `retained()`. Parameters are resolved from `sources` on demand and never cached; `retain` selects what `retained()` hands back for persistence: `"report"` (default, the parameter closure of the named expressions), `"all"` or `"none"`.
- `sources` is any `Mapping` pulled by key and never iterated, so a lazy view over PyPSA's `n.c.<component>.da` accessor is a valid source. A single `xr.Dataset` is accepted too.
- math-spec's three binding rules are enforced: members only from the dimension's own source, source order never sorted, parameters never read for labels. Unknown labels, duplicate rows, wrong rank and wrong dtype raise `SpecDataError`. A missing row is not an error: it becomes NaN (bool: False) and the builder decides later.
- No copies of aligned data: indexes are compared first and `reindex` runs only on a real mismatch. An aligned array keeps its buffer, asserted with `np.shares_memory`. A matching index with a different dtype is re-stamped onto the master dtype without copying.
- Accepted parameter shapes: scalar, `xr.DataArray`, `pd.Series`, tidy or wide `pd.DataFrame`, dict. Lookups: Series, dict or DataArray over the dimension.
- The malformed-data cases of lpspec's `test_data_parity.py` are ported and give the same verdicts.
- math-spec is not on PyPI, so there is no extra yet. Tests skip without it; run them with `uv run --with <math-spec checkout> pytest test/test_spec_binder.py`.

Left for the model integration step: object-dtype string lookups through netcdf, float promotion of partly mapped lookups into integer dimensions, and derived piecewise parameters.

<details>
<summary>Verification</summary>

```
uv run --with <math-spec> pytest test/test_spec_binder.py -q
240 passed
coverage linopy/spec/binder.py 100%
uv run --with <math-spec> mypy linopy/spec
Success: no issues found in 3 source files
uv run pytest test/test_spec_binder.py -q
1 skipped
[...]

PR #923, opened 2026-09-03
Round trip spec-built models through netcdf and benchmark from_spec

FabianHofmann:
Step 4 of the plan in #919, stacked on #922: a spec-built model survives `to_netcdf` / `read_netcdf` and `Model.copy()`, and `Model.from_spec` gets a codspeed benchmark.

> [!NOTE]

- `to_netcdf` writes the spec YAML text as the top-level attribute `_linopy_spec`, next to `_linopy_version`, following the relaxed-registry and piecewise-formulation precedent. `read_netcdf` re-lowers the `Program` from that text; a file without a spec loads without math-spec installed.
- One extra sub-dataset under the prefix `spec-` carries what `model.parameters` cannot round-trip on its own: master coordinates as `coords__<dim>` data variables (an index coordinate no retained parameter reaches would otherwise be dropped on read), and lookups plus every string-kind parameter as `pandas.factorize` output, `codes__<name>` (int32, -1 for a hole) plus `cats__<name>` with the in-memory dtype in a variable attribute. No netcdf type carries an object array of strings with NaN: netCDF4 returns `<U` and turns a hole into an empty string, scipy narrows int64 to int32. Codes plus categories round-trip str->str, str->int, int->str and int->int lookups, full and partial, exactly on both engines. This resolves the two lookup items left open in #921.
- `Model.copy()` carries the spec; the accessor is rebound to the copy. On read, every parameter is cast back to its in-memory dtype and the master coordinates are stamped onto all containers, so one dimension has one dtype even after the scipy engine narrows int64 to int32. `assert_model_equal` now compares parameters with dtypes. `linopy/spec/testing.py` holds the `synthetic_sources` generator shared by tests and benchmark.
- `assert_model_equal` compares the spec text; parameters, coordinates and lookups are already covered through `model.parameters`.
- `test/test_spec_io.py`: round trips for `retain="report"|"all"|"none"` on both engines, a solved model whose named expression is identical before and after, `evaluate` on a restored `retain="none"` model, exact-dtype assertions per lookup kind, copy semantics, and math-spec's `examples/pypsa.yaml` with synthetic data when `MATH_SPEC_EXAMPLES` points at a math-spec checkout.
- `benchmarks/models/spec_pypsa.py`: `from_spec` on `examples/pypsa.yaml` plus the two netcdf phases; skips without math-spec or `MATH_SPEC_EXAMPLES`. Running it in CI needs math-spec installed and the examples directory available in the codspeed workflow, which waits for a math-spec PyPI release.

<details>

```
uv run --with <math-spec> pytest test/test_spec_binder.py test/test_spec_builder.py test/test_spec_io.py test/test_io.py -q
465 passed, 155 skipped
uv run pytest test/test_io.py test/test_spec_io.py -q          (without math-spec)
93 passed, 16 skipped
uv run --with <math-spec> mypy linopy benchmarks
Success
MATH_SPEC_EXAMPLES=<math-spec>/examples uv run --with <math-spec> --with pytest-benchmark pytest benchmarks -k spec --benchmark-disable
6 passed, 1 skipped
```

</details>

## Checklist

### DIFF
--- a/benchmarks/models/__init__.py
+++ b/benchmarks/models/__init__.py
@@ -21,5 +21,6 @@ from benchmarks.models import (  # noqa: F401
     qp,
     sos,
     sparse_network,
+    spec_pypsa,
     storage,
 )
--- a/benchmarks/models/spec_pypsa.py
+++ b/benchmarks/models/spec_pypsa.py
@@ -4,67 +4,26 @@ Model built from math-spec's ``pypsa.yaml`` example (requires math-spec).
 The subject is :meth:`linopy.Model.from_spec`: lowering a spec of PyPSA's full
 statement, binding synthetic data to it and building every variable and
 constraint it declares. The example lives outside the wheel, so its directory
-comes from ``MATH_SPEC_EXAMPLES`` and the case skips without it.
+comes from ``MATH_SPEC_EXAMPLES`` and the case skips without it. A sweep
+value is the number of labels per dimension; 40 of them is about 20k
+variables.
 """
 
 from __future__ import annotations
 
 import os
 from pathlib import Path
-from typing import TYPE_CHECKING, Any
-
-import numpy as np
-import pandas as pd
-import xarray as xr
+from typing import TYPE_CHECKING
 
 from benchmarks.registry import BUILD, FROM_NETCDF, TO_NETCDF, BenchSpec, register
 
 if TYPE_CHECKING:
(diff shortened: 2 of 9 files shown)


FabianHofmann:
Folded into #922 (fast-forwarded `spec-builder` onto this branch) so build, fold and persistence are reviewed together.

PR #927, opened 2026-09-04
fix(constraints): CSRConstraint stores label columns; frozen constraints survive later variable changes

### LINKED ISSUE #926: CSRConstraint: use label columns; frozen constraints break after later variable add/remove
> [!NOTE]
> The following content was generated by AI (Claude Code), reviewed by @FabianHofmann.

## Bug: frozen constraints break after any later variable change

`CSRConstraint` stores **dense variable positions** as column indices and a fixed column count (`n_active_vars` at freeze time). Any `add_variables` or `remove_variables` after freezing changes both, so `model.matrices` fails when stacking the blocks:

```
ValueError: inconsistent shapes (3, 6) and (1, 8)
```

This is independent of #870; it reproduces on `master` with a plain dense `freeze=True` constraint.

<details>
<summary>Reproduction</summary>

```python
import linopy, pandas as pd

m = linopy.Model()
i = pd.RangeIndex(3, name="i")
a = m.add_variables(coords=[i], name="a")
b = m.add_variables(coords=[i], name="b")
m.add_constraints(2 * b >= 1, name="c1", freeze=True)

d = m.add_variables(coords=[pd.RangeIndex(2, name="i")], name="d")
m.add_constraints(d <= 5, name="c2", freeze=True)
m.matrices.A  # ValueError: inconsistent shapes (3, 6) and (1, 8)

# same after m.remove_variables("a") instead of adding: (3, 6) vs (1, 5)
```

</details>

## Proposal: label columns for `CSRConstraint`

Store **raw variable labels** as column indices (width `model._xCounter`), and map labels to dense positions once at matrix assembly via `VariableLabelIndex.label_to_pos`. This is the convention `CSRPayload` (#870) already uses for expressions, so the two CSR representations would agree.

Consequences:

1. `to_matrix` / `to_matrix_with_rhs` gather positions through `label_to_pos` and return `n_active_vars`-wide blocks; the bug above disappears by construction. `has_labels`, `to_polars` and the dataset reconstruction no longer need `vlabels` lookups.
2. `from_payload` drops its label-to-position mapping and no longer needs the label index at construction time.
3. The netcdf `_linopy_format: csr` writer stores labels; the reader gets a compatibility path for files written with positions (they carry `vlabels` in the same file).
4. `LinearExpression.to_constraint` on a payload-backed expression can return an **unassigned** `CSRConstraint` (`cindex=None`) directly. `Constraint._from_pending`, the lazy `Constraint.data` branch and `extract_pending` introduced in #870 as a stopgap can then be deleted; `add_constraints` only assigns labels.
5. `linopy/persistent/snapshot.py` reads `to_matrix_with_rhs` output and needs a check that buffer ownership assumptions still hold.

This is also the prerequisite for making the CSR-backed constraint the default without a dense-to-sparse conversion step.

Refs: #870

FabianHofmann:
_(handwritten intent goes here)_

> [!NOTE]
> The following content was generated by AI (Claude Code), reviewed by @FabianHofmann.

Fixes #926 (points 1, 3 and 5; points 2 and 4 depend on #870).

`CSRConstraint` stored dense variable positions as CSR column indices with a width fixed at freeze time. Any `add_variables` or `remove_variables` after a `freeze=True` constraint made `model.matrices` fail with `inconsistent shapes`.

**Changes**

- `CSRConstraint` stores raw variable labels as columns (width `model._xCounter`). A shared helper maps labels to dense positions once at matrix assembly, for both the dense `Constraint` and the frozen path, and raises if a constraint references a removed variable.
- The returned block keeps the stored `data` and `indptr` arrays by identity and preserves the int32 index dtype, so the persistent snapshot diff can still skip untouched frozen constraints and index arrays do not double in size.
- netcdf: the writer tags `_csr_columns: labels`; the reader maps legacy position files back through the stored `vlabels` and raises on unknown tags.
- Separate commit: `CSRConstraint.indexes` built an xarray `Indexes` without a variables mapping, which made `ModelSnapshot.capture` fail on any model with a frozen constraint.

**Tests**

Add/remove variables after freeze against the unfrozen model, netcdf round-trip incl. legacy layout, buffer identity and empty diff for an untouched frozen model.

<details>

| Check | Result |
| --- | --- |
| `ruff check` / `ruff format` | clean |
| `mypy linopy` | 28 errors, all pre-existing in `solvers.py` |
| `pytest -k "not xpress" --ignore=test/remote` | 6350 passed, 532 skipped |

</details>


FabianHofmann:
> [!NOTE]
> This comment was generated by AI (Claude Code) on behalf of @FabianHofmann.

Superseded by #929, which rebases these commits onto the merged #870 and completes points 2 and 4 of #926.

PR #959, opened 2026-09-19
Compose spec fragments into a whole model via given

FabianHofmann:
Closes # (if applicable).

_TODO (human): write your intent by hand here._

> [!NOTE]

Stacked on #958. This PR adds math-spec's `given:` key through **composition**, not binding.

- **Compose fragments.** Re-export `math_spec.merge` and `math_spec.override` as `linopy.spec.merge` / `linopy.spec.override`. A fragment declares under `given:` the variables and constraints it reads but does not build; `merge` folds those reads into the fragment that introduces them, `override` lays a patch over a base. Both return a whole-model dict that `Model.from_spec` builds as usual.
- **Refuse a residual `given`.** `build_into` (`linopy/spec/accessor.py`) raises `ValueError` when a spec still reads a `given:` name after lowering. The guard sits right after `normalize_spec` and before any data is attached, so a fragment fails with a clear message instead of a downstream attach or coverage error. This keeps the whole-model-only stance of #958.
- **Tests.** New `test/test_spec_composition.py` covers merge/override folding, building and solving a composed model, refusal of a lone fragment and of a merge with no introducer, and a netcdf round-trip.
- **Docs.** Release-notes bullet, `doc/api.rst` composition note plus `spec.merge` / `spec.override` in the API listing.

> [!WARNING]
> **math-spec pin is not release-ready.** `pyproject.toml` pins `math-spec` to the `feat/given` branch head SHA behind a `TODO`, because no released alpha ships `merge` / `override` / `Program.given` yet. An alpha tag must be cut from `feat/given` and the pin updated before merge. #958 must merge first.

## Checklist


FBumann:
@FabianHofmann Im confused by the pr description

The power of `given:` as i imagine it is, that users can add a fragment to any linopy model, not just ones buildt with a spec.

Linopy itself just needs to ensure the contract is ok (given variables exist and dims match)?

Or does it not work as imago ed? Or is it just the next step?

